PR #15, opened 2024-01-09
Really fix jq format in minifying index

stevenyomi:


FourTOne5:
3rd times the charm 🤞

PR #59, opened 2024-01-10
Updated parsing of manga title from URL.

### LINKED ISSUE #58: Madokami incorrect manga titles
### Source information

Madokami 1.4.7

### Source language

English

### Steps to reproduce

Search or browse manga from Madokami extension.

### Expected behavior

Titles for the manga returned are the correct manga titles.

### Actual behavior

Many manga returned have titles like "!MangaPlus", "!(Digital) (danke-Empire)]", etc instead of the correct title.

Screenshot_20240109_202659_Tachiyomi

### Tachiyomi version

r6404

### Android version

Android 14

### Other details

This has been an issue for ages, and has always annoyed me. The reason this issues occurs is because of the way Madokami uploads are organized.
For example, this screenshot shows how it looks if you browse the latest madokami manga in tachiyomi. You can see there are multiple manga that have strange titles.

Screenshot_20240109_202659_Tachiyomi

if you look at the first incorrect title in that screenshot it is actually "Oshi no Ko" and the following screenshot shows the directory structure in Madokami and why it appears as "!MangaPlus" instead. The extension currently uses the last part of the URL as the manga title.

Screenshot_20240109_202712_Tachiyomi

I've attempted a fix for this already and will create a PR shortly:

Screenshot_20240109_202829_Tachiyomi

### Acknowledgements

kooper100:
Checklist:

Closes #58 

Tested by building apk and installing on my Samsung S23.

Screenshot_20240109_202829_Tachiyomi

### DIFF
--- a/src/en/madokami/build.gradle
+++ b/src/en/madokami/build.gradle
@@ -6,7 +6,7 @@ ext {
     extName = 'Madokami'
     pkgNameSuffix = 'en.madokami'
     extClass = '.Madokami'
-    extVersionCode = 7
+    extVersionCode = 8
 }
 
 apply from: "$rootDir/common.gradle"
--- a/src/en/madokami/src/eu/kanade/tachiyomi/extension/en/madokami/Madokami.kt
+++ b/src/en/madokami/src/eu/kanade/tachiyomi/extension/en/madokami/Madokami.kt
@@ -69,6 +69,7 @@ class Madokami : ConfigurableSource, ParsedHttpSource() {
         manga.url = element.attr("href")
         val pathSegments = element.attr("href").split("/")
         var i = pathSegments.size
+        manga.description = URLDecoder.decode(pathSegments[i - 1], "UTF-8")
         do { i--; manga.title = URLDecoder.decode(pathSegments[i], "UTF-8") } while (URLDecoder.decode(pathSegments[i], "UTF-8").startsWith("!"))
         return manga
     }
@@ -111,8 +112,7 @@ class Madokami : ConfigurableSource, ParsedHttpSource() {
     override fun mangaDetailsParse(document: Document): SManga {
         val manga = SManga.create()
         manga.author = document.select("a[itemprop=\"author\"]").joinToString(", ") { it.text() }
-        manga.description = "Tags: " + document.select("div.genres[itemprop=\"keywords\"] a.tag.tag-category").joinToString(", ") { it.text() }
-        manga.genre = document.select("div.genres a.tag[itemprop=\"genre\"]").joinToString(", ") { it.text() }
+        manga.genre = document.select("div.genres a.tag").joinToString(", ") { it.text() }
         manga.status = if (document.select("span.scanstatus").text() == "Yes") SManga.COMPLETED else SManga.UNKNOWN
         manga.thumbnail_url = document.select("div.manga-info img[itemprop=\"image\"]").attr("src")
         return manga
--- a/src/en/madokami/src/eu/kanade/tachiyomi/extension/en/madokami/Madokami.kt
+++ b/src/en/madokami/src/eu/kanade/tachiyomi/extension/en/madokami/Madokami.kt
@@ -67,7 +67,9 @@ class Madokami : ConfigurableSource, ParsedHttpSource() {
     override fun popularMangaFromElement(element: Element): SManga {
         val manga = SManga.create()
         manga.url = element.attr("href")
-        manga.title = URLDecoder.decode(element.attr("href").split("/").last(), "UTF-8").trimStart('!')
+        val pathSegments = element.attr("href").split("/")
+        var i = pathSegments.size
+        do { i--; manga.title = URLDecoder.decode(pathSegments[i], "UTF-8") } while (URLDecoder.decode(pathSegments[i], "UTF-8").startsWith("!"))
         return manga
     }
 


FourTOne5:
Can you send an example url?

mage-goo:
[I misread the code, never mind this]

mage-goo:
Also, just fyi, I always encountered broken image and really slow image loading in general for madokami whenever I used DoH for some reason, so you might want to try to turning it off during development of this specifically

mage-goo:
Ok, here's some edge case:
- [!Trap Oneshots](https://manga.madokami.al/Manga/Oneshots/%21Trap%20Oneshots) would show Oneshots as the name (even thought it should have !Trap Oneshots)
- [Searching suki na ko megane](https://manga.madokami.al/search?q=suki+na+ko+megane) would result in 5 different entries but all have the same name and image, which is confusing to say the least. You won't be able to tell which is which from inside tachiyomi at all until you actually use the open in webview feature

kooper100:
> This will make some entries have obscenely long title, and isn't there a hard limits on Tachiyomi's title length? Also, it stops as soon as it encounters path that starts with exclamation mark, right? There's some entries that still continues even after that, will post example later. Also have you checked how this works with manga from search function? I think the href works different there And there's also the case of the old !Autouploaded from source folder with does have exclamation mark at the start

Not sure I understand why the title would be obscenely long? Can you explain if I'm missing something? It splits the url at each "/" and uses the last path segment that doesn't start with an exclamation, which from what I've seen should be the actual manga title.

Yes It stops when it encounters a path that doesn't have an exclamation point, but starting from the end. So it checks the last path segment like the extension currently uses, and then checks the previous path segment until it hits a path segment that doesn't start with an exclamation point.
For example here it would check "!chapters" first, then "!(Digital) (danke-Empire)", then finally "Amagami-san Chi no Enmusubi" which it would use as the title.
2024-01-10_08-59-41_firefox


I have tested search and haven't seen any issues with this change.

Since the change is looking at the **last** path segment and each previous segment until it hits a path without an exclamation point I don't think !Autoupload would have an issue. Also I don't even see a "!Autoupload" folder, I only see "_Autouploads".



> Also, just fyi, I always encountered broken image and really slow image loading in general for madokami whenever I used DoH for some reason, so you might want to try to turning it off during development of this specifically

Good to know, I'll keep it in mind. Thanks!

kooper100:
> Ok, here's some edge case:
> 
>     * [!Trap Oneshots](https://manga.madokami.al/Manga/Oneshots/%21Trap%20Oneshots) would show Oneshots as the name
> 
>     * [Searching suki na ko megane](https://manga.madokami.al/search?q=suki+na+ko+megane) would result in 5 different entries but all have the same name and image, which is confusing to say the least. You won't be able to tell which is which from inside tachiyomi at all until you actually use the open in webview feature

Not sure I agree it's any more confusing than what is currently shown. You wouldn't even know what the manga is the way it currently works without opening webview.

Agree with the Oneshots though, not sure how to fix that yet.

kooper100:
> Ok, here's some edge case:
> 
>     * [!Trap Oneshots](https://manga.madokami.al/Manga/Oneshots/%21Trap%20Oneshots) would show Oneshots as the name (even thought it should have !Trap Oneshots)
> 
>     * [Searching suki na ko megane](https://manga.madokami.al/search?q=suki+na+ko+megane) would result in 5 different entries but all have the same name and image, which is confusing to say the least. You won't be able to tell which is which from inside tachiyomi at all until you actually use the open in webview feature

Actually it doesn't look like any of those oneshots even show up in the extension currently anyway. I can't search or find them in tachiyomi using the current extension.

mage-goo:
> You wouldn't even know what the manga is the way it currently works without opening webview.

The manga detail processing still works and it uses the right cover for me somehow 
image

>Not sure I agree it's any more confusing than what is currently shown

Yeah, fair, but the new way also doesn't seems to be ideal
It just seems like there's a better way to do this but the other option like adding the sub-folder(s) name as suffix also seems like it got worse trade-off

kooper100:
> > You wouldn't even know what the manga is the way it currently works without opening webview.
> 
> The manga detail processing still works and it uses the right cover for me somehow image
> 
> > Not sure I agree it's any more confusing than what is currently shown
> 
> Yeah, fair, but the new way also doesn't seems to be ideal It just seems like there's a better way to do this but the other option like adding the sub-folder(s) name as suffix also seems like it got worse trade-off

Do you use Tachiyomi main or a fork like J2K? I noticed that grabbing the image in search or latest manga list works in J2K but not in main.
It's gets the image if you open the manga, but not in search or latest. Not sure if tachi main made some changes that removed that functionality while it remains in j2k?

I considered appending subfolders to the end of the title but I agree it would get messy for some manga. I could add it to the description or something in the manga details so you at least don't have to open it in webview?

mage-goo:
ok here's the search that will "break" this
- [afternoon kodansha](https://manga.madokami.al/search?q=afternoon+kodansha) (raw)

but I guess that doesn't really affect much, it'll just show entries with duplicate name but it seems to work fine when opened?

mage-goo:
> Do you use Tachiyomi main or a fork like J2K? I noticed that grabbing the image in search or latest manga list works in J2K but not in main.

I'm using main tachiyomi though still on 1.49, haven't updated it since this shenanigans began

> I could add it to the description or something in the manga details so you at least don't have to open it in webview?

Oh, that sounds great, and while you're at it, if you could make the tags actually work as tags instead of description, that'd be great.

Though I think that would go out of the PR scope a bit, but eh

kooper100:
Added the subfolder to manga description and fixed tags to actually be tags instead of listed in the description.

Screenshot_20240110_174219_Tachiyomi

Screenshot_20240110_174240_Tachiyomi

mage-goo:
lgtm 👍

AwkwardPeak7 (approved):

PR #113, opened 2024-01-10
Improve Dynasty Scans

rtlow:
Checklist:

These fixes address two outstanding issues:

1. In all subextensions except `DynastyChapters`, browsing the category would never advance to the next page, only showing the content from the first page.
2. In `DynastyDoujins`, trying to access an entry that contains no chapters but does contain images [(example)](https://dynasty-scans.com/doujins/the_amazing_digital_circus) would result in an unhandled IndexOutOfBoundsException, preventing users from viewing the images as intended.

### DIFF
--- a/src/en/dynasty/build.gradle
+++ b/src/en/dynasty/build.gradle
@@ -6,7 +6,7 @@ ext {
     extName = 'Dynasty'
     pkgNameSuffix = 'en.dynasty'
     extClass = '.DynastyFactory'
-    extVersionCode = 21
+    extVersionCode = 22
 }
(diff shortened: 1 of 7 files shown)


beer-psi (approved):

beer-psi (changes requested):
Since this affects most Dynasty sources, could you apply the changes to the base class instead?

rtlow:
> Since this affects most Dynasty sources, could you apply the changes to the base class instead?

Done

beer-psi (approved):

PR #138, opened 2024-01-11
ci: Properly configure the rest of issue moderator

beer-psi:
- Add a blurb for Cloudflare `/blurb cf`
- Point issue moderator to the correct repo URL so the existing extension check works
- Move to a separate token for issue moderation. The token only has access to keiyoushi/extensions-source and these permissions:
  - Read access to members 
  - Read access to metadata  
  - Read and Write access to issues 

Checklist:


beerpiss:
Just realized commands may not work because the default `GITHUB_TOKEN` can't see private org members. Oh well.

AwkwardPeak7 (approved):

PR #139, opened 2024-01-11
Load all chapters for Dynasty Doujins

rtlow:
Checklist:

Fixes `Dynasty Doujins` only showing the first page of results for Doujin chapters. Some Doujin entries contain multiple pages, which need to be individually parsed [(example)](https://dynasty-scans.com/doujins/touhou_project).

This is a lazy/basic implementation that needs improvement and user quality of life enhancements.

Possible Improvements:

- Currently fetches one page of results at a time. This is slow, especially with only so few chapters per page and some sources listing many tens of pages. A parallel, asynchronous way to fetch chapters would be preferable.
- Chapter list doesn't load until parsing is complete. A major pain considering the previous point.
- Basic method of checking for a successful response and retrying. Some form of backoff would be preferable.

Possible quality of life enhancements:

- Display a Toast whenever a source has over 5 pages of content, then successive Toasts every 5 pages notifying the user of the current page being parsed
- Display a Toast whenever a page needs to be re-requested with number of retries before failure. 

As I am reaching the limits of my experience and patience with Kotlin, I will leave these improvements for others to make.


AwkwardPeak7:
jfyi there are conflicts

rtlow:
> jfyi there are conflicts

No idea why there were, but they're resolved now

src/en/dynasty/src/eu/kanade/tachiyomi/extension/en/dynasty/DynastyDoujins.kt
```diff
@@ -49,10 +49,20 @@ class DynastyDoujins : DynastyScans() {
 
     override fun chapterListSelector() = "div#main > dl.chapter-list > dd"
 
+    private fun doujinChapterParse(document: Document): List<SChapter> {
+        return try {
+            document.select(chapterListSelector()).map { chapterFromElement(it) }.toMutableList()
+        } catch (e: IndexOutOfBoundsException) {
+            mutableListOf()
```
stevenyomi:
```suggestion
            document.select(chapterListSelector()).map { chapterFromElement(it) }
        } catch (e: IndexOutOfBoundsException) {
            emptyList()
```

src/en/dynasty/src/eu/kanade/tachiyomi/extension/en/dynasty/DynastyDoujins.kt
```diff
@@ -64,7 +74,31 @@ class DynastyDoujins : DynastyScans() {
...
+        while (hasNextPage) {
+            page += 1
+            val doujinURL = document.location() + "?page=$page"
 
+            val newRequest = GET(doujinURL, headers)
+            val newResponse = client.newCall(newRequest).execute()
+
+            if (!newResponse.isSuccessful) {
+                retry += 1
+                if (retry > maxRetry) { break } else { continue }
```
stevenyomi: If you `continue`, you'll hit `page += 1` and actually request the next page. Are you sure about this?

rtlow: You are correct, this will lead to unintended behavior. The latest commit now completely aborts chapter parsing if the max retries are reached.

src/en/dynasty/src/eu/kanade/tachiyomi/extension/en/dynasty/DynastyDoujins.kt
```diff
@@ -64,7 +74,38 @@ class DynastyDoujins : DynastyScans() {
...
+                retry += 1
+                if (retry > maxRetry) { break }
+                else {
+                    /*
+                    TODO: Toast to notify max retries exceeded
+                          and chapter parsing aborted
+                     */
+                    return chapters
+                }
+            }
```
stevenyomi:
```kotlin
if (!newResponse.isSuccessful) {
    retry += 1
    if (retry > maxRetry) { break } // THIS is where max retries exceeded
    else { // THIS is where you are gonna retry
        /*
        TODO: Toast to notify max retries exceeded
              and chapter parsing aborted
         */
        return chapters
    }
}
```

I think you got it wrong. What I meant was, when you're gonna retry, you continue and hit page++ and actually "retry" on the next page. Currently you're just aborting on `retry = 1`. I think you should make a function to handle retrying, or just yeet the entire retrying logic and just toast and abort.

rtlow: We'll just do the KISS option for now and remove the retry logic, especially since I don't have a better plan in mind at the moment.

As a side note, I do not know if [advancing to an unintended page count](https://dynasty-scans.com/doujins/touhou_project?page=666&view=chapters) produces any unintended behavior.

stevenyomi: I mean, if you get an HTTP 404 when requesting page 3, on the next loop you'll request page 4. Normal retrying logic should be requesting page 3. If skipping page 3 is correct, this is not "retrying", but "skipping error pages" - where the max retries count doesn't make sense.

stevenyomi (approved):

src/en/dynasty/src/eu/kanade/tachiyomi/extension/en/dynasty/DynastyDoujins.kt
rtlow: I see what you're trying to get at now. Here's what I was going for. I'm not worried about 404 errors since the URL always produces a valid page. Rather, Dynasty has a tendency to temporarily return 50X errors for a few seconds up to minutes at a time. I wanted to retry the current page in this case until either it resolves or we run out of retry attempts. If we run out, try to march through the rest of the pages without continuing to retry. I think the way I implemented it before would have caused unintended behavior if the page it 50X's on is actually the last page, and with no kind of back off it would hammer the site potentially causing additional problems on their end.

PR #162, opened 2024-01-12
Add Anchira

### LINKED ISSUE #29: Anchira : Source Request
### Source name

Anchira

### Source link

https://anchira.to/

### Source language

English

### Other details

NSFW = Yes

### Acknowledgements

- [X] I have checked that the extension does not already exist by searching the [GitHub repository](https://github.com/keiyoushi/extensions-source/) and verified it does not appear in the code base.

LetrixZ:
Closes #29 

Checklist:


src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/AnchiraHelper.kt
```diff
@@ -0,0 +1,33 @@
+package eu.kanade.tachiyomi.extension.en.anchira
+
+import kotlinx.serialization.decodeFromString
+import kotlinx.serialization.json.Json
+import okhttp3.ResponseBody
+import okio.ByteString.Companion.decodeBase64
+
+object AnchiraHelper {
+    const val KEY = "fuck niggers and faggots, and death to all jews"
```
beer-psi: Will do a full review on my laptop later, but I recommend obfuscating this key somehow (even a simple base64 encode would suffice) so the repo doesn't potentially get nuked over a hentai extension.

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
```diff
@@ -0,0 +1,342 @@
...
+                    SManga.create().apply {
+                        url = "/g/${it.id}/${it.key}"
+                        title = it.title
+                        thumbnail_url = "$cdnUrl/${it.id}/${it.key}/m/${it.cover.name}"
+                        artist = it.tags.filter { it.namespace == 1 }.joinToString(", ") { it.name }
+                        author = it.tags.filter { it.namespace == 2 }.joinToString(", ") { it.name }
+                        genre = prepareTags(it.tags)
+                        update_strategy = UpdateStrategy.ONLY_FETCH_ONCE
+                        status = SManga.COMPLETED
+                        initialized = false
```
beer-psi: `SManga.initialized` is false by default, so this is unneeded: https://github.com/tachiyomiorg/tachiyomi/blob/master/source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/model/SMangaImpl.kt#L23

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
```diff
@@ -0,0 +1,342 @@
...
+                    }
+                }.toList(),
+                data.entries.size < data.total,
```
beer-psi: Would this not always return `true`, because the size of each page is necessarily always smaller than the total number of galleries?

If there is a fixed size for each page (say 25), I would recommend something like this instead:
```kotlin
(page + 1) * 25 < data.total
```

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
```diff
@@ -0,0 +1,342 @@
...
+            )
+        } else {
+            MangasPage(listOf(), false)
```
beer-psi: Also related to the previous comment, returning an empty list would display a confusing toast "No entries found" in app. It's better to get `MangasPage.hasNextPage` correct the first time around.

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
```diff
@@ -0,0 +1,342 @@
...
+        filters.forEach { filter ->
+            when (filter) {
+                is CategoryGroup -> {
+                    var sum = 0
+
+                    filter.state.forEach { category ->
+                        when (category.name) {
+                            "Manga" -> if (category.state) sum += 1
+                            "Doujinshi" -> if (category.state) sum += 2
+                            "Illustration" -> if (category.state) sum += 4
```
beer-psi:
```suggestion
                            "Manga" -> if (category.state) sum = sum or 1
                            "Doujinshi" -> if (category.state) sum = sum or 2
                            "Illustration" -> if (category.state) sum = sum or 4
```

Maybe use bitwise OR for these instead. Adding also works, but it is better to use bitwise OR for bit flags.

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
```diff
@@ -0,0 +1,342 @@
...
+            artist = data.tags.filter { it.namespace == 1 }.joinToString(", ") { it.name }
+            author = data.tags.filter { it.namespace == 2 }.joinToString(", ") { it.name }
+            genre = prepareTags(data.tags)
+            update_strategy = UpdateStrategy.ONLY_FETCH_ONCE
+            status = SManga.COMPLETED
+        }
+    }
+
+    override fun getMangaUrl(manga: SManga) = if (preferences.openSource) {
+        val id = manga.url.split("/").reversed()[1].toInt()
```
beer-psi, line 173:
```suggestion
        val id = manga.url.split("/").last().toInt()
```

Use [`kotlin.collections.last`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/last.html), which is more idiomatic Kotlin.

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/AnchiraDto.kt
```diff
@@ -0,0 +1,57 @@
+package eu.kanade.tachiyomi.extension.en.anchira
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class ListEntry(
+    @SerialName("id") val id: Int,
+    @SerialName("key") val key: String,
```
beer-psi: Are these `@SerialName` decorators needed? They seem like the same as the property names.

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/AnchiraDto.kt
```diff
@@ -0,0 +1,57 @@
...
+@Serializable
+data class LibraryResponse(
+    val entries: List<ListEntry>? = null,
+    val total: Int,
+)
+
+@Serializable
+data class Entry(
+    @SerialName("id") val id: Int,
+    @SerialName("key") val key: String,
```
beer-psi: Same comment as above.

beer-psi (changes requested):
Insane job figuring out how the website works, though I have a few (maybe nitpicky) comments.

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
LetrixZ: I need the 2nd to last here `/g/{ID}/{KEY}` which would be the ID.

beer-psi: Ah, my bad, confused it for `.reversed()[0]`.

src/en/anchira/build.gradle
```diff
@@ -0,0 +1,17 @@
...
+ext {
+    extName = 'Anchira'
+    pkgNameSuffix = 'en.anchira'
+    extClass = '.Anchira'
+    extVersionCode = 1
+    isNsfw = true
+}
+
+dependencies {
+    implementation "com.ensarsarajcic.kotlinx:serialization-msgpack:0.5.5"
```
beer-psi: This seems unused. Is there anything msgpack-encoded?

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/AnchiraDto.kt
LetrixZ: These were leftovers from when they had a different name.

src/en/anchira/build.gradle
LetrixZ: Another leftover. The API originally was encoded with MessagePack.

beer-psi (approved):
Thanks for the work. Will merge when CI passes. Let's see how long this lasts.

PR #259, opened 2024-01-15
Address some gradle warnings

Claudemirovsky:


settings.gradle.kts
```diff
@@ -7,6 +7,20 @@ File(rootDir, "lib").eachDir {
...
+    ":extensions:individual" to "$rootDir/src",
+    ":extensions:multisrc" to "$rootDir/generated-src",
+).forEach { (name, path) ->
+    val projectDir = file(path)
+    if (projectDir.exists()) {
+        include(name)
+        project(name).projectDir = projectDir
+    }
+}
+
```
stevenyomi: I think we'll rename the projects after we migrate to the new multisrc structure so this is probably not needed.

NB: we can probably just conform to the Gradle default and use directory paths as project names, i.e. `:lib:dataimage` instead of `:lib-dataimage`.

core/build.gradle.kts
```diff
@@ -9,7 +9,7 @@ android {
         minSdk = AndroidConfig.minSdk
     }
 
-    namespace = "eu.kanade.tachiyomi.extension"
+    namespace = "eu.kanade.tachiyomi.lib.core"
```
stevenyomi: `core` is actually not a lib. Can we use `eu.kanade.tachiyomi.extension.core`? 

Also, does this affect the merging of `AndroidManifest.xml` and default icon resources?

Claudemirovsky: No, it doesn't affect much. Extensions built with the renamed `core` namespace are going to behave as usual (thanks god).
Anyway, I'll apply your suggestion.

settings.gradle.kts
Claudemirovsky: Oooh that's good news, didn't know about that. In that case, should I revert this part, or keep it until the multisrc refactoration + project renaming steps are complete?

stevenyomi: Just revert this part.

Claudemirovsky: Done.

stevenyomi (approved):

PR #281, opened 2024-01-15
Move plugin section to common.gradle

AwkwardPeak7:


common.gradle
```diff
@@ -1,4 +1,7 @@
 apply plugin: 'org.jmailen.kotlinter'
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlinx-serialization'
```
stevenyomi, line 3: Nit: add these three lines before kotlinter.

stevenyomi: Or probably do
```groovy
plugin {
    id 'xxx'
}
```

src/ru/newbie/build.gradle
```diff
@@ -1,16 +1,12 @@
...
 }
 
+apply from: "$rootDir/common.gradle"
+
 dependencies {
     implementation project(':lib-dataimage')
-}
-
-apply from: "$rootDir/common.gradle"
+}
```
stevenyomi: Add a new line here.

stevenyomi (approved):

PR #336, opened 2024-01-18
TMO & LectorManga: Ignore ssl in the problematic CDN

bapeey:
I'll make multisrc after theme-lib

Checklist:

### DIFF
--- a/src/es/lectormanga/src/eu/kanade/tachiyomi/extension/es/lectormanga/LectorManga.kt
+++ b/src/es/lectormanga/src/eu/kanade/tachiyomi/extension/es/lectormanga/LectorManga.kt
@@ -1,15 +1,12 @@
 package eu.kanade.tachiyomi.extension.es.lectormanga
 
+import android.annotation.SuppressLint
 import android.app.Application
 import android.content.SharedPreferences
-import android.os.Handler
-import android.os.Looper
-import android.view.View
-import android.webkit.WebView
-import android.webkit.WebViewClient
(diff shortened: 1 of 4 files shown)


src/es/lectormanga/src/eu/kanade/tachiyomi/extension/es/lectormanga/LectorManga.kt
```diff
@@ -71,8 +72,42 @@ class LectorManga : ConfigurableSource, ParsedHttpSource() {
...
+            preferences.getString(IMAGE_CDN_RATELIMIT_PREF, IMAGE_CDN_RATELIMIT_PREF_DEFAULT_VALUE)!!.toInt(),
+            60,
+        )
+        .build()
+
     override val client: OkHttpClient = network.client.newBuilder()
+        .addInterceptor { chain ->
+            val request = chain.request()
+            val url = request.url
+            if (url.host.contains("japanreader.com")) {
```
AwkwardPeak7, line 106: maybe ignore ssl for all cdns on these sources considering they are often reported because of SSL issues on their cdn

bapeey: the issue is only for that host, all others are fine, 
or maybe they only use that cdn now,  the previous solution only loaded the webview for that host

AwkwardPeak7 (approved):

PR #440, opened 2024-01-21
Wicked Witch Scan: Add new website

### LINKED ISSUE #129: Wicked Witch Scan changed URL
### Source information

Wicked Witch Scan 1.4.32

### Source language

Português (Brasil)

### Source new URL

https://wicked-witch-scan.com/

### Other details

_No response_

### Acknowledgements

- [X] I have opened WebView and checked that the source URL is not updated yet.

beer-psi:
Closes #129 

~~Since their old website is still active and not all content has been migrated over, Wicked Witch Scan is now a source factory containing "Wicked Witch Scan" and "Wicked Witch Scan (Novo)". The old source will be removed when the team shuts down the old website.~~

Checklist:


vetleledaal:
Not sure about appending "(Novo)", would cause issues next time they switch sites. Renaming the old one isn't great either, as downloads would have to be moved.

beer-psi:
> Not sure about appending "(Novo)", would cause issues next time they switch sites. Renaming the old one isn't great either, as downloads would have to be moved.

Need something to differentiate the sources on the user side temporarily though. Would making the displayed name "Wicked Witch Scan (Novo)", but the name used for ID generation just "Wicked Witch Scan" (with `versionId = 2` of course) work?

AwkwardPeak7:
I think you should just add their new site as a new separate extension considering it doesn't use Madara. Old one can be removed later

beer-psi:
> I think you should just add their new site as a new separate extension considering it doesn't use Madara. Old one can be removed later

What would the naming be for the new package then? Icon is probably not a problem because the old Madara website used the default icon. If we used `pt.wickedwitchscannew`, that'd cause issues if they switch sites again, like Vetle said.

AwkwardPeak7:
pt.wickedwitchscansnovo?  Package name can be kept if they change site again

beer-psi:
> pt.wickedwitchscansnovo? Package name can be kept if they change site again

There might be a chance of something like this (i.e. old and new coexisting) happening again though. 

> I think you should just add their new site as a new separate extension considering it doesn't use Madara.

It's stuck in Madara because of the old site. When the old site goes offline, this will become an individual extension.

beer-psi:
There's still the issue of the downloads folder conflicting though. Not sure how to move forward.

beer-psi:
I think I'm moving forward with splitting the new website into an individual extension (`pt.wickedwitchscannovo`). If something like this happens again we can probably recycle `pt.wickedwitchscan` by then.

src/pt/wickedwitchscannovo/src/eu/kanade/tachiyomi/extension/pt/wickedwitchscannovo/WickedWitchScan.kt
```diff
@@ -0,0 +1,289 @@
...
+import java.text.SimpleDateFormat
+import java.util.Locale
+import java.util.TimeZone
+import java.util.concurrent.TimeUnit
+import java.util.zip.ZipInputStream
+
+@SuppressLint("WrongConstant")
+class WickedWitchScan : ParsedHttpSource() {
+
+    override val name = "Wicked Witch Scan"
```
AwkwardPeak7, line 48: change the name here as id should never be same for two sources

src/pt/wickedwitchscannovo/src/eu/kanade/tachiyomi/extension/pt/wickedwitchscannovo/WickedWitchScan.kt
```diff
@@ -0,0 +1,289 @@
...
+            addPathSegments("auto-complete/")
+            addQueryParameter("term", query)
+        }.build()
+
+        return GET(url, headers)
+    }
+
+    override fun searchMangaParse(response: Response): MangasPage {
+        val manga = json.parseToJsonElement(response.body.string()).jsonArray.map {
+            val element = Jsoup.parse(it.jsonObject["html"]!!.jsonPrimitive.content)
```
AwkwardPeak7:
```suggestion
            val element = Jsoup.parseBodyFragment(it.jsonObject["html"]!!.jsonPrimitive.content)
```

src/pt/wickedwitchscannovo/src/eu/kanade/tachiyomi/extension/pt/wickedwitchscannovo/WickedWitchScan.kt
```diff
@@ -0,0 +1,289 @@
...
+
+        val images = generateSequence { zis.nextEntry }
+            .map {
+                val entryName = it.name
+                val splitEntryName = entryName.split('.')
+                val entryIndex = splitEntryName.first().toInt()
+                val entryType = splitEntryName.last()
+
+                val imageData = if (entryType == "avif") {
+                    Log.d("wickedwitch", "Extracting AVIF image $entryName")
```
AwkwardPeak7: remove logs

AwkwardPeak7 (changes requested):

src/pt/wickedwitchscannovo/src/eu/kanade/tachiyomi/extension/pt/wickedwitchscannovo/WickedWitchScan.kt
beer-psi: I already bumped the versionId though?

```kotlin
    // Source changed from Madara to homegrown website
    override val versionId = 2
```

AwkwardPeak7: Oh right nvm then 🫠

AwkwardPeak7 (approved):

PR #454, opened 2024-01-21
komiktap/id: added sucuriInterceptor

Sofie99:
Checklist:

### DIFF
--- a/multisrc/overrides/mangathemesia/komiktap/src/Komiktap.kt
+++ b/multisrc/overrides/mangathemesia/komiktap/src/Komiktap.kt
@@ -1,11 +1,49 @@
 package eu.kanade.tachiyomi.extension.id.komiktap
 
+import app.cash.quickjs.QuickJs
 import eu.kanade.tachiyomi.multisrc.mangathemesia.MangaThemesia
-import eu.kanade.tachiyomi.network.interceptor.rateLimit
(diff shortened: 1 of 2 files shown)


multisrc/overrides/mangathemesia/komiktap/src/Komiktap.kt
```diff
@@ -1,11 +1,49 @@
...
+import okhttp3.Cookie
+import okhttp3.Interceptor
+import okhttp3.Response
+import java.io.IOException
 
 class Komiktap : MangaThemesia("Komiktap", "https://komiktap.me", "id") {
-    override val client: OkHttpClient = super.client.newBuilder()
-        .rateLimit(4)
-        .build()
+    override val client by lazy {
```
beer-psi: Making the client lazy is unnecessary since it will be used as soon as the user enters the extension anyways. Also, why was the rate limit removed?

beer-psi (changes requested):

multisrc/overrides/mangathemesia/komiktap/src/Komiktap.kt
Sofie99: never needed in the first place. no idea why it there

AwkwardPeak7 (approved):

PR #579, opened 2024-01-24
Add Komga

### LINKED ISSUE #225: Add back Komga
### Source name

Komga

### Source link

https://demo.komga.org/

### Source language

English

### Other details

Add back Komga and the other self hosted sources here now that official extension repo has been archived. 

### Acknowledgements

- [X] I have checked that the extension does not already exist by searching the [GitHub repository](https://github.com/keiyoushi/extensions-source/) and verified it does not appear in the code base.

beer-psi:
Closes #225 

Changes from the official extension:
- Allowed some customization of the chapter name
- Allowed changing the number of additional Komga sources created (max 10). One source will always be created.

Checklist:


THElegend5:
Would it be possible to add ability to remap base url? Basically this https://web.archive.org/web/20221102194619/https://github.com/tachiyomiorg/tachiyomi-extensions/issues/13749

THElegend5:
> Would it be possible to add ability to remap base url? Basically this https://web.archive.org/web/20221102194619/https://github.com/tachiyomiorg/tachiyomi-extensions/issues/13749

There was more to this conversation along with some possible ways to do so but they are lost now as far as I can tell.

beer-psi:
> > Would it be possible to add ability to remap base url? Basically this https://web.archive.org/web/20221102194619/https://github.com/tachiyomiorg/tachiyomi-extensions/issues/13749
> 
> There was more to this conversation along with some possible ways to do so but they are lost now as far as I can tell.

I guess the extension can keep its own internal list of previous base URLs? I can't think of anything elegant to work with currently.

THElegend5:
> > > Would it be possible to add ability to remap base url? Basically this https://web.archive.org/web/20221102194619/https://github.com/tachiyomiorg/tachiyomi-extensions/issues/13749
> > 
> > There was more to this conversation along with some possible ways to do so but they are lost now as far as I can tell.
> 
> I guess the extension can keep its own internal list of previous base URLs? I can't think of anything elegant to work with currently.

I think what the proposed solution was to save something like 1/series-url in the app database and the extension replaced the 1 with the base url when searching for stuff and other things,  the number is so as to differentiate between which komga extension you are using as there can be multiple. 
The issue I can think with this might be caused by how this method would work with the komga tracker.

AwkwardPeak7 (approved):

beer-psi:
> I think what the proposed solution was to save something like 1/series-url in the app database and the extension replaced the 1 with the base url when searching for stuff and other things, the number is so as to differentiate between which komga extension you are using as there can be multiple. The issue I can think with this might be caused by how this method would work with the komga tracker.

I think this works as an idea, however implementing it will break two-way tracking built into Mihon/Tachiyomi. I also doubt we have the power to influence any app changes. Feel free to make a feature request for this feature however.

jindalpratik:
> > I think what the proposed solution was to save something like 1/series-url in the app database and the extension replaced the 1 with the base url when searching for stuff and other things, the number is so as to differentiate between which komga extension you are using as there can be multiple. The issue I can think with this might be caused by how this method would work with the komga tracker.
> 
> I think this works as an idea, however implementing it will break two-way tracking built into Mihon/Tachiyomi. I also doubt we have the power to influence any app changes. Feel free to make a feature request for this feature however.

There were already talks to do this in old tachi after a komga refactor that was in works but am unable to find anywhere anymore so it might be possible they do something about it, I'll open a FR for this here.

THElegend5:
> > I think what the proposed solution was to save something like 1/series-url in the app database and the extension replaced the 1 with the base url when searching for stuff and other things, the number is so as to differentiate between which komga extension you are using as there can be multiple. The issue I can think with this might be caused by how this method would work with the komga tracker.
> 
> I think this works as an idea, however implementing it will break two-way tracking built into Mihon/Tachiyomi. I also doubt we have the power to influence any app changes. Feel free to make a feature request for this feature however.

There were already talks to do this in old tachi after a komga refactor that was in works but am unable to find anywhere anymore so it might be possible they do something about it, I'll open a FR for this here.

PR #686, opened 2024-01-27
Add BlogTruyen.vn (unoriginal)

beer-psi:
Both sites are practically the same thing, but owned by different people. The owner of the real BlogTruyen had to sell it off in order to avoid legal issues, but then set up another site with the same code, just with a different theme color.

Checklist:


multisrc/overrides/blogtruyen/blogtruyenvn/AndroidManifest.xml
AwkwardPeak7, line 1: you have a similar manifest file in default override, so why duplicate it?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/blogtruyen/BlogTruyen.kt
```diff
@@ -0,0 +1,383 @@
...
+        return pages
+    }
+
+    override fun imageUrlParse(document: Document) = throw UnsupportedOperationException()
+
+    override fun getFilterList() = FilterList(
+        Author(),
+        Scanlator(),
+        Status(),
+        GenreList(getGenreList()),
```
AwkwardPeak7: nitpick: don't show `GenreList` filter if `getGenreList()` is empty

AwkwardPeak7 (changes requested):

multisrc/overrides/blogtruyen/blogtruyenvn/AndroidManifest.xml
beer-psi: It was originally because BlogTruyen.vn still owned `blogtruyen.vn`:
```xml
                <data android:host="blogtruyen.vn" />
                <data android:host="m.blogtruyen.vn" />
```

But I just checked again and `blogtruyen.vn` redirects to `blogtruyenvn.com` anyways.

AwkwardPeak7 (approved):

PR #776, opened 2024-01-29
Restore Hikari Scan

SnakeDoc83:
Closes #773 

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [x] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [x] Have removed `web_hi_res_512.png` when adding a new extension

PR #814, opened 2024-01-30
Clean up and remove misplaced dead overrides

stevenyomi:
Some changes are lint fixes for the multisrc migration.

Fixed misplaced Hikari Scan override files by #776.

Checklist:


multisrc/overrides/madara/pikiranwibu/src/PikiranWibu.kt
AwkwardPeak7, line 1: I just checked, this site seems to be dead/link to mgkomik.id which exists as separate extension.

stevenyomi: True, it's become a blog site.

AwkwardPeak7 (approved):

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/sinmh/ProgressiveParser.kt
```diff
@@ -1,6 +1,6 @@
 package eu.kanade.tachiyomi.multisrc.sinmh
 
-internal class ProgressiveParser(private val text: String) {
+class ProgressiveParser(private val text: String) {
```
beer-psi, line 3: was there a need to change the visibility of this class?

multisrc/overrides/bilibili/bilibilimanga/src/BilibiliManga.kt
```diff
@@ -27,9 +27,15 @@ class BilibiliManga : Bilibili(
...
-            .filter { episode -> episode.isInFree || !episode.isLocked }
-            .map { ep -> chapterFromObject(ep, result.data.id) }
+        val data = result.data!!
+        val id = data.id
+        return data.episodeList.mapNotNull { episode ->
+            if (episode.isInFree || !episode.isLocked) {
+                chapterFromObject(episode, id)
+            } else {
+                null
+            }
```
beer-psi, line 37: I know this is probably more efficient because the collection only needs to be iterated once, but a `.filter .map` pair is more FP-like. Only a suggestion though, ignore if disagree.

beer-psi:
Guess I was late to this. Oh well, Ace Attorney was pretty good.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/sinmh/ProgressiveParser.kt
stevenyomi: Yeah, it won't be visible when the theme becomes a lib.

multisrc/overrides/bilibili/bilibilimanga/src/BilibiliManga.kt
stevenyomi: True. I just conveniently rewrote this because the `result.data!! -> result.data` smart cast won't be possible.

PR #931, opened 2024-02-02
Rewrite MMRCMS

beer-psi:
### Changelog
- Remove dead sources (Scan FR, Phoenix-Scans, Manga-FR, HentaiShark)
- Mark Mangadoor as NSFW
- Remove use of `MMRCMSJsonGen` and `SourceData` in favor of dynamically fetching genres a la Madara
- Enable advanced search for sources that allow it
- Add pagination for searching by query (the search endpoint dumps all results at once)
- General clean-up

Checklist:


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
```diff
@@ -1,529 +1,487 @@
...
+     * The path used in the URL for the manga pages. Can be
+     * changed if needed as some sites modify it to other words.
      */
-    private val json: Json by injectLazy()
-    val jsonObject = json.decodeFromString<JsonObject>(jsonData)
-    override val supportsLatest = jsonObject["supports_latest"]!!.jsonPrimitive.boolean
-    open val itemUrl = jsonObject["item_url"]!!.jsonPrimitive.content
-    open val categoryMappings = mapToPairs(jsonObject["categories"]!!.jsonArray)
-    open var tagMappings = jsonObject["tags"]?.jsonArray?.let { mapToPairs(it) } ?: emptyList()
+    protected open val itemPath = "manga"
```
stevenyomi: I'm not a fan of adding a bunch of `open val`s (i.e. the current approach of Madara and MangaThemesia). These configurations are hard to discover and less performant. How about moving them into the constructor like the date format?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
```diff
@@ -1,529 +1,487 @@
...
+                    thumbnail_url = MMRCMSUtils.guessCover(baseUrl, url, null)
+                },
+            )
         }
 
-        return out
+        return MangasPage(manga, endRange < searchDirectory.lastIndex)
     }
 
+    protected open val detailAuthor = setOf("author(s)", "autor(es)", "auteur(s)", "著作", "yazar(lar)", "mangaka(lar)", "pengarang/penulis", "pengarang", "penulis", "autor", "المؤلف", "перевод", "autor/autorzy")
```
stevenyomi: You can use `HashSet` to save a little bit overhead.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
```diff
@@ -1,529 +1,487 @@
...
+            Page(i, imageUrl = it.imgAttr())
+        }
+
+    override fun imageUrlParse(document: Document) = throw UnsupportedOperationException()
 
-            // Parse date
-            val dateText = element.getElementsByClass("date-chapter-title-rtl").text().trim()
-            chapter.date_upload = parseDate(dateText)
+    override fun getFilterList(): FilterList {
+        val filters = mutableListOf<Filter<*>>().apply {
```
stevenyomi: `buildList`?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
```diff
@@ -1,529 +1,487 @@
...
 
-        companion object {
-            private val sortables = arrayOf(
-                "name" to "Name",
-                "views" to "Popularity",
-                "last_release" to "Last update",
-            )
-        }
-    }
+            val result = runCatching {
```
stevenyomi: Just use `try`.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
```diff
@@ -1,529 +1,487 @@
...
-            "Sort",
-            sortables.map { it.second }.toTypedArray(),
-            Selection(0, true),
-        ),
-        UriFilter {
-        override fun addToUri(uri: Uri.Builder) {
-            uri.appendQueryParameter("sortBy", sortables[state!!.index].first)
-            uri.appendQueryParameter("asc", state!!.ascending.toString())
-        }
+            if (fetchFiltersAttempts > 3 || (fetchFiltersAttempts > 0 && !fetchFiltersFailed)) {
```
stevenyomi: Should check this condition before locking too, I guess.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
```diff
@@ -1,529 +1,487 @@
...
-        override fun addToUri(uri: Uri.Builder) {
-            if (state != 0 || !firstIsUnspecified) {
-                uri.appendQueryParameter(uriParam, vals[state].first)
+    private var categories = emptyList<Pair<String, String>>()
+
+    private var statuses = emptyList<Pair<String, String>>()
+
+    private var types = emptyList<Pair<String, String>>()
+
+    private var tags = emptyList<Pair<String, String>>()
```
stevenyomi, line 420: Could you combine these into a single `HashMap`? Plus, is it possible to infer the filters from the page instead of hardcoding them?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
beer-psi: I intentionally went with `runCatching` to reduce nesting and get a nice `result.isFailure` at the end.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
beer-psi: That was originally how I wrote it, but I realized these may be mutated by a previous call to `fetchFilterOptions()` if the user is quick enough.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
beer-psi: I'm not a fan of having too many things inside a constructor/function call, unless there's a way to "enforce" passing named arguments a la Python's `kwargs`.

beer-psi: There's apparently a hack using `varargs`:
```kotlin
public class Forbidden private constructor()

abstract class MMRCMS(
    override val name: String,
    override val baseUrl: String,
    final override val lang: String,
    vararg forbiddens: Forbidden,
    private val dateFormat: SimpleDateFormat = SimpleDateFormat("d MMM. yyyy", Locale.US),
) : ParsedHttpSource()
```

stevenyomi: I think you mean the `/` and `*`. I think you can hack it in Kotlin by inserting a class with private constructor lol.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
stevenyomi:
```Kotlin
fetchFiltersFailed = try {
    // ...
    false
} catch (e: Throwable) {
    // Log.e
    true
}
```

`runCatching` has the overhead of boxing something into a `Result`, and functional chain might lead to very complicated logic so I generally avoid it.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
stevenyomi: Edit: A regular map for insert order, sorry.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
```diff
@@ -1,529 +1,474 @@
...
-     *
-     *
-     */
-    private val json: Json by injectLazy()
-    val jsonObject = json.decodeFromString<JsonObject>(jsonData)
-    override val supportsLatest = jsonObject["supports_latest"]!!.jsonPrimitive.boolean
-    open val itemUrl = jsonObject["item_url"]!!.jsonPrimitive.content
-    open val categoryMappings = mapToPairs(jsonObject["categories"]!!.jsonArray)
-    open var tagMappings = jsonObject["tags"]?.jsonArray?.let { mapToPairs(it) } ?: emptyList()
+    vararg forbiddens: Forbidden,
```
stevenyomi: Probably rename to `UseNamedArgumentsBelow` so it's easier to understand.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
beer-psi: Wouldn't this need to be a sequence anyway for the purpose of creating `Filter`s?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
```diff
@@ -1,529 +1,474 @@
...
+    protected val itemPath: String = "manga",
+    protected val fetchFilterOptions: Boolean = true,
+    protected val supportsAdvancedSearch: Boolean = true,
+    protected val detailsTitleSelector: String = ".listmanga-header, .widget-title",
+    protected val chapterNamePrefix: String = "",
+    protected val chapterString: String = when (lang) {
+        "es" -> "Capítulo"
+        "fr" -> "Chapitre"
+        else -> "Chapter"
+    },
```
stevenyomi: Make them `private`, otherwise it defeats the purpose of moving them into the constructor. ~~Subclasses should definitely have no need to override them anymore.~~ Sorry I mistook them as `open`.

beer-psi: Keeping `itemPath` `protected` for Mangas.in.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
stevenyomi: Never mind, I actually mean if the filters have the same HTML structure, but it looks like your logic is more complicated.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
```diff
@@ -1,529 +1,476 @@
...
+        return MangasPage(manga, endRange < searchDirectory.lastIndex)
     }
 
+    protected open val detailAuthor = hashSetOf("author(s)", "autor(es)", "auteur(s)", "著作", "yazar(lar)", "mangaka(lar)", "pengarang/penulis", "pengarang", "penulis", "autor", "المؤلف", "перевод", "autor/autorzy")
+    protected open val detailArtist = hashSetOf("artist(s)", "artiste(s)", "sanatçi(lar)", "artista(s)", "artist(s)/ilustrator", "الرسام", "seniman", "rysownik/rysownicy", "artista")
+    protected open val detailGenre = hashSetOf("categories", "categorías", "catégories", "ジャンル", "kategoriler", "categorias", "kategorie", "التصنيفات", "жанр", "kategori", "tagi", "género")
+    protected open val detailStatus = hashSetOf("status", "statut", "estado", "状態", "durum", "الحالة", "статус")
+    protected open val detailStatusComplete = hashSetOf("complete", "مكتملة", "complet", "completo", "zakończone", "concluído", "finalizado")
+    protected open val detailStatusOngoing = hashSetOf("ongoing", "مستمرة", "en cours", "em lançamento", "prace w toku", "ativo", "em andamento", "activo")
+    protected open val detailStatusDropped = hashSetOf("dropped")
```
stevenyomi: Any reason why these are `open`?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
stevenyomi: > I'm not a fan of having too many things inside a constructor/function call

Think data class constructors and Composable functions.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMS.kt
beer-psi: I copied Madara. `protected` because it's also used by Onma but there's probably no point to it being `open`.

multisrc/overrides/mmrcms/lelscanvf/src/LelscanVF.kt
```diff
@@ -0,0 +1,7 @@
+package eu.kanade.tachiyomi.extension.fr.lelscanvf
+
+import eu.kanade.tachiyomi.multisrc.mmrcms.MMRCMS
+
+class LelscanVF : MMRCMS("Lelscan-VF", "https://lelscanvf.cc", "fr") {
+    override val supportsAdvancedSearch = false
```
stevenyomi:
```
e: /home/runner/work/extensions-source/extensions-source/generated-src/fr/lelscanvf/src/eu/kanade/tachiyomi/extension/fr/lelscanvf/LelscanVF.kt: (6, 5): 'public open val supportsAdvancedSearch: Boolean defined in eu.kanade.tachiyomi.extension.fr.lelscanvf.LelscanVF' has no access to 'private final val supportsAdvancedSearch: Boolean defined in eu.kanade.tachiyomi.multisrc.mmrcms.MMRCMS', so it cannot override it
```

multisrc/overrides/mmrcms/mangascan/src/MangaScan.kt
```diff
@@ -3,21 +3,16 @@ package eu.kanade.tachiyomi.extension.fr.mangascan
...
 
 class MangaScan : MMRCMS("Manga-Scan", "https://mangascan-fr.com", "fr") {
+    override val supportsAdvancedSearch = false
 
-    override fun mangaDetailsParse(response: Response): SManga {
-        return super.mangaDetailsParse(response).apply {
-            title = title.substringBefore("Chapitres en ligne").substringAfter("Scan").trim()
-        }
-    }
+    override val detailsTitleSelector = "div.col-sm-12 h1"
```
stevenyomi: This too.

stevenyomi (approved):

PR #1015, opened 2024-02-05
Mark NSFW: Bakaman

cuong-tran:
Since it only mark NSFW so I don't think it worth update version

Checklist:


beer-psi:
You need to bump the version no matter how small a change it may be.

I'll also have to ask if CoffeeManga and MangaFoxFull really deserves the NSFW flag, since there's very few entries tagged Adult in the first place (~40 out of 2000).

Asking @stevenyomi to weigh in on ManhuaGui.

cuong-tran:
> You need to bump the version no matter how small a change it may be.
> 
> I'll also have to ask if CoffeeManga and MangaFoxFull really deserves the NSFW flag, since there's very few entries tagged Adult in the first place (~40 out of 2000).
> 
> Asking @stevenyomi to weigh in on ManhuaGui.

isn't NSFW flag a keep-out indicator, for kids?

beer-psi:
> isn't NSFW flag a keep-out indicator, for kids?

No? The NSFW flag is both a warning and a sign, depending on the specific person. If every source with even the slightest hint of NSFW was marked as NSFW the tag would become meaningless.

Bakaman is legitimate, there's a sizeable number of NSFW entries on there, I hit a few just browsing the home page, however...
- You can't find NSFW content on CoffeeManga/MangaFoxFull unless you intentionally go out and look for them, say, by filtering with the Adult tag.
- For ManhuaGui, NSFW content may show up in search results, but the extension will refuse to load details and chapters unless you enable the specific extension setting for it.

cuong-tran:
So final decision will be Bakaman only? If yes then I will make the change.

beer-psi (approved):

PR #1023, opened 2024-02-05
Add Manhua AZ

cuong-tran:
It's a spinned off from Manhua ES so I keep its version same as Manhua ES

Checklist:


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/MadaraGenerator.kt
```diff
@@ -282,6 +282,7 @@ class MadaraGenerator : ThemeSourceGenerator {
         SingleLang("Mangá Nanquim", "https://mangananquim.com", "pt-BR", className = "MangaNanquim"),
         SingleLang("Manhastro", "https://manhastro.com", "pt-BR"),
         SingleLang("Manhatic", "https://manhatic.com", "ar", isNsfw = true),
+        SingleLang("Manhua AZ", "https://manhuaaz.com", "en", overrideVersionCode = 6),
```
beer-psi:
```suggestion
        SingleLang("Manhua AZ", "https://manhuaaz.com", "en"),
```

> It's a spinned off from Manhua ES so I keep its version same as Manhua ES

This is not how it works. A new extension doesn't suddenly have the same version as an existing extension because they are sister sources.

beer-psi (changes requested):

AwkwardPeak7 (approved):

PR #1041, opened 2024-02-06
Fix hentai20 and moved it to mangathemesia!

devil6venom:
Checklist:

Any idea how can i add icon? Thanks


multisrc/overrides/mangathemesia/hentai20/src/hentai20.kt
beer-psi, line 1: You don't need to create an empty override, the multisrc build process will do that.

multisrc/overrides/madara/hentai20/res/mipmap-hdpi/ic_launcher.png
beer-psi, line 1: Why don't you move the icons over instead of outright deleting them?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mangathemesia/MangaThemesiaGenerator.kt
```diff
@@ -48,6 +48,7 @@ class MangaThemesiaGenerator : ThemeSourceGenerator {
         SingleLang("Glory Scans", "https://gloryscans.fr", "fr"),
         SingleLang("Gremory Mangas", "https://gremorymangas.com", "es"),
         SingleLang("Hanuman Scan", "https://hanumanscan.com", "en"),
+        SingleLang("Hentai20", "https://hentai20.io", "en", isNsfw = true, overrideVersionCode = 4),
```
beer-psi, line 51:
```suggestion
        SingleLang("Hentai20", "https://hentai20.io", "en", isNsfw = true, overrideVersionCode = 9),
```
This wouldn't be enough to trigger an extension update in-app.
- Madara's `baseVersionCode` is 33, `overrideVersionCode` is 3, so the version would be 36, which is currently the version on the repo: https://keiyoushi.github.io/extensions/#en.hentai20
- MangaThemesia's `baseVersionCode` is 28, so an `overrideVersionCode` of 4 would make the final version 32, which is a downgrade.

beer-psi (changes requested):

multisrc/overrides/mangathemesia/hentai20/src/hentai20.kt
devil6venom: Okay!!! So i just add it in Mangathemesia? And remove from madara then?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mangathemesia/MangaThemesiaGenerator.kt
devil6venom: Oh.... I understand! I will make new one then!

PR #1042, opened 2024-02-06
Add condition to only run codeberg_mirror on keiyoushi repo...

cuong-tran:
… instead of other forks

Checklist:

### DIFF
--- a/.github/workflows/codeberg_mirror.yml
+++ b/.github/workflows/codeberg_mirror.yml
@@ -10,14 +10,13 @@ on:
 
 jobs:
   codeberg:
+    if: "github.repository == 'keiyoushi/extensions-source'"
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


stevenyomi:
Could you apply the condition to the entire workflow instead of the individual steps?

beer-psi:
> Could you apply the condition to the entire workflow instead of the individual steps?

I think you can't apply it at workflow-level, only job-level.

stevenyomi:
Then apply it on the job? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

beer-psi:
> Then apply it on the job? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

That's what I suggested in the review.

stevenyomi (approved):

PR #1052, opened 2024-02-06
id/mangaku: added override baseUrl setting

Sofie99:
Checklist:

### DIFF
--- a/src/id/mangaku/src/eu/kanade/tachiyomi/extension/id/mangaku/Mangaku.kt
+++ b/src/id/mangaku/src/eu/kanade/tachiyomi/extension/id/mangaku/Mangaku.kt
@@ -12,7 +12,6 @@ import android.webkit.WebViewClient
 import android.widget.Toast
 import androidx.preference.EditTextPreference
 import androidx.preference.PreferenceScreen
-import eu.kanade.tachiyomi.AppInfo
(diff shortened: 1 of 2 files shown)


src/id/mangaku/src/eu/kanade/tachiyomi/extension/id/mangaku/Mangaku.kt
```diff
@@ -35,11 +41,11 @@ import uy.kohesive.injekt.injectLazy
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 
-class Mangaku : ParsedHttpSource() {
+class Mangaku : ParsedHttpSource(), ConfigurableSource {
 
     override val name = "Mangaku"
 
-    override val baseUrl = "https://mangaku.mom"
+    override val baseUrl by lazy { preferences.getString(PREF_DOMAIN_KEY, PREF_DOMAIN_DEFAULT)!! }
```
beer-psi: No need to be lazy since it'll be used immediately anyways.

beer-psi (changes requested):

stevenyomi:
Preference should not depend on either AppInfo or extension's BuildConfig.

Only update the base URL if you don't know how to handle it.

PR #1059, opened 2024-02-06
fix "impossible de controuner Cloudflare" Manga Starz

mohamedotaku:
Checklist:

### DIFF
--- a/multisrc/overrides/madara/mangastarz/src/MangaStarz.kt
+++ b/multisrc/overrides/madara/mangastarz/src/MangaStarz.kt
@@ -1,6 +1,17 @@
 package eu.kanade.tachiyomi.extension.ar.mangastarz
 
 import eu.kanade.tachiyomi.multisrc.madara.Madara
+import eu.kanade.tachiyomi.network.GET
+import eu.kanade.tachiyomi.network.POST
+import eu.kanade.tachiyomi.source.model.FilterList
+import eu.kanade.tachiyomi.source.model.MangasPage
+import eu.kanade.tachiyomi.source.model.SManga
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.decodeFromString
+import okhttp3.CacheControl
+import okhttp3.FormBody
+import okhttp3.Request
+import okhttp3.Response
 import java.text.SimpleDateFormat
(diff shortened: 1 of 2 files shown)


yemenihandsome:
> Checklist:
> 
> * [ ]  Updated `extVersionCode` value in `build.gradle` for individual extensions
> * [x]  Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
> * [ ]  Referenced all related issues in the PR body (e.g. "Closes #xyz")
> * [ ]  Added the `isNsfw = true` flag in `build.gradle` when appropriate
> * [x]  Have not changed source names
> * [ ]  Have explicitly kept the `id` if a source's name or language were changed
> * [x]  Have tested the modifications by compiling and running the extension through Android Studio
> * [ ]  Have removed `web_hi_res_512.png` when adding a new extension

and MangaLionz And MangaStarz And Mangalink

yemenihandsome:
@mohamedotaku  and MangaLionz And MangaStarz And Mangalink

stevenyomi:
This chunk of code seems to be from https://github.com/keiyoushi/extensions-source/blob/main/multisrc/overrides/madara/mangalek/src/Mangalek.kt - not sure if this should be pulled up to the Madara class, probably need someone familiar with Madara to review.

beer-psi:
> This chunk of code seems to be from https://github.com/keiyoushi/extensions-source/blob/main/multisrc/overrides/madara/mangalek/src/Mangalek.kt - not sure if this should be pulled up to the Madara class, probably need someone familiar with Madara to review.

Normal Madara search works fine on all of these sites, it's just their normal Madara search endpoint being Cloudflare-protected. Maybe this can be pulled up to the Madara class under an option (e.g. `useFullSearch`) but I don't think we should add more code onto Madara to mitigate a few sites.

Riztard:
> but I don't think we should add more code onto Madara to mitigate a few sites.

how do you know if its just few? since cloudflare issue will get auto closed

beer-psi:
> how do you know if its just few? since cloudflare issue will get auto closed

The point is not how many, the point is trying to fix something that's not broken because Cloudflare is in the way.

PR #1063, opened 2024-02-06
fix "impossible de controuner Cloudflare" MangaLionz

mohamedotaku:
Checklist:


beer-psi:
It seems like MangaLionz put Cloudflare in front of their search page, but normal Madara search works fine on MangaLionz, I don't see the point. We can't and don't mitigate Cloudflare issues outside of randomizing user agents.

PR #1131, opened 2024-02-08
NepNep - conditionally add scanlator info

SnakeDoc83:
Useful in some cases where a user would like to filter out certain chapters of a manga.

Checklist:

### DIFF
--- a/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/nepnep/NepNep.kt
+++ b/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/nepnep/NepNep.kt
@@ -310,11 +310,16 @@ abstract class NepNep(
     override fun chapterListParse(response: Response): List<SChapter> {
         val vmChapters = response.asJsoup().select("script:containsData(MainFunction)").first()!!.data()
             .substringAfter("vm.Chapters = ").substringBefore(";")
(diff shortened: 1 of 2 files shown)


AwkwardPeak7 (commented):
wouldn't this mess with existing downloads since they don't have the scanlator in the file name. 
Will also cause same issue if and when the source changes the chapters to official version, and scanlator changes

SnakeDoc83:
> wouldn't this mess with existing downloads since they don't have the scanlator in the file name. Will also cause same issue if and when the source changes the chapters to official version, and scanlator changes

To point 1: I hadn't considered that but it doesn't seem to. I just ran a quick test in AS: downloaded a chapter of OPM using current extension, replaced that extension with this version, refreshed chapter list, turned off wifi/data, and was able to read the downloaded chapter (and nothing else). 

To point 2: I replaced this version with a new one that set scanlator to a name different than what test 1 had set, refreshed chapter list again, turned off wifi/data, and was able to read the downloaded chapter (and nothing else).

Entirely possible these are flawed tests or I'm not considering something though.

AwkwardPeak7 (approved):
I often see MangaBox sources loose downloaded chapters when theychange chapter urls which inturn changes the scanlator (which is the chapter url host). But if it works fine in the tests you conducted then it's fine

PR #1146, opened 2024-02-08
Update WPComics multisrc

cuong-tran:
* Add Filter to Nhattruyen by moving FilterList to WPComics
* Fix searching Nettruyen/Nhattruyen with filter Genre on
* Fix searching Nhattruyen when filter status=Completed
* Update Nettruyen/Nhattruyen genres list
* XoxoComics: support filter by genre & status

Checklist:

### DIFF
--- a/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/wpcomics/WPComics.kt
+++ b/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/wpcomics/WPComics.kt
@@ -73,6 +73,9 @@ abstract class WPComics(
 
     protected open val searchPath = "tim-truyen"
 
+    protected open val replaceSearchPathOld = ""
+    protected open val replaceSearchPathNew = ""
+
     override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
         val filterList = filters.let { if (it.isEmpty()) getFilterList() else it }
         return if (filterList.isEmpty()) {
@@ -94,7 +97,7 @@ abstract class WPComics(
                 addQueryParameter("sort", "0")
             }
 
-            GET(url.toString().replace("/tim-truyen?status=2&", "/truyen-full?"), headers)
+            GET(url.toString().replace(replaceSearchPathOld, replaceSearchPathNew), headers)
         }
     }
 
@@ -287,6 +290,13 @@ abstract class WPComics(
         "webtoon" to "Webtoon",
         "xuyen-khong" to "Xuyên Không",
     )
+
+    override fun getFilterList(): FilterList {
+        return FilterList(
+            StatusFilter(getStatusList()),
+            GenreFilter(getGenreList()),
+        )
+    }
 
     protected open class UriPartFilter(displayName: String, val vals: Array<Pair<String?, String>>) :
         Filter.Select<String>(displayName, vals.map { it.second }.toTypedArray()) {
--- a/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/wpcomics/WPComics.kt
+++ b/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/wpcomics/WPComics.kt
@@ -73,8 +73,7 @@ abstract class WPComics(
 
     protected open val searchPath = "tim-truyen"
 
-    protected open val replaceSearchPathOld = ""
-    protected open val replaceSearchPathNew = ""
+    protected open fun String.replaceSearchPath() = this
 
     override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
         val filterList = filters.let { if (it.isEmpty()) getFilterList() else it }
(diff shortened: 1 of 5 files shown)


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/wpcomics/WPComics.kt
```diff
@@ -288,6 +291,13 @@ abstract class WPComics(
         "xuyen-khong" to "Xuyên Không",
     )
 
+    override fun getFilterList(): FilterList {
```
SnakeDoc83, line 293: These would now also show up in XoxoComics where they do nothing. You could add an override in XoxoComics with appropriate filters or at least an emptylist so that these don't show up.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/wpcomics/WPComics.kt
```diff
@@ -94,7 +97,7 @@ abstract class WPComics(
                 addQueryParameter("sort", "0")
             }
 
-            GET(url.toString().replace("/tim-truyen?status=2&", "/truyen-full?"), headers)
+            GET(url.toString().replace(replaceSearchPathOld, replaceSearchPathNew), headers)
```
SnakeDoc83: I'm not sure this is actually accomplishing something. NhatTruyen overrides searchPath; so the `replace()` as it was should only be effecting NetTruyen.

SnakeDoc83 (dismissed):

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/wpcomics/WPComics.kt
cuong-tran: okie

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/wpcomics/WPComics.kt
cuong-tran: yes, currently it only affects NetTruyen but what if another source is added in the future or in another fork decides to use the same "tim-truyen" as `searchPath`? They would have to overwrite the whole `searchMangaRequest` function with the same code as the generator just to avoid the replacement.
Or like XoxoComics has to overwrite the `searchMangaRequest` function instead of only needing to overwrite `searchPath`.
Because the replacement code was there previously so I only fixed it and kept its position. But actually, maybe it should be placed inside NetTruyen only since only it needs that. NetTruyen would still need to overwrite the whole `searchMangaRequest` function though.
How do you think it should be?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/wpcomics/WPComics.kt
cuong-tran: XoxoComics is not actually using WPComics generator. Should I create another PR to move it out of multisrc?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/wpcomics/WPComics.kt
SnakeDoc83: Originally WPComics came together as a hodgepodge of a few sources and `searchMangaRequest()` was a basic implementation. It was slowly reworked to add filters and accommodate the sources that relied on it; no one got around to refactoring so that XoxoComics wouldn't need to override. In other words, it ended up that way rather than it needed to be that way.

I kind of think it'd be better to wait for an actual case where the `replace()` as it is creates a conflict rather than changing for a hypothetical case. What if new sources get added that would have worked with the `replace()` as it was? Then those new sources have to include overrides for your new `vals` when they could have worked without having to add overrides.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/wpcomics/WPComics.kt
SnakeDoc83: What do you mean it isn't?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/wpcomics/WPComics.kt
cuong-tran: It's not really a hypothetical case for me. My fork has 4 different sources: nettruyenss.com, nhattruyento.com, nettruyenx.com & nhattruyens.com. Only nettruyenss.com needs to use the replace. That means the replace should not be default for the generator but should be used only when a source needs it.

I think one should take as less effort as possible to add a new source. Leaving the replace there, to me, is kind of like a trap for future contributors/maintainers :) Just trying to save some efforts

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/wpcomics/WPComics.kt
cuong-tran: Sorry, just my misunderstanding. I'm implementing XoxoComics's filter.

stevenyomi (approved):

Issue #265, opened 2024-01-15
Migrate to the new multisrc structure


stevenyomi:
- [x] Migrate
  - #278
  - #1200
  - #1228
  - #1237
- [x] Rename projects https://github.com/keiyoushi/extensions-source/pull/259#discussion_r1451996277
  - #1240
- [ ] Enforce PR build check on merge (after existing PRs got merged)
- [x] Documentation
- [x] Reuse `baseUrl` in source code (depends on #1617)

## General/long-term todos
- [x] Partial build in CI and PR build checks
- [ ] Scan scripts and audit scripts
- [ ] Custom audit rules enforced with ktlint
- [ ] Rename namespace (depends on #1617)

AwkwardPeak7:
mostly resolved

PR #1200, opened 2024-02-11
lib-multisrc migration (draft for review)

stevenyomi:
References #265

See each commit for migration effect.

- Used Kotlin DSL and precompiled script plugin
- Retained the variable names: `baseVersionCode`, `overrideVersionCode`, `themePkg`


common.gradle
```diff
@@ -34,8 +36,21 @@ android {
         manifestPlaceholders = [
                 appName : "Tachiyomi: $extName",
                 extClass: extClass,
-                nsfw: project.ext.find("isNsfw") ? 1 : 0,
+                nsfw    : project.ext.find("isNsfw") ? 1 : 0,
         ]
+        String baseUrl = project.ext.find("baseUrl") ?: ""
+        if (theme != null && !baseUrl.isEmpty()) {
```
AwkwardPeak7, line 42: I think baseUrl field should be required if the theme lib have a deeplink manifest. Could add an extra boolean property to the themelib `requireBaseUrl` which governs this.

src/all/coomer/build.gradle
```diff
@@ -0,0 +1,10 @@
+ext {
+    extName = 'Coomer'
+    extClass = '.Coomer'
+    themePkg = 'kemono'
+    baseUrl = 'https://coomer.su'
```
AwkwardPeak7, line 5: not important but `baseUrl` shouldn't be added when themelib doesn't support deeplinks.

common.gradle
stevenyomi: It is guaranteed by AGP - if you do not fill in `baseUrl`, AGP will complain about missing substitution values.

src/all/coomer/build.gradle
stevenyomi: It's added because it was a required field in the generator, and it doesn't hurt.

Also, if it's omitted and the theme package adds deep link some day, you'll need to go through all overrides and add it anyway.

common.gradle
```diff
@@ -34,8 +36,21 @@ android {
...
+            if (baseUrl.endsWith("/")) {
+                println("::warning ::baseUrl[$baseUrl] has a trailing slash")
+            }
+            def split = baseUrl.split("://")
+            assert split.length == 2
```
AwkwardPeak7, line 47: I think build should fail if `baseUrl` contains anything after the domain. so `http://example.com/something` or `http://example.com/` all fail.

beer-psi: What if the site hosts its manga reader in a subdirectory? `https://example.com/something` should be valid IMO.

stevenyomi: If so, this field should be renamed to `schemeWithDomain` or something. Some sources do require `baseUrl` having more path segments or trailing slash. My intention was to keep this in sync with the one inside the source, and warn if someone updated both to have a trailing slash. This could be moved to some auditing script, etc. I'm open to suggestions.

PR #1292, opened 2024-02-16
Madara refactor

AwkwardPeak7:
- Removed Random UA; it should be added on a per source basis
- Added support for `madara_load_more` ajax request to fetch popular/latest/search manga. It was removed a while back from the multisrc since new Madara theme update removed it, but many sites still use it and regular `/page/$page` request can lead to 404. 
   - Detects if site uses `madara_load_more` and enables it automatically.
   - explicitly enable with `useLoadMoreRequest`

- changed `searchPage` to not add `/page/$page` when on page=1 as it leads to 302 redirects on many sites.
- some cleanup
   - `:lib:i18n` for localization
   - fixed/removed some useless overrides
   - select().first() -> selectFirst()
   - xhrHeaders variable


vetleledaal:
> * changed `searchPage` to not add `/page/$page` when on page=1 as it leads to 302 redirects on many sites.

I ran `grep -r ' : Madara' --files-with-matches | xargs grep -A3 'fun searchPage'` and found one you missed:
```
src/es/mangasnosekai/src/eu/kanade/tachiyomi/extension/es/mangasnosekai/MangasNoSekai.kt:    override fun searchPage(page: Int): String {
src/es/mangasnosekai/src/eu/kanade/tachiyomi/extension/es/mangasnosekai/MangasNoSekai.kt-        return if (page > 1) "page/$page/" else ""
src/es/mangasnosekai/src/eu/kanade/tachiyomi/extension/es/mangasnosekai/MangasNoSekai.kt-    }
src/es/mangasnosekai/src/eu/kanade/tachiyomi/extension/es/mangasnosekai/MangasNoSekai.kt-
```

The remaining matches did not follow the same pattern
<details>

```
src/en/shoujohearts/src/eu/kanade/tachiyomi/extension/en/shoujohearts/ShoujoHearts.kt:    override fun searchPage(page: Int): String = "reader/page/$page/"
src/en/shoujohearts/src/eu/kanade/tachiyomi/extension/en/shoujohearts/ShoujoHearts.kt-}
--
src/en/isekaiscantop/src/eu/kanade/tachiyomi/extension/en/isekaiscantop/IsekaiScanTop.kt:    override fun searchPage(page: Int) = "search?page=$page"
src/en/isekaiscantop/src/eu/kanade/tachiyomi/extension/en/isekaiscantop/IsekaiScanTop.kt-
src/en/isekaiscantop/src/eu/kanade/tachiyomi/extension/en/isekaiscantop/IsekaiScanTop.kt-    override fun searchMangaNextPageSelector() = "ul.pagination li:last-child a"
src/en/isekaiscantop/src/eu/kanade/tachiyomi/extension/en/isekaiscantop/IsekaiScanTop.kt-}
--
src/en/mangapure/src/eu/kanade/tachiyomi/extension/en/mangapure/MangaPure.kt:    override fun searchPage(page: Int): String = "search?page=$page"
src/en/mangapure/src/eu/kanade/tachiyomi/extension/en/mangapure/MangaPure.kt-
src/en/mangapure/src/eu/kanade/tachiyomi/extension/en/mangapure/MangaPure.kt-    override fun popularMangaRequest(page: Int): Request =
src/en/mangapure/src/eu/kanade/tachiyomi/extension/en/mangapure/MangaPure.kt-        GET("$baseUrl/popular-manga?page=$page", headers)
--
src/en/toonily/src/eu/kanade/tachiyomi/extension/en/toonily/Toonily.kt:    private fun searchPage(page: Int, query: String): String {
src/en/toonily/src/eu/kanade/tachiyomi/extension/en/toonily/Toonily.kt-        val urlQuery = query.trim()
src/en/toonily/src/eu/kanade/tachiyomi/extension/en/toonily/Toonily.kt-            .lowercase(Locale.US)
src/en/toonily/src/eu/kanade/tachiyomi/extension/en/toonily/Toonily.kt-            .replace(titleSpecialCharactersRegex, "-")
```
</details>

lib-multisrc/madara/src/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
```diff
@@ -958,9 +1109,8 @@ abstract class Madara(
             .toByteArray()
     }
 
-    override fun setupPreferenceScreen(screen: PreferenceScreen) {
-        addRandomUAPreferenceToScreen(screen)
-    }
+    protected fun launchIO(block: () -> Unit) =
```
vetleledaal: IDE says:
> This is a delicate API and its use requires care. Make sure you fully read and understand documentation of the declaration that is marked as a delicate API.
```suggestion
    @OptIn(DelicateCoroutinesApi::class)
    protected fun launchIO(block: () -> Unit) =
```

Any reason why this shouldn't be added?

AwkwardPeak7: that propogates the warning to every call site of the function

AwkwardPeak7:
> > * changed `searchPage` to not add `/page/$page` when on page=1 as it leads to 302 redirects on many sites.
> 
> I ran `grep -r ' : Madara' --files-with-matches | xargs grep -A3 'fun searchPage'` and found one you missed:
> 
> ```
> src/es/mangasnosekai/src/eu/kanade/tachiyomi/extension/es/mangasnosekai/MangasNoSekai.kt:    override fun searchPage(page: Int): String {
> src/es/mangasnosekai/src/eu/kanade/tachiyomi/extension/es/mangasnosekai/MangasNoSekai.kt-        return if (page > 1) "page/$page/" else ""
> src/es/mangasnosekai/src/eu/kanade/tachiyomi/extension/es/mangasnosekai/MangasNoSekai.kt-    }
> src/es/mangasnosekai/src/eu/kanade/tachiyomi/extension/es/mangasnosekai/MangasNoSekai.kt-
> ```

thanks, that slipped in when I rebased on main

lib-multisrc/madara/src/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
```diff
@@ -912,16 +1063,16 @@ abstract class Madara(
...
+                fetchGenresFailed = true
+            } finally {
+                fetchGenresAttempts++
             }
-
-            fetchGenresFailed = genres.isFailure
-            genresList = genres.getOrNull().orEmpty()
-            fetchGenresAttempts++
         }
     }
```
vetleledaal: Should be doable without `fetchGenresFailed`? The variable isn't used outside of here.
```kt
protected fun fetchGenres() {
    if (!fetchGenres || fetchGenresAttempts >= 3 || genresList.isNotEmpty()) return

    try {
        genresList = client.newCall(genresRequest()).execute()
            .use { parseGenres(it.asJsoup()) }
    } catch (_: Exception) {
    } finally {
        fetchGenresAttempts++
    }
}
```

lib-multisrc/madara/src/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
beer-psi: why not use a CoroutineScope instead of a GlobalScope then?

AwkwardPeak7: I was under the impression it gets cancelled once function scope ends. turns out that isn't how it works.

vetleledaal (approved):

PR #1338, opened 2024-02-17
NetTruyen/NhatTruyen: avoid return un-relevant searching results

cuong-tran:
They redirect back to catalog page if searching query is not found. That makes both sites always return un-relevant results when searching should have returned empty.

Checklist:

### DIFF
--- a/lib-multisrc/wpcomics/build.gradle.kts
+++ b/lib-multisrc/wpcomics/build.gradle.kts
@@ -2,4 +2,4 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 4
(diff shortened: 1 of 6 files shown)


lib-multisrc/wpcomics/build.gradle.kts
```diff
@@ -2,4 +2,4 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 4
+baseVersionCode = 5
```
AwkwardPeak7: no functionality was changed in the base class so don't bump `baseVersionCode`

AwkwardPeak7 (approved):

PR #1344, opened 2024-02-17
Update MangaDistrict

cuong-tran:
* Remove comment from title e.g. (Official), (Doujinshi)
* Settings to parse only chapters with full or high or both quality
* Fix searchMangaNextPageSelector for last page

Checklist:

### DIFF
--- a/src/en/mangadistrict/src/eu/kanade/tachiyomi/extension/en/mangadistrict/MangaDistrict.kt
+++ b/src/en/mangadistrict/src/eu/kanade/tachiyomi/extension/en/mangadistrict/MangaDistrict.kt
@@ -6,6 +6,7 @@ import androidx.preference.ListPreference
 import androidx.preference.PreferenceScreen
 import androidx.preference.SwitchPreferenceCompat
 import eu.kanade.tachiyomi.multisrc.madara.Madara
+import eu.kanade.tachiyomi.source.ConfigurableSource
 import eu.kanade.tachiyomi.source.model.SChapter
 import eu.kanade.tachiyomi.source.model.SManga
 import eu.kanade.tachiyomi.util.asJsoup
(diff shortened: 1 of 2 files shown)


src/en/mangadistrict/src/eu/kanade/tachiyomi/extension/en/mangadistrict/MangaDistrict.kt
```diff
@@ -1,11 +1,124 @@
...
+        val document = asJsoup()
+        document.select(selector).remove()
+
+        val body: ResponseBody = document.body().toString().toResponseBody(contentType)
+        return newBuilder().body(body).build()
+    }
+
+    override fun chapterListParse(response: Response): List<SChapter> {
+        return super.chapterListParse(response.stripElements(chaptersStripSelector()))
+    }
```
AwkwardPeak7: don't rebuild the response, just reimplement the chapterListParse.

Also, wouldn't this remove chapters if only one of "High quality" or "low quality" chapter exists?

cuong-tran: That's why I don't select the required quality, but I remove the oposite quality instead. If manga doesn't have separate qualities then the selector won't exist and nothing will be removed.
Would you explain why shouldn't rebuild the response? I want to call the super so that any future updates to multisrc would also update MangaDistrict.
Thanks.

AwkwardPeak7: would it be possible to remove those chapters based on the url slug? so you can simply filter the result of super.chapterListParse

src/en/mangadistrict/src/eu/kanade/tachiyomi/extension/en/mangadistrict/MangaDistrict.kt
```diff
@@ -29,80 +29,71 @@ class MangaDistrict :
...
         SwitchPreferenceCompat(screen.context).apply {
-            key = REMOVE_TITLE_COMMENT_PREF
-            title = "Remove comments from title, e.g. \"(Official)\", \"(Doujinshi)\""
-            summary = "To help app find duplicate entries in library"
-            setDefaultValue(true)
+            key = REMOVE_TITLE_VERSION_PREF
+            title = "Remove version information from entry titles"
+            summary = "This removes version tags like “(Official)” or “(Doujinshi)” from entry titles " +
+                "and helps identify duplicate entries in your library. " +
+                "To update existing entries, remove them from your library and clear database in advanced settings."
```
cuong-tran: No need to remove entry from library, just need to refresh each entry to update its title.

stevenyomi: The app won't update the titles of entries in the library (i.e. favorited). Clearing database is not necessary indeed.

AwkwardPeak7 (approved):

PR #1356, opened 2024-02-18
move CookieInterceptor to shared lib

AwkwardPeak7:


beer-psi (changes requested):
Can we make it accept more than 1 cookie by accepting a `List<Cookie>` instead? OkHttp3 has a [`Cookie.Builder`](https://square.github.io/okhttp/3.x/okhttp/okhttp3/Cookie.Builder.html#build--) class.

beer-psi (approved):

PR #1432, opened 2024-02-20
[RU]ComX no Tachiyomi (maybe because of the closure), fix 520

e-shl:
Checklist:

### DIFF
--- a/src/ru/comx/src/eu/kanade/tachiyomi/extension/ru/comx/ComX.kt
+++ b/src/ru/comx/src/eu/kanade/tachiyomi/extension/ru/comx/ComX.kt
@@ -107,7 +107,7 @@ class ComX : ParsedHttpSource() {
         }
         .build()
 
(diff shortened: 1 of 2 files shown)


src/ru/comx/src/eu/kanade/tachiyomi/extension/ru/comx/ComX.kt
```diff
@@ -108,7 +108,6 @@ class ComX : ParsedHttpSource() {
         .build()
 
     override fun headersBuilder(): Headers.Builder = Headers.Builder()
```
AwkwardPeak7:
```suggestion
    override fun headersBuilder(): Headers.Builder = super.headersBuilder()
```
so default useragent is preserved from super class

AwkwardPeak7 (changes requested):

AwkwardPeak7 (approved):

PR #1476, opened 2024-02-22
Fix Madara was unable to reset Genre’s checkbox

cuong-tran:
Checklist:

### DIFF
--- a/lib-multisrc/madara/build.gradle.kts
+++ b/lib-multisrc/madara/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 35
+baseVersionCode = 36
 
 dependencies {
     api(project(":lib:cryptoaes"))
--- a/lib-multisrc/madara/build.gradle.kts
+++ b/lib-multisrc/madara/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 36
+baseVersionCode = 35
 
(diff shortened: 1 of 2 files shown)


lib-multisrc/madara/build.gradle.kts
```diff
@@ -2,7 +2,7 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 35
+baseVersionCode = 36
```
AwkwardPeak7: not too important so don't bump version

AwkwardPeak7 (changes requested):
rebase on main as well.
also make sure to update it in any source that have overrides related to this.

lib-multisrc/madara/build.gradle.kts
cuong-tran: on another PR earlier, the change was even less significant but @beerpiss said need to bump version no matter how small change is. It's kind of hard to understand :)

cuong-tran:
> also make sure to update it in any source that have overrides related to this.

```kotlin
protected class GenreList(title: String, genres: List<Genre>) : Filter.Group<GenreCheckBox>(title, genres.map { GenreCheckBox(it.name, it.id) })
``` 
It is not overridable or inheritable.

lib-multisrc/madara/build.gradle.kts
AwkwardPeak7: What beer said holds true but Madara is too big to push update every day

AwkwardPeak7 (approved):

PR #1643, opened 2024-03-01
Anchira: Add Entry Bundling

BrutuZ:
- Prefixing queries with `bundle:` will create a single SManga entry with results as chapters. Capped at 5 pages (120 chapters)
  - The goal is to have multi-work series as a single entry in the library
- Chapters now use the entry title instead of the generic `Chapter` and a parsed chapter number (helps with tracking)
- Simplified the DTO to a single _Entry_ class by setting defaults to fields absent in Library entries
- Removed unused decrypt file

image

Checklist:


src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
```diff
@@ -193,26 +202,42 @@ class Anchira : HttpSource(), ConfigurableSource {
...
 
     override fun mangaDetailsParse(response: Response): SManga {
-        val data = json.decodeFromString<Entry>(response.body.string())
+        return if (response.request.url.pathSegments.count() == apiUrl.toHttpUrl().pathSegments.count()) {
+            val manga = latestUpdatesParse(response).mangas.first()
+            val query = response.request.url.queryParameter("s")
+            manga.apply {
+                url = "?${response.request.url.query}"
+                description = "Bundled from $query"
+                title = "[Bundle] $query"
```
AwkwardPeak7: title should maybe contain series title instead of the query

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
```diff
@@ -109,6 +112,12 @@ class Anchira : HttpSource(), ConfigurableSource {
             fetchMangaDetails(manga).map {
                 MangasPage(listOf(it), false)
             }
+        } else if (query.startsWith(SLUG_BUNDLE_PREFIX)) {
```
AwkwardPeak7, line 116: what about a filter entry (checkbox?) to toggle this.

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
```diff
@@ -109,6 +112,12 @@ class Anchira : HttpSource(), ConfigurableSource {
...
+            // bundle entries as chapters
+            val manga = SManga.create().apply { this.url = "?s=${query.substringAfter(SLUG_BUNDLE_PREFIX)}&order=1" }
```
AwkwardPeak7: consider putting the filters as well (final url after filters are applied?)

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
BrutuZ: That can be arranged. I went for this instead because another way I use it is for artist collections

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
BrutuZ: Haven't considered it. I avoid messing with filters because it breaks TachiSY's saved searches and adding `bundle:` is simple enough. I meant it as an advanced feature due to the extra calls, so the friction works in favor

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
BrutuZ: Last I tried with the "site's last iteration" adding other parameters didn't work that well so I went straight for the safe approach on this one.

The filters are also directly inside `searchMangaRequest` on this one, I'd have to put them on its own function

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
```diff
@@ -193,26 +202,42 @@ class Anchira : HttpSource(), ConfigurableSource {
...
 
     override fun mangaDetailsParse(response: Response): SManga {
-        val data = json.decodeFromString<Entry>(response.body.string())
+        return if (response.request.url.pathSegments.count() == libraryUrl.toHttpUrl().pathSegments.count()) {
+            val manga = latestUpdatesParse(response).mangas.first()
+            val query = response.request.url.queryParameter("s")
+            manga.apply {
+                url = "?${response.request.url.query}"
+                description = "Bundled from $query"
+                title = "[Bundle] ${manga.title}"
```
AwkwardPeak7: possible to remove the chapter number from the title? or maybe use the title from "title:<>" in the query if it exists

[something like this](https://github.com/AwkwardPeak7/tachiyomi-extensions/blob/ksk/src/en/koushoku/src/eu/kanade/tachiyomi/extension/en/koushoku/Koushoku.kt#L187-L191)

BrutuZ: It's possible since it's already done for chapter titles, but would need to either duplicate that couple lines of code or make it it's own function since it's part of `createChapter`

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
BrutuZ: Turns out it wasn't so hard since that whole function was basically only the filter already so managed to repurpose it for both the Fetch and Request functions

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/Anchira.kt
BrutuZ: Other than duplicating the RegEx it wasn't as bad as I initially thought :)

src/en/anchira/src/eu/kanade/tachiyomi/extension/en/anchira/AnchiraHelper.kt
```diff
@@ -25,4 +29,31 @@ object AnchiraHelper {
             }
         }
         .joinToString(", ") { it }
+
+    fun createChapter(entry: Entry, response: Response, anchiraData: List<EntryKey>) =
+        SChapter.create().apply {
+            val ch =
+                Regex("(?<!20\\d\\d-)\\b[\\d.]{1,4}$").find(entry.title)?.value?.trim('.') ?: "1"
```
AwkwardPeak7: make this regex a constant variable in this file and use that in main class as well

AwkwardPeak7: I meant making the regex itself a constant

```
val CHAPTER_SUFFIX_REGEX = Regex("(?<!20\\d\\d-)\\b[\\d.]{1,4}$")
```

BrutuZ: In my defense, you did say _constant_ and those only take primitives xD

AwkwardPeak7 (approved):

PR #1756, opened 2024-03-07
Toonkor: update baseUrl automatically

### LINKED ISSUE #227: [URL change] Toonkor
### Source information

Toonkor 1.4.4

### Source language

Korean

### Source new URL

https://toonkor325.com

### Other details

_No response_

### Acknowledgements

- [x] I have opened WebView and checked that the source URL is not updated yet.

vetleledaal:
Closes #227

Probably a horrible way to solve the issue.

Checklist:

### DIFF
--- a/src/ko/toonkor/src/eu/kanade/tachiyomi/extension/ko/toonkor/Toonkor.kt
+++ b/src/ko/toonkor/src/eu/kanade/tachiyomi/extension/ko/toonkor/Toonkor.kt
@@ -29,9 +29,13 @@ class Toonkor : ConfigurableSource, ParsedHttpSource() {
 
     override val name = "Toonkor"
 
-    private val defaultBaseUrl = "auto"
+    private val defaultBaseUrl = "https://toonkor342.com"
+
+    private val defaultAutoBaseUrl = true
 
     private val BASE_URL_PREF = "overrideBaseUrl_v${AppInfo.getVersionName()}"
+
+    private val AUTO_BASE_URL_PREF = "autoBaseUrl"
 
     override val baseUrl by lazy { getPrefBaseUrl() }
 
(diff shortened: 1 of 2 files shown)


src/ko/toonkor/src/eu/kanade/tachiyomi/extension/ko/toonkor/Toonkor.kt
```diff
@@ -213,7 +216,21 @@ class Toonkor : ConfigurableSource, ParsedHttpSource() {
         screen.addPreference(baseUrlPref)
     }
 
-    private fun getPrefBaseUrl(): String = preferences.getString(BASE_URL_PREF, defaultBaseUrl)!!
+    private fun getPrefBaseUrl(): String {
+        val prefUrl = preferences.getString(BASE_URL_PREF, defaultBaseUrl) ?: defaultBaseUrl
+        if (prefUrl != "auto") return prefUrl
```
AwkwardPeak7: I am not too sure on adding "auto" in the url field...

maybe a switch pref to use auto or manual links, in auto mode, it will fetch the url and in manual mode it will use the one from preference.
can hide the manual link entering when auto is is enabled through `setVisible`

vetleledaal: I tested with `setVisible`, but the UX was kinda bad. You have to reopen the preferences screen to input the override URL. Opted to clarify the override URL option instead.

I set it to be enabled by default as it will lessen the support burden. The setting is kept across versions, as opposed to the override URL.

AwkwardPeak7: I would say make the baseUrl pref also not depend on version.
implementation like [this](https://github.com/keiyoushi/extensions-source/blob/main/src/zh/zerobyw/src/eu/kanade/tachiyomi/extension/zh/zerobyw/UpdateUrl.kt#L24)

stevenyomi:
I think I'll rework this to make it work similarly as NewToki.

PR #1758, opened 2024-03-07
I Roved Out: Fix downloads not finishing

GoldenRover:
Downloads were stuck because of a redundant unimplemented override.

Checklist:

### DIFF
--- a/src/en/irovedout/build.gradle
+++ b/src/en/irovedout/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'I Roved Out'
     extClass = '.IRovedOut'
-    extVersionCode = 3
+    extVersionCode = 4
     isNsfw = true
 }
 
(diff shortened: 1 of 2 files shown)


AwkwardPeak7 (commented):
can you tell what was the issue? since removing this override shouldn't have any effect on app's ability to read online / download the chapter since `fetchPageList` is already overriden here.

And download stuck at end likely means last image link is dead, from the site

GoldenRover:
I actually checked it, I'm not sure why but it seems `fetchPageList` is called at the end of the download. Throwing the error made the download fail, but since I'm no longer throwing it I can only assume there is some additional functionality that can happen after downloads that this source doesn't make use of.

After removing the override I was able to download a chapter without the download failing.

AwkwardPeak7 (approved):
I suppose removing the override won't have any negative effects but I still doubt this is the reason for the described issue

Issue #1617, opened 2024-02-29
Helper code to add configurable base URL/fetch filters


stevenyomi:
See #1461

- [ ] Add lib and remove BuildConfig version constant usages
- [ ] Add ktlint rule to avoid using BuildConfig version constant (and R class)
- [ ] Update extension namespaces

AwkwardPeak7:
resolved

PR #1781, opened 2024-03-09
WPComics: baseUrl configurable so user can change domain while waitin…

cuong-tran:
…g for extension update

Checklist:

### DIFF
--- a/lib-multisrc/wpcomics/build.gradle.kts
+++ b/lib-multisrc/wpcomics/build.gradle.kts
@@ -2,4 +2,4 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 4
+baseVersionCode = 5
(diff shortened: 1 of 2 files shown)


AwkwardPeak7:
I don't think adding condifurable baseUrl to a multisrc theme is desirable since not all of them need it. Furthermore, we want to remove usage of `BuildConfig.VERSION_CODE` and `AppInfo.getVersionName()` from baseUrl preferences, tracked in #1617. So please wait for that to happen.

If you still want to add it to the vi sources, then make a similar implementation to [this](https://github.com/keiyoushi/extensions-source/blob/main/src/zh/zerobyw/src/eu/kanade/tachiyomi/extension/zh/zerobyw/UpdateUrl.kt#L24)

cuong-tran:
Will wait for #1617, that sounds better

PR #1808, opened 2024-03-11
Add Clown Corps comic source

altaccosc:
Checklist:


altaccosc:
I forgot the dates
Will add those tomorrow!

src/en/clowncorps/build.gradle
```diff
@@ -0,0 +1,11 @@
...
+    extName = 'Clown Corps'
+    extClass = '.ClownCorps'
+    extVersionCode = 1
+}
+
+apply from: "$rootDir/common.gradle"
+
+dependencies {
+    implementation(project(':lib:textinterceptor'))
+}
```
AwkwardPeak7:
```suggestion
}

```
newline

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,158 @@
...
+    private val creator = "Joe Chouinard"
+
+    //Text from: https://clowncorps.net/about/
+    private val synopsis = "Clown Corps is a comic about crime-fighting clowns. " +
+        "It's pronounced \"core.\" Like marine corps."
+
+    override val client: OkHttpClient
+        get() = super.client.newBuilder()
+            .addInterceptor(TextInterceptor())
+            .build()
```
AwkwardPeak7:
```suggestion
    override val client = network.client.newBuilder()
            .addInterceptor(TextInterceptor())
            .build()
```

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,158 @@
...
+        SManga.create().apply {
+            title = name
+            artist = creator
+            author = creator
+            description = synopsis
+            status = SManga.ONGOING
+            //Image from: https://clowncorps.net/about/
+            thumbnail_url = "https://clowncorps.net/wp-content/uploads/2022/11/clowns41.jpg"
+            setUrlWithoutDomain("/comic")
+        }.let { Observable.just(MangasPage(listOf(it), false))!! }
```
AwkwardPeak7:
```suggestion
    override fun fetchPopularManga(page: Int): Observable<MangasPage> = Observable.fromCallable {
        MangasPage(
            listOf(
                SManga.create().apply {
                    title = name
                    artist = creator
                    author = creator
                    description = synopsis
                    status = SManga.ONGOING
                    //Image from: https://clowncorps.net/about/
                    thumbnail_url = "https://clowncorps.net/wp-content/uploads/2022/11/clowns41.jpg"
                    setUrlWithoutDomain("/comic")
                }
            ),
            hasNextPage = false
        )
    }
```

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    override fun fetchSearchManga(page: Int, query: String, filters: FilterList) =
+        Observable.just(MangasPage(emptyList(), false))!!
+
+    override fun fetchMangaDetails(manga: SManga) =
+        Observable.just(manga.apply { initialized = true })!!
```
AwkwardPeak7:
```suggestion
        Observable.just(manga.apply { initialized = true })
```

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,158 @@
...
+            author = creator
+            description = synopsis
+            status = SManga.ONGOING
+            //Image from: https://clowncorps.net/about/
+            thumbnail_url = "https://clowncorps.net/wp-content/uploads/2022/11/clowns41.jpg"
+            setUrlWithoutDomain("/comic")
+        }.let { Observable.just(MangasPage(listOf(it), false))!! }
+
+    override fun fetchSearchManga(page: Int, query: String, filters: FilterList) =
+        Observable.just(MangasPage(emptyList(), false))!!
```
AwkwardPeak7:
```suggestion
    override fun fetchSearchManga(page: Int, query: String, filters: FilterList) = fetchPopularManga(page)
```

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    override fun fetchMangaDetails(manga: SManga) =
+        Observable.just(manga.apply { initialized = true })!!
+
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val chapters = mutableListOf<SChapter>()
+
+        for (page in 1..Int.MAX_VALUE) {
```
AwkwardPeak7: get the last page from the response document

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,158 @@
...
+                    name = title
+                }
+                chapters.add(chapter)
+            }
+        }
+
+        chapters.forEachIndexed { index, sChapter ->
+            sChapter.chapter_number = chapters.size - index.toFloat()
+        }
+
```
AwkwardPeak7: app already populates the chapter number based on title, if it works fine then no need for this

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,158 @@
...
+        val authorsNotesPref = SwitchPreferenceCompat(screen.context).apply {
+            key = SHOW_AUTHORS_NOTES_KEY
+            title = "Show author's notes"
+            summary = "Enable to see the author's notes at the end of chapters (if they're there)."
+            setDefaultValue(false)
+
+            setOnPreferenceChangeListener { _, newValue ->
+                val checkValue = newValue as Boolean
+                preferences.edit().putBoolean(SHOW_AUTHORS_NOTES_KEY, checkValue).commit()
+            }
```
AwkwardPeak7:
```suggestion
```
not needed, app already saves the preferences

AwkwardPeak7 (changes requested):

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: Done in a1e1aca7b8a8de7127dd850fc3f2c51b4235d5fa

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: Done in f50c5421bed6f54c54cb13a40e19614efe9de4d8

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,183 @@
...
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val pageStatus = response.asJsoup().select("#paginav li.paginav-pages").text()
+        val pageCount = pageStatus.split(" ").last().toInt()
+
+        val chapters = mutableListOf<SChapter>()
+
+        for (page in 1..pageCount) {
+            val url = "https://clowncorps.net/comic/page/$page/"
+            val resp = client.newCall(GET(url, headers)).execute()
+            if (resp.code == 404) break
```
AwkwardPeak7:
```suggestion
```
remove

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,183 @@
...
+        MangasPage(
+            listOf(
+                SManga.create().apply {
+                    title = name
+                    artist = creator
+                    author = creator
+                    description = synopsis
+                    status = SManga.ONGOING
+                    // Image from: https://clowncorps.net/about/
+                    thumbnail_url = "https://clowncorps.net/wp-content/uploads/2022/11/clowns41.jpg"
```
AwkwardPeak7:
```suggestion
                    thumbnail_url = "$baseUrl/wp-content/uploads/2022/11/clowns41.jpg"
```

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,183 @@
...
+        Observable.just(manga.apply { initialized = true })
+
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val pageStatus = response.asJsoup().select("#paginav li.paginav-pages").text()
+        val pageCount = pageStatus.split(" ").last().toInt()
+
+        val chapters = mutableListOf<SChapter>()
+
+        for (page in 1..pageCount) {
+            val url = "https://clowncorps.net/comic/page/$page/"
```
AwkwardPeak7:
```suggestion
            val url = "$baseUrl/comic/page/$page/"
```

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,183 @@
...
+            }
+        }
+
+        // Add chapter numbers to the chapters, to ensure correct sorting.
+        // Seems to be unnecessary, as it already bases the order on the chapter names.
+        // Kept here, commented out, just in case.
+        /*chapters.forEachIndexed { index, sChapter ->
+            sChapter.chapter_number = chapters.size - index.toFloat()
+        }*/
+
```
AwkwardPeak7:
```suggestion
```
remove
also does it work without it?

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,183 @@
...
+        return chapters
+    }
+
+    private fun parseDate(dateStr: String): Long {
+        return try {
+            dateFormat.parse(dateStr)!!.time
```
AwkwardPeak7, line 126:
```suggestion
            dateFormat.parse(dateStr)?.time ?: 0L
```

AwkwardPeak7 (changes requested):

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
stevenyomi: @AwkwardPeak7 No it won't be `null` at all, see [`Date parse(String) throws ParseException`](https://docs.oracle.com/javase/8/docs/api/java/text/DateFormat.html#parse-java.lang.String-). IDE wrongly prompts it as nullable because it wraps [`Date? parse(String, ParsePosition)`](https://docs.oracle.com/javase/8/docs/api/java/text/DateFormat.html#parse-java.lang.String-java.text.ParsePosition-) and throws on a check of `ParsePosition` so the IDE can't tell.

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: It works with it commented out, yes. I've been using it to read the comic for the past few days without any issues.
But I wanted to keep this snippet around in case the automatic sorting happens to break at some point.
It's not hurting anything by being there, is it?

stevenyomi: This snippet forces chapter number to align with "sort by source order" in the app, so it's completely useless.

altaccosc: With it commented out, the Sort By Source option sorts the chapters properly.
But the Sort By Chapter Number option sorts the pages reversed:
- chapter 1 page 42
- chapter 1 page 41
- (...)
- chapter 1 page 1
- chapter 1 cover
- chapter 2 page 73
- chapter 2 page 72
- etc...

altaccosc: But sure, I can remove it completely, if you prefer.
I must admit I don't see much point in discussing a small commented out piece of code.

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,175 @@
...
+import java.text.SimpleDateFormat
+import java.util.Locale
+
+class ClownCorps : ConfigurableSource, HttpSource() {
+    override val baseUrl = "https://clowncorps.net"
+    override val lang = "en"
+    override val name = "Clown Corps"
+    override val supportsLatest = false
+
+    private val creator = "Joe Chouinard"
```
stevenyomi: Move to companion object.

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,175 @@
...
+
+    // Text from: https://clowncorps.net/about/
+    private val synopsis = "Clown Corps is a comic about crime-fighting clowns. " +
+        "It's pronounced \"core.\" Like marine corps."
```
stevenyomi: Inline it.

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,175 @@
...
+
+    override val client = network.client.newBuilder()
+        .addInterceptor(TextInterceptor())
+        .build()
+
+    override fun fetchPopularManga(page: Int): Observable<MangasPage> = Observable.fromCallable {
```
stevenyomi: Use `Observable.just()`.

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,175 @@
...
+        MangasPage(
+            listOf(
+                SManga.create().apply {
```
stevenyomi: Extract to `getManga()` and use it in `fetchMangaDetails()`. Also add `initialized = true` in it.

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,175 @@
...
+        return try {
+            dateFormat.parse(dateStr)!!.time
+        } catch (_: ParseException) {
+            0L
+        }
+    }
+
+    private val dateFormat by lazy {
+        SimpleDateFormat("MMMM dd, yyyy hh:mm aa", Locale.ENGLISH)
+    }
```
stevenyomi:
```suggestion
    override fun chapterListParse(response: Response): List<SChapter> {
        return response.asJsoup().select("#chapter-select .level-0").flatMap { chapterElement ->
            val url = chapterElement.attr("value")
            val document = client.newCall(GET(url, headers)).execute().asJsoup()
            document.select(".comic-list-dropdown-form .level-0").map {
                SChapter.create().apply {
                    setUrlWithoutDomain(it.attr("value"))
                    name = it.text()
                }
            }
        }.asReversed()
    }
```

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: Done in 576a230f2a118b2bf12fae4fbf1256974af08105

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: Done in 576a230f2a118b2bf12fae4fbf1256974af08105

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -26,14 +26,8 @@ import java.util.Locale
...
-    override val name = "Clown Corps"
-    override val supportsLatest = false
-
-    private val creator = "Joe Chouinard"
-
-    // Text from: https://clowncorps.net/about/
-    private val synopsis = "Clown Corps is a comic about crime-fighting clowns. " +
-        "It's pronounced \"core.\" Like marine corps."
+    override val name = NAME
+    override val supportsLatest = SUPPORTS_LATEST
```
stevenyomi: Inline `name` and `supportsLatest`. I was asking you to move the creator constant only.

altaccosc: Yeah I realised that too just now. Sorry, I don't use Kotlin a lot, so I'm not yet too caught-up on the exact conventions.
I'm learning quickly, though.
My next extensions will hopefully be better from the get-go.

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: Done in 0f555df0e9ed6801351247a6399cc608eed416b6

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: Done in 0f555df0e9ed6801351247a6399cc608eed416b6

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: Done in 6dd186f4268be8df13a72c37617506f26a0ff005

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: This does not set the date. Don't we want that?
All chapters are now all listed as having been posted today.

stevenyomi: Your implementation is too slow - it will iterate through the 45 pages and send that many requests. That puts a lot of stress onto the source server.

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,168 @@
...
+        title = name
+        artist = CREATOR
+        author = CREATOR
+        status = SManga.ONGOING
+        initialized = true
+        // Image and description from: https://clowncorps.net/about/
+        thumbnail_url = "$baseUrl/wp-content/uploads/2022/11/clowns41.jpg"
+        description = "Clown Corps is a comic about crime-fighting clowns.\n" +
+            "It's pronounced \"core.\" Like marine corps."
+        setUrlWithoutDomain("/comic")
```
stevenyomi:
```suggestion
        url = "/comic"
```

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: That's true and fair.
Is there nothing else that could be done for this, though?

altaccosc: Oh, I just found something new!
Maybe we could loop over the years?
https://clowncorps.net/2024/?post_type=comic

stevenyomi: Nothing much we can do. The current extension API doesn't provide the existing chapter list, so we can only fetch from scratch.

You can hack it by storing chapter info into shared preferences (considering you have only one manga entry) but code logic to do patch updates will be complicated and error-prone.

stevenyomi: > Oh, I just found something new!\nMaybe we could loop over the years?\nhttps://clowncorps.net/2024/?post_type=comic

That's still 10 pages per webpage.

altaccosc: Drat

altaccosc: I added caching in 6ed7adf6be7170a20bdb913e9d115bbd18d67542, so hopefully that should lessen the load on the server enough to the point that my original method can still be used.

Quick summary of the way this caching works:
The first time, it loads all (45 as of the time of writing) webpages, and collects the chapters from those.
It then saves all of those chapters in the SharedPreferences, serialized to JSON.

The next time the comic gets refreshed, it gets all the chapters it can from the cache, and stores them in a **set**.
It then fetches every webpage starting at 1 again, which is where the most recent new chapters get added to.
Then, it continues going through every next page (thus going further and further into the past).
It keeps adding the newly fetched chapters into the set, but when all the chapters it just fetched are already in the set, it knows it has reached where the cache left off, so it stops the refresh.
If you regularly update, that means it only ever does one request.

Hopefully this explanation makes the code make a bit more sense.

I also added a button in the extension preferences menu to clear the cache.

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,241 @@
...
+
+        override fun equals(other: Any?): Boolean {
+            if (this === other) return true
+            if (javaClass != other?.javaClass) return false
+
+            other as SerializableChapter
+
+            return name == other.name
+        }
+    }
```
stevenyomi:
```suggestion
    @Serializable
    class SerializableChapter(val fullLink: String, val name: String, val dateUpload: Long) {
        override fun hashCode() = fullLink.hashCode()
        override fun equals(other: Any?) = other is SerializableChapter && fullLink == other.fullLink
    }
```

Should use URL as identifier. And don't use `data class` if not needed.

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,241 @@
...
+        // The total number of webpages with chapters on them
+        val currentPageIndicator = response.asJsoup().select("#paginav li.paginav-pages").text()
+        val totalWebpageCount = currentPageIndicator.split(" ").last().toInt()
+
+        val allChapters = getChaptersFromCache().toMutableSet()
+        // Fetch all the chapters from the website until we reached where the cache left off
+        for (webpageIndex in 1..totalWebpageCount) {
+            val anyChaptersWereAdded = allChapters.addAll(fetchWebpageChapters(webpageIndex))
+            if (!anyChaptersWereAdded) break // No new chapters were added from this webpage, so we're done
+        }
```
stevenyomi: Isn't the current response already the first page?

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,241 @@
...
+
+        // Save the chapters to cache
+        val fullJsonString = Json.encodeToString(allChapters)
+        preferences.edit().putString(CACHE_KEY_CHAPTERS, fullJsonString).apply()
+
+        // Convert the serializable chapters to SChapters
+        return allChapters.map { chapter ->
```
stevenyomi: How are you maintaining the chapters' order? It should be sorted from new to old. Either sort it by upload date (and hope post edits don't reset it) or maintain the order from the beginning.

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,241 @@
...
+                name = chapter.name
+                date_upload = chapter.dateUpload
+            }
+        }
+    }
+
+    private fun getChaptersFromCache(): Set<SerializableChapter> {
+        val cachedChaps = preferences.getString(CACHE_KEY_CHAPTERS, null)
+        if (cachedChaps != null) return Json.decodeFromString(cachedChaps)
+        return emptySet()
```
stevenyomi:
```suggestion
        val cachedChaps = preferences.getString(CACHE_KEY_CHAPTERS, null) ?: return emptySet()
        return Json.decodeFromString(cachedChaps)
```

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,241 @@
...
+
+    private fun fetchWebpageChapters(webpageIndex: Int): List<SerializableChapter> {
+        val url = "$baseUrl/comic/page/$webpageIndex/"
+        val resp = client.newCall(GET(url, headers)).execute()
+        val comics = resp.asJsoup().select(".comic")
+        return comics.mapNotNull {
+            val link = it.selectFirst(".post-title a")?.attr("href")
+            val title = it.selectFirst(".post-title a")?.text()
+            val postDate = it.selectFirst(".post-date")?.text()
+            val postTime = it.selectFirst(".post-time")?.text()
```
stevenyomi: Any reason they could be missing? If not, use `!!`.

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
```diff
@@ -0,0 +1,241 @@
...
+            title = "Show author's notes"
+            summary =
+                "Enable to see the author's notes at the end of chapters (if they're there)."
+            setDefaultValue(false)
+        }
+        screen.addPreference(authorsNotesPref)
+
+        // I couldn't find a way to create a simple button, so here's a workaround that uses
+        // a MultiSelectListPreference with a single option as a kind of confirmation window.
+        val clearCachePref = MultiSelectListPreference(screen.context).apply {
```
stevenyomi, line 212: Using `SwitchPreferenceCompat` is enough (just don't confirm). And you don't need to specify a key for `SwitchPreferenceCompat` if you use it like a button.

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: Done in d2da1c0b5dc049f4bd998511b6941cd3e54f775a

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: Probably not, but in case they do happen to be missing, should it really crash at runtime instead of handling it properly?

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: It works just as well, yes, but I feel like it's a less clear user experience.
This UX signifies that it's an option that toggles something, which it is not.
While using a pop-up, like I did previously, makes it feel completely seamless to the user.

I have tested this with some actual users, who all confirmed this.
I posted GIFs of both the toggle button (1), and the pop-up (2), which the question "What do you find clearer?"

> the second one because a toggle for something that isnt a toggle is weird

> By far the second option
> There is no world in which the first makes any sense
> I would consider option 1 broken software lol

The GIFs, for reference:
| Option 1 | Option 2|
|-----------|----------------|
| option1 | option2 |

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: Done in ac05b17e338f4d0a5ebccce441b9b4fbba025135

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
stevenyomi: It would cause the app to show an error instead, so you could identify the change on the website. Ignoring the error is not "handling it properly".

altaccosc: Fair enough :+1:

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
stevenyomi: Yeah but I'd prefer keeping the code as simple as possible. People don't visit the settings page very often, so it shouldn't need such a complicated workaround.

AwkwardPeak7: I would say having a good ux is important too

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: Done in 6cb092ddbb898d6163edb84230985b7ca40c5b6b

src/en/clowncorps/src/eu/kanade/tachiyomi/extension/en/clowncorps/ClownCorps.kt
altaccosc: I feel like we shouldn't compromise on UX, especially when the alternative option is as confusing as this.
And honestly, this code is not bad at all.
The only "odd" part is this:
```kotlin
setOnPreferenceChangeListener { _, newValue ->
    val checkValue = newValue as Set<*>
    if (checkValue.contains(VALUE_CONFIRM)) {
```
And this was adapted from how the MangaDex source does it: https://github.com/keiyoushi/extensions-source/blob/dfb59ebae0c8fa2b1a118e86b5bc5be8a183f3e6/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt#L684-L685

stevenyomi: Sure.

altaccosc:
I've been thinking... Wouldn't it be better if I stored the chapter cache outside the main reader app, and in the extension's own SharedPreferences instead?
That way, if the app gets reinstalled, the cache gets cleared.
I could have used this feature once or twice during development.
It's probably not very important, though.
Just a thought.

stevenyomi:
> I've been thinking... Wouldn't it be better if I stored the chapter cache outside the main reader app, and in the extension's own SharedPreferences instead?

It's not possible. The code is loaded and run in the app's context, so you don't have access to the extension's shared preferences.

PR #1921, opened 2024-03-16
migrate common.gradle to Kotlin DSL

AwkwardPeak7:
This pr is mainly for review

Some migrated `build.gradle` files include examples of:
- akuma -> regular extension
- coomer -> multisrc extension
- batoto -> implements a library dependency
- komga -> implements external dependency

Known issue(s):
- assert statments aren't working


stevenyomi:
We'll use one single extension function to register the metadata. Recorded in #1926 for now.

PR #2012, opened 2024-03-20
NhatTruyen/NetTruyen: new advanced search

cuong-tran:
Checklist:

### DIFF
--- a/src/vi/nettruyen/src/eu/kanade/tachiyomi/extension/vi/nettruyen/NetTruyen.kt
+++ b/src/vi/nettruyen/src/eu/kanade/tachiyomi/extension/vi/nettruyen/NetTruyen.kt
@@ -33,21 +33,7 @@ class NetTruyen : WPComics(
 
     // Advanced search
     override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
-        val advancedGenreFilter = filters.filterIsInstance<AdvancedGenreFilter>().firstOrNull()
-        val advancedStatusFilter = filters.filterIsInstance<AdvancedStatusFilter>().firstOrNull()
-        val chaptersNumFilter = filters.filterIsInstance<ChaptersNumFilter>().firstOrNull()
-        val genderFilter = filters.filterIsInstance<GenderFilter>().firstOrNull()
-        val orderFilter = filters.filterIsInstance<OrderFilter>().firstOrNull()
-
-        if (query.isBlank() &&
-            (
-                advancedGenreFilter?.state?.any { it.isIncluded() || it.isExcluded() } == true ||
-                    advancedStatusFilter != null && advancedStatusFilter.state > 0 ||
-                    chaptersNumFilter != null && chaptersNumFilter.state > 0 ||
-                    genderFilter != null && genderFilter.state > 0 ||
-                    orderFilter != null && orderFilter.state > 0
-                )
-        ) {
+        if (query.isBlank()) {
             val url = "$baseUrl/tim-truyen-nang-cao".toHttpUrl().newBuilder()
 
             filters.forEach { filter ->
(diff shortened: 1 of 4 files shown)


src/vi/nettruyen/src/eu/kanade/tachiyomi/extension/vi/nettruyen/NetTruyen.kt
```diff
@@ -18,9 +25,163 @@ class NetTruyen : WPComics(
...
+        val orderFilter = filters.filterIsInstance<OrderFilter>().firstOrNull()
+
+        if (query.isBlank() &&
+            (
+                advancedGenreFilter?.state?.any { it.isIncluded() || it.isExcluded() } == true ||
+                    advancedStatusFilter != null && advancedStatusFilter.state > 0 ||
+                    chaptersNumFilter != null && chaptersNumFilter.state > 0 ||
+                    genderFilter != null && genderFilter.state > 0 ||
+                    orderFilter != null && orderFilter.state > 0
+                )
```
AwkwardPeak7: why such a convoluted logic? just put default values if the filter isn't applied

AwkwardPeak7 (changes requested):
can this be adjusted in wpcomics?
if only these two sources support it then maybe add another class to wpcomics like `WpComicsAlt` with this support and use that in these sources

src/vi/nettruyen/src/eu/kanade/tachiyomi/extension/vi/nettruyen/NetTruyen.kt
cuong-tran: There are currently 2 working search methods:
- Old method: query string with status & genre selection
- New advanced method: genre tristate, gender, chapter number.. but won't support query.

This code is intended to invoke the new advanced search method if any of the new filters are applied, fall-back to old search if no new filter is used. On the other hand, if the user doesn't enter a query string but still uses old search's filters, we can't decide the search method solely depends on the query string alone.

AwkwardPeak7: does the new advanced method on the site doesn't support filters that are available in old method?

cuong-tran: Only the genre & status filter is common between 2 methods, but the value used for genre filters are different. Also some genres is available in the new search but not in the old one. I don't think it's possible to merge 2 methods:
- Historical:
  - Old: Name: Historical - Value: historical
  - New: Name: Lich su - Value: 17
- Yuri: only in new search
- Yaoi: only in new search

cuong-tran:
> can this be adjusted in wpcomics? if only these two sources support it then maybe add another class to wpcomics like `WpComicsAlt` with this support and use that in these sources

Sure that can be done. I was just afraid that changing to multi-src is not preferred.

AwkwardPeak7:
> Sure that can be done. I was just afraid that changing to multi-src is not preferred.

none of the other sites support it? if so should be fine to leave it as is

src/vi/nettruyen/src/eu/kanade/tachiyomi/extension/vi/nettruyen/NetTruyen.kt
AwkwardPeak7: imo this would be confusing for the users, which filter works in which condition...

cuong-tran:
> > Sure that can be done. I was just afraid that changing to multi-src is not preferred.
> 
> none of the other sites support it? if so should be fine to leave it as is

I would say just not yet. NetTruyen and NhatTruyen only supports this very recently. Other sites might follow later.

src/vi/nettruyen/src/eu/kanade/tachiyomi/extension/vi/nettruyen/NetTruyen.kt
cuong-tran: I agree it does look confused. How about add a string to clarify like this:
image

And simplify condition to only use old search if query string is entered.

AwkwardPeak7 (approved):

PR #2303, opened 2024-04-08
Gmanga multisrc: filter out novel & fix null

cuong-tran:
Checklist:

### DIFF
--- a/lib-multisrc/gmanga/src/eu/kanade/tachiyomi/multisrc/gmanga/Dto.kt
+++ b/lib-multisrc/gmanga/src/eu/kanade/tachiyomi/multisrc/gmanga/Dto.kt
@@ -84,7 +84,7 @@ class Manga(
             else -> SManga.UNKNOWN
         }
         genre = buildList {
-            add(type.title)
+            add(type.title ?: "")
             add(type.name)
             categories.forEach { add(it.name) }
         }.joinToString()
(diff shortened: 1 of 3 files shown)


lib-multisrc/gmanga/src/eu/kanade/tachiyomi/multisrc/gmanga/Dto.kt
```diff
@@ -123,7 +124,7 @@ class NameDto(val name: String)
 @Serializable
 class TypeDto(
     val name: String,
-    val title: String,
+    val title: String?,
```
AwkwardPeak7, line 127: also check for nullability when adding to genres, so "null" doesn't end up in the genre string

lib-multisrc/gmanga/src/eu/kanade/tachiyomi/multisrc/gmanga/Dto.kt
```diff
@@ -83,7 +84,7 @@ class Manga(
             else -> SManga.UNKNOWN
         }
         genre = buildList {
-            add(type.title)
+            add(type.title ?: "")
```
AwkwardPeak7:
```suggestion
            type.title?.let { add(it) }
```

AwkwardPeak7 (approved):

PR #2862, opened 2024-05-05
Add Mangamo

TheUnlocked:
Checklist:

This one is quite a bit more complicated than most since there's no web client to scrape so everything has to make direct API calls. I've done quite a bit of testing and it seems to work well, ignoring a couple minor cases where limitations on the Firestore API can cause issues with pagination.


src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/Mangamo.kt
```diff
@@ -0,0 +1,373 @@
...
+
+    override fun getMangaUrl(manga: SManga): String {
+        return manga.url
+    }
+
+    override fun mangaDetailsRequest(manga: SManga): Request {
+        val uri = manga.url.toHttpUrl()
+
+        val seriesId = uri.queryParameter(MangamoConstants.SERIES_QUERY_PARAM)?.toInt()
+            ?: throw Exception("${manga.url} missing series param. This shouldn't happen")
```
AwkwardPeak7:
```suggestion
        val seriesId = uri.queryParameter(MangamoConstants.SERIES_QUERY_PARAM)!!.toInt()
```

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/Mangamo.kt
```diff
@@ -0,0 +1,373 @@
...
+        return processSeries(dto)
+    }
+
+    // Chapter list section
+
+    override fun fetchChapterList(manga: SManga): Observable<List<SChapter>> {
+        val uri = manga.url.toHttpUrl()
+
+        val seriesId = uri.queryParameter(MangamoConstants.SERIES_QUERY_PARAM)?.toInt()
+            ?: throw Exception("${manga.url} missing series param. This shouldn't happen")
```
AwkwardPeak7:
```suggestion
        val seriesId = uri.queryParameter(MangamoConstants.SERIES_QUERY_PARAM)!!.toInt()
```

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/Mangamo.kt
```diff
@@ -0,0 +1,373 @@
...
+
+    override fun pageListRequest(chapter: SChapter): Request {
+        val uri = chapter.url.toHttpUrl()
+
+        val seriesId = uri.queryParameter(MangamoConstants.SERIES_QUERY_PARAM)?.toInt()
+        val chapterId = uri.queryParameter(MangamoConstants.CHAPTER_QUERY_PARAM)?.toInt()
+
+        if (seriesId == null || chapterId == null) {
+            throw java.lang.Exception("${chapter.url} missing param. This shouldn't happen")
+        }
```
AwkwardPeak7: this too

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/MangamoHelper.kt
```diff
@@ -0,0 +1,80 @@
...
+            }
+        }
+
+        @Suppress("UNCHECKED_CAST")
+        inline fun <reified T> String.parseJson(): T {
+            return when (T::class) {
+                DocumentDto::class -> json.decodeFromString<T>(
+                    DocumentSerializer(serializer<T>() as KSerializer<out DocumentDto<out Any?>>) as KSerializer<T>,
+                    this,
+                )
```
AwkwardPeak7, line 36: I belive you can add ` @Serializable(with = DocumentSerializer::class)` to `DocumentDto` override the default serializer

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/dto/ChapterDto.kt
```diff
@@ -0,0 +1,20 @@
+package eu.kanade.tachiyomi.extension.en.mangamo.dto
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class ChapterDto(
```
AwkwardPeak7: don't serialize values you are not using

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/Mangamo.kt
```diff
@@ -0,0 +1,373 @@
...
+        .build()
+
+    private fun processSeries(dto: SeriesDto) = SManga.create().apply {
+        author = dto.authors?.joinToString { it.name }
+        description = dto.description
+        genre = dto.genres?.joinToString { it.name }
+        status = helper.getSeriesStatus(dto)
+        thumbnail_url = dto.titleArt
+        title = dto.name!!
+        url = helper.getSeriesUrl(dto)
```
AwkwardPeak7, line 133: add `initialized = true` to avoide an extra call for details.

since it appears all the information is already available during browsing

AwkwardPeak7 (changes requested):
Good effort, I mostly skimmed through the code, will try to review properly later

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/MangamoHelper.kt
TheUnlocked: That was what I had initially tried, but unfortunately it didn't work because `DocumentDto` is generic and `JsonTransformingSerializer` doesn't play very nicely with that. I spent quite a while trying to get serialization to work properly before I ended up with this solution.

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/dto/ChapterDto.kt
TheUnlocked: Commented out unused fields.

TheUnlocked:
Addressed all of the issues so far. While I was at it I also optimized queries by only requesting the fields that are used, and found a race condition in the authorization code and fixed that.

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/dto/ChapterDto.kt
AwkwardPeak7: I would prefer removing them

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/dto/DocumentDto.kt
```diff
@@ -0,0 +1,64 @@
...
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonTransformingSerializer
+import kotlinx.serialization.json.jsonArray
+import kotlinx.serialization.json.jsonObject
+
+@Serializable
+data class DocumentWrapper<T>(val document: DocumentDto<T>?)
```
AwkwardPeak7: also, make them regular classes, no need for data class

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/cachedBy.kt
```diff
@@ -0,0 +1,22 @@
+package eu.kanade.tachiyomi.extension.en.mangamo
+
+import kotlin.reflect.KProperty
+
+@Suppress("ClassName")
+class cachedBy<T>(private val dependencies: () -> Any?, private val callback: () -> T) {
```
AwkwardPeak7, line 6: why is this necessary?

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/Mangamo.kt
```diff
@@ -0,0 +1,435 @@
...
+                        chapter_number = chapter.chapterNumber
+                        date_upload = chapter.createdAt!!
+                        name = chapter.name +
+                            if (isFreeChapter) {
+                                " (Free)"
+                            } else if (isCoinChapter) {
+                                " (Coin)"
+                            } else {
+                                ""
+                            }
```
AwkwardPeak7: imo, add a lock emoji on paid chapters and/or coin empji on coin chapters; free chapters shouldn't have any indicator

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/Mangamo.kt
```diff
@@ -0,0 +1,435 @@
...
+                return@parseMangaPage false
+            }
+        }
+        true
+    }
+
+    // Manga details
+
+    override fun getMangaUrl(manga: SManga): String {
+        return manga.url
```
AwkwardPeak7: would be better to store relative urls in the database

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/Mangamo.kt
```diff
@@ -0,0 +1,435 @@
...
+        return getPagesImagesRequest(seriesId, chapterId)
+    }
+
+    override fun pageListParse(response: Response): List<Page> {
+        val data = response.body.string().parseJson<List<PageDto>>()
+
+        return data.map {
+            Page(it.pageNumber - 1, imageUrl = it.uri)
+        }.sortedBy { it.index }
+    }
```
AwkwardPeak7: should throw custom exception when user tries reading paid chapters without payment; maybe in an interceptor detect http 401?

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/MangamoHelper.kt
AwkwardPeak7: looks like you are only using `SeriesDto` as type argument for `DocumentDto`, so should be fine to simplify it

TheUnlocked: It's also used by `QueryResultDto<T>`, which is used for both chapters and series. As of the most recent change, `DocumentDto` is also being directly used with `UserDto`.

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/cachedBy.kt
TheUnlocked: It allows properties to lazily update in response to changes in other properties. For example, it allows `auth` in `Mangamo.kt` to automatically re-create a `MangamoAuth` object and regenerate tokens when the user changes their user token in the settings.

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/Mangamo.kt
TheUnlocked: Added phrasing for this. It's a little more vague than talking about payment since there are a number of reasons someone might not be able to access a chapter, but it is more comprehensible than seeing a 401.

src/en/mangamo/src/eu/kanade/tachiyomi/extension/en/mangamo/Mangamo.kt
TheUnlocked: Converted to relative URLs

AwkwardPeak7 (approved):

PR #2971, opened 2024-05-11
Bato.to: Add deep link support for v3x

Bartuzen:
Checklist:

### DIFF
--- a/src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoToUrlActivity.kt
+++ b/src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoToUrlActivity.kt
@@ -42,8 +42,25 @@ class BatoToUrlActivity : Activity() {
 
     private fun fromBatoTo(pathSegments: MutableList<String>): String? {
         return if (pathSegments.size >= 2) {
-            val id = pathSegments[1]
-            "ID:$id"
+            val path = pathSegments[1] as java.lang.String?
+            if (path != null) {
+                var index = -1
+                for (i in path.indices) {
+                    if (path[i] == '-') {
+                        index = i
+                        break
+                    }
+                }
+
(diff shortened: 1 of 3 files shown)


Bartuzen:
I explained this in Discord but let me reiterate it here: Kotlin classes like Intrinsics and StringsKt don't work in URL activities as explained [here](https://github.com/keiyoushi/extensions-source/blob/main/CONTRIBUTING.md#url-intent-filter). So I had to make sure compiled code doesn't use those classes. That's why the code looks weird.

src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoToUrlActivity.kt
```diff
@@ -42,8 +42,25 @@ class BatoToUrlActivity : Activity() {
 
     private fun fromBatoTo(pathSegments: MutableList<String>): String? {
         return if (pathSegments.size >= 2) {
-            val id = pathSegments[1]
-            "ID:$id"
+            val path = pathSegments[1] as java.lang.String?
```
AwkwardPeak7, line 45: wouldn't `pathSegment[1]` already be `String`? so why cast

Bartuzen: Because if I remove the cast or cast it to String?, it uses Intrinsics class:
```java
if (index == -1) {
    var10000 = path;
} else {
    var10000 = path.substring(0, index);
    Intrinsics.checkNotNullExpressionValue(var10000, "this as java.lang.String…ing(startIndex, endIndex)");
}

String id = var10000;
```
If I cast it to java.lang.String?, it doesn't use Intrinsics class:
```java
Object id = index == -1 ? path : path.substring(0, index);
```

AwkwardPeak7 (approved):

PR #3016, opened 2024-05-14
ci: create repo `index.html` lising apk

aaronisles:
I'd like to use https://github.com/ImranR98/Obtainium to track and install these extensions. For this all we need is a directory listing rendered in html index with links to the apk files.

Unfortunately the directory listing on [github](https://github.com/keiyoushi/extensions/tree/repo/apk/) and [jsdelivr](https://cdn.jsdelivr.net/gh/keiyoushi/extensions@repo/apk/) cause errors because of the large number of apk files.

The proposed change creates a listing of the apk files at `index.html` in ci.

Checklist:

### DIFF
--- a/.github/scripts/commit-repo.sh
+++ b/.github/scripts/commit-repo.sh
@@ -11,7 +11,6 @@ if [ -n "$(git status --porcelain)" ]; then
     git push
 
     curl https://purge.jsdelivr.net/gh/keiyoushi/extensions@repo/index.min.json
-    curl https://purge.jsdelivr.net/gh/keiyoushi/extensions@repo/index.html
 else
     echo "No changes to commit"
 fi
--- a/.github/scripts/commit-repo.sh
+++ b/.github/scripts/commit-repo.sh
@@ -11,6 +11,7 @@ if [ -n "$(git status --porcelain)" ]; then
     git push
 
     curl https://purge.jsdelivr.net/gh/keiyoushi/extensions@repo/index.min.json
+    curl https://purge.jsdelivr.net/gh/keiyoushi/extensions@repo/index.html
 else
     echo "No changes to commit"
 fi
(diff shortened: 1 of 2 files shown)


AwkwardPeak7:
extensions are listed on the site too, check if that works for you

https://keiyoushi.github.io/extensions/

aaronisles:
Unfortunately it doesn't. https://keiyoushi.github.io/extensions/ uses javascript to fetch it's data while Obtainium expects the link rendered in the HTML.

MikeZeDev:
the linked page just read the extension list from JSON, so its static.

https://raw.githubusercontent.com/keiyoushi/extensions/repo/index.json

Not sure it helps if you want html.

aaronisles:
Yes, Obtainium wouldn't work with JSON. It can work with HTML with links to the apk. It can also work with a direct link to the apk (without an intermediate HTML) if the link stays the same across versions.

.github/scripts/commit-repo.sh
```diff
@@ -11,6 +11,7 @@ if [ -n "$(git status --porcelain)" ]; then
     git push
 
     curl https://purge.jsdelivr.net/gh/keiyoushi/extensions@repo/index.min.json
+    curl https://purge.jsdelivr.net/gh/keiyoushi/extensions@repo/index.html
```
AwkwardPeak7:
```suggestion
```
since this file isn't something that will be used by many people, no need to force clear jsdelivr cache

.github/scripts/create-repo.py
```diff
@@ -106,3 +107,10 @@
 
 with (REPO_DIR / "index.min.json").open("w", encoding="utf-8") as f:
     json.dump(index_min_data, f, ensure_ascii=False, separators=(",", ":"))
+
+with (REPO_DIR / "index.html").open("w", encoding="utf-8") as f:
+    f.write('<!DOCTYPE html>\n<html>\n<head>\n<meta charset="UTF-8">\n<title>apks</title>\n</head>\n<body>\n<pre>\n')
+    for entry in index_data:
+        apk_escaped = 'apk/' + html.escape(entry["apk"])
+        f.write(f'<a href="{apk_escaped}">{apk_escaped}</a>\n')
```
AwkwardPeak7: nitpick: put the `apk/{}` url in href attribute but keep apk name as the link name

AwkwardPeak7 (approved):

PR #3051, opened 2024-05-16
Fix no description NettruyenCO and Update domain NettruyenCO

dejavui:
Checklist:


AwkwardPeak7:
the first link still works for me. and both of them have different google tag id, so are they related?

dejavui:
> the first link still works for me. and both of them have different google tag id, so are they related?

I only know NettruyenCO. VN redirects to Nettruyencc. com when i open webview :v

AwkwardPeak7 (approved):

PR #3189, opened 2024-05-22
WPComics: Allow change base url

hajaulee:
Checklist:


AwkwardPeak7 (changes requested):
not all `WPComics` sources need customzied baseurl.
so implement it in any specific source you need. plus having the prefernece orphaned on version update isn't desirable so implement [like so](https://github.com/keiyoushi/extensions-source/blob/main/src/zh/zerobyw/src/eu/kanade/tachiyomi/extension/zh/zerobyw/UpdateUrl.kt#L24)

PR #3219, opened 2024-05-24
Rewrote LibGroup to use api instead of parsing document

### LINKED ISSUE #2660: Adding new tags to Hentailib filters
### Source name

HentaiLib

### Source language

Russian

Please add "лоликон" and "сетакон" new tags to the filters on hentailib. These tags are present on the site.
markup_1000063681
markup_1000063680

### Other details

_No response_

### Acknowledgements

### LINKED ISSUE #2998: Hentailib parser not working after redesign
### Source information

HentaiLib 1.4.44

### Source language

Russian

### Steps to reproduce

I just open the extension

### Expected behavior

when you open it, it should open a list of titles.

### Actual behavior

It throws a NullPointerException error

### Mihon/Tachiyomi version

0.16.4

### Android version

Android 14

### Other details

Since HentaiLib has undergone a redesign, the parser needs to be updated for the new version. YaoiLib will also get a new design next week ([news source](https://lib.social/news/obnovlenie-prilozeniya-i-skoroe-obnovlenie-saita)). You can also take care of this issue in parallel: #2660

### Acknowledgements

nedius:
Closes #2998 
Closes #2660 

Changes MangaLib, HentaiLib and YaoiLib

Requires migration of each source to itself
May get extension settings resets

Checklist:


lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +24,203 @@ import eu.kanade.tachiyomi.source.model.Page
...
-    private val json: Json by injectLazy()
+    private val json: Json = Json {
+        ignoreUnknownKeys = true
+        explicitNulls = false
+        encodeDefaults = true
+    }
 
     private val preferences: SharedPreferences by lazy {
-        Injekt.get<Application>().getSharedPreferences("source_${id}_2", 0x0000)
+        Injekt.get<Application>().getSharedPreferences("source_$id", 0x0000)
```
AwkwardPeak7, line 60: consider migrating old preferences instead of making a new preference store,

you can migrate them like [this](https://github.com/keiyoushi/extensions-source/blob/main/src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt#L68-L71) and [this](https://github.com/keiyoushi/extensions-source/blob/main/src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt#L481-L498)

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +24,203 @@ import eu.kanade.tachiyomi.source.model.Page
...
+        } else {
+            getToken().subscribe {
+                if (it != null) {
+                    val str: String = if (it.first() == '"' && it.last() == '"') {
+                        it.substringAfter("\"").substringBeforeLast("\"").replace("\\", "")
+                    } else {
+                        it.replace("\\", "")
+                    }
+                    val token = json.decodeFromString<JsonObject>(str).jsonObject["token"]
+                    bearerToken = token!!.jsonObject["token_type"]!!.jsonPrimitive.content + " " + token.jsonObject["access_token"]!!.jsonPrimitive.content
```
AwkwardPeak7: add token in an interceptor to make sure token gets applied after fetching

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +24,203 @@ import eu.kanade.tachiyomi.source.model.Page
...
         }
-        return fetchLatestMangaFromApi(page)
     }
 
-    private fun fetchLatestMangaFromApi(page: Int): Observable<MangasPage> {
-        return client.newCall(POST("$baseUrl/filterlist?dir=desc&sort=last_chapter_at&page=$page&chapters[min]=1", catalogHeaders()))
-            .asObservableSuccess()
-            .map { response ->
-                latestUpdatesParse(response)
+    private fun apiHeaders(): Headers = headersBuilder().build()
```
AwkwardPeak7: just use `headers` instead?

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +24,203 @@ import eu.kanade.tachiyomi.source.model.Page
...
+
+    private var _constants: LibGroupConstantsDto? = null
+    private fun getConstants(): LibGroupConstantsDto {
+        if (_constants == null) {
+            try {
+                _constants = client.newCall(
+                    GET("https://api.$apiDomain/api/constants?fields[]=genres&fields[]=tags&fields[]=types&fields[]=scanlateStatus&fields[]=status&fields[]=format&fields[]=ageRestriction&fields[]=imageServers", apiHeaders()),
+                ).execute().body.string().substringAfter("\"data\":").substringBeforeLast("}").parseAs<LibGroupConstantsDto>()
```
AwkwardPeak7: do proper json parsing instead of substrings

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +24,203 @@ import eu.kanade.tachiyomi.source.model.Page
...
+                return _constants as LibGroupConstantsDto
```
AwkwardPeak7:
```suggestion
                return _constants!!
```

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +24,203 @@ import eu.kanade.tachiyomi.source.model.Page
...
+            } catch (ex: SerializationException) {
+                throw Exception("Ошибка сериализации. Проверьте сайт.")
             }
+        }
+        return _constants as LibGroupConstantsDto
```
AwkwardPeak7:
```suggestion
        return _constants!!
```

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +24,203 @@ import eu.kanade.tachiyomi.source.model.Page
...
-            return client.newCall(popularMangaRequest(page))
-                .asObservableSuccess()
-                .flatMap { response ->
-                    // Obtain token
-                    val resBody = response.body.string()
-                    csrfToken = "_token\" content=\"(.*)\"".toRegex().find(resBody)!!.groups[1]!!.value
-                    return@flatMap fetchPopularMangaFromApi(page)
+    @SuppressLint("SetJavaScriptEnabled")
+    @Suppress("NAME_SHADOWING")
+    private fun getToken(): Observable<String?> = Single.create<String?> { subscriber ->
```
AwkwardPeak7: you can use coroutine scope here too, instead of rxjava

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -304,226 +229,120 @@ abstract class LibGroup(
...
         }
-
-        val urls = page.url.split(",,")
-
-        return Observable.from(urls).filter { checkImage(it) }.first()
+        // api changed id of servers, remap old to new
+        if (isServer == "fourth") {
+            isServer = "secondary"
+            preferences.edit().putString(SERVER_PREF, isServer).apply()
+        }
```
AwkwardPeak7: migrate preferences on init as mentioned previously

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -716,38 +493,51 @@ abstract class LibGroup(
         arrayOf("Да", "Все"),
     )
 
+    // Utils
+    private inline fun <reified T> String.parseAs(): T = json.decodeFromString(this)
+
+    private inline fun <reified T> Response.parseAs(): T = body.string().parseAs()
```
AwkwardPeak7, line 528: remove usage of `json.decodeFromString` since you have this function

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroupDto.kt
AwkwardPeak7, line 1: no need for `data class` regular classes should suffice

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroupDto.kt
```diff
@@ -0,0 +1,283 @@
...
+            title = getSelectedLanguage(isEng, rusName, engName, name)
+            thumbnail_url = cover.default.orEmpty()
+            url = "/$slugUrl"
+        }
+    }
+}
+
+@Serializable
+data class MangaDetailsDto(
+    val data: Manga,
```
AwkwardPeak7: make a template class for data, `class Data<T>(val data: T)`

src/ru/hentailib/AndroidManifest.xml
AwkwardPeak7, line 1: move the manifest to multisrc, since all of them are similar (check madara for SOURCEHOST usage)

src/ru/hentailib/src/eu/kanade/tachiyomi/extension/ru/hentailib/HentaiLib.kt
```diff
@@ -5,227 +5,30 @@ import android.content.SharedPreferences
...
-import eu.kanade.tachiyomi.network.POST
-import eu.kanade.tachiyomi.source.model.Filter
 import eu.kanade.tachiyomi.source.model.FilterList
-import okhttp3.Request
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
 
 class HentaiLib : LibGroup("HentaiLib", "https://hentailib.me", "ru") {
 
-    override val id: Long = 6425650164840473547
```
AwkwardPeak7, line 17: keep id since you added url change exceptions

AwkwardPeak7 (changes requested):

nedius:
Regardless id 
Old id implementation wasn't working right .
If set id explicitly there will be 2 distinct sharedPreferencef instances,
one `source_0_2` and another `source_{source implementation id}`

Because when `LibGroup.preferences` initializes id is set to 0 (???) 
So because of that we have this 
image
And variables was always reverting to default

I checked and id is not mandatory, here is all 3 sources id automatically generated
MangaLib: 6111047689498497237
HentaiLib: 425650164840473547
YaoiLib: 2730544188738947015

e-shl:
Thank for taking up the development of extensions. Maybe it will help in some way https://github.com/keiyoushi/extensions-source/pull/3241

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +23,203 @@ import eu.kanade.tachiyomi.source.model.Page
...
-                    // Obtain token
-                    val resBody = response.body.string()
-                    csrfToken = "_token\" content=\"(.*)\"".toRegex().find(resBody)!!.groups[1]!!.value
-                    return@flatMap fetchLatestMangaFromApi(page)
-                }
+    private fun checkForToken(chain: Interceptor.Chain): Response {
+        val req = chain.request().newBuilder()
+        if (chain.request().url.toString().contains("api.$apiDomain")) {
+            if (bearerToken.isNullOrBlank()) {
+                launchIO { getToken() }
```
AwkwardPeak7: don't launch in background

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +23,203 @@ import eu.kanade.tachiyomi.source.model.Page
...
     }
 
-    private fun fetchLatestMangaFromApi(page: Int): Observable<MangasPage> {
-        return client.newCall(POST("$baseUrl/filterlist?dir=desc&sort=last_chapter_at&page=$page&chapters[min]=1", catalogHeaders()))
-            .asObservableSuccess()
-            .map { response ->
-                latestUpdatesParse(response)
+    @SuppressLint("SetJavaScriptEnabled")
+    @Suppress("NAME_SHADOWING")
+    private fun getToken() {
```
AwkwardPeak7: consider changing it to [this implementation](https://github.com/keiyoushi/extensions-source/pull/3241/files#diff-b2fcea53275f77fafad564c13e2be399de15e208fd405ac4c00d3e0d2681af4bR94-R100)
as webview should be avoided where possible

@e-shl as [co-author](https://coauthoredby.netlify.app/)

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +23,203 @@ import eu.kanade.tachiyomi.source.model.Page
...
-                    csrfToken = "_token\" content=\"(.*)\"".toRegex().find(resBody)!!.groups[1]!!.value
-                    return@flatMap fetchPopularMangaFromApi(page)
-                }
-        }
-        return fetchPopularMangaFromApi(page)
+    // Latest
+    override fun latestUpdatesRequest(page: Int): Request {
+        val url = "https://api.$apiDomain/api/latest-updates".toHttpUrl().newBuilder()
+            .addQueryParameter("page", page.toString())
+        return GET(url.build(), headersBuilder().build())
```
AwkwardPeak7:
```suggestion
        return GET(url.build(), headers)
```

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +23,203 @@ import eu.kanade.tachiyomi.source.model.Page
...
-                popularMangaParse(response)
-            }
+    override fun latestUpdatesParse(response: Response): MangasPage = popularMangaParse(response)
+
+    // Popular
+    override fun popularMangaRequest(page: Int): Request {
+        val url = "https://api.$apiDomain/api/manga".toHttpUrl().newBuilder()
+            .addQueryParameter("site_id[]", siteId.toString())
+            .addQueryParameter("page", page.toString())
+        return GET(url.build(), headersBuilder().build())
```
AwkwardPeak7:
```suggestion
        return GET(url.build(), headers)
```

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +23,203 @@ import eu.kanade.tachiyomi.source.model.Page
...
+            .addQueryParameter("fields[]", "tags")
+            .addQueryParameter("fields[]", "teams")
+            .addQueryParameter("fields[]", "authors")
+            .addQueryParameter("fields[]", "publisher")
+            .addQueryParameter("fields[]", "userRating")
+            .addQueryParameter("fields[]", "manga_status_id")
+            .addQueryParameter("fields[]", "status_id")
+            .addQueryParameter("fields[]", "artists")
+
+        return GET(url.build(), headersBuilder().build())
```
AwkwardPeak7:
```suggestion
        return GET(url.build(), headers)
```

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -304,226 +228,106 @@ abstract class LibGroup(
...
+        // throw exception if old url
+        if (!manga.url.contains("--")) throw Exception(urlChangedError(name))
+
+        return GET("https://api.$apiDomain/api/manga${manga.url}/chapters", headersBuilder().build())
+    }
+
+    override fun getChapterUrl(chapter: SChapter): String = "$baseUrl${chapter.url}"
+
+    private fun getDefaultBranch(id: String): List<Branch> =
+        client.newCall(GET("https://api.$apiDomain/api/branches/$id", headersBuilder().build())).execute().parseAs<Data<List<Branch>>>().data
```
AwkwardPeak7:
```suggestion
        client.newCall(GET("https://api.$apiDomain/api/branches/$id", headers))
            .execute().parseAs<Data<List<Branch>>>().data
```

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -553,44 +354,62 @@ abstract class LibGroup(
...
+                is TagList -> filter.state.forEach { tag ->
+                    if (tag.state != Filter.TriState.STATE_IGNORE) {
+                        url.addQueryParameter(if (tag.isIncluded()) "tags[]" else "tags_exclude[]", tag.id)
                     }
                 }
                 else -> {}
             }
         }
-        return POST(url.toString(), catalogHeaders())
+        return GET(url.build(), headersBuilder().build())
```
AwkwardPeak7:
```suggestion
        return GET(url.build(), headers)
```

src/ru/hentailib/build.gradle
```diff
@@ -3,7 +3,7 @@ ext {
     extClass = '.HentaiLib'
     themePkg = 'libgroup'
     baseUrl = 'https://hentailib.me'
-    overrideVersionCode = 19
+    overrideVersionCode = 20
```
AwkwardPeak7: no need to bump when you are bumping the multisrc

src/ru/yaoilib/src/eu/kanade/tachiyomi/extension/ru/yaoilib/YaoiLib.kt
```diff
@@ -5,14 +5,10 @@ import android.content.SharedPreferences
...
 import eu.kanade.tachiyomi.multisrc.libgroup.LibGroup
-import eu.kanade.tachiyomi.network.POST
-import eu.kanade.tachiyomi.source.model.Filter
-import eu.kanade.tachiyomi.source.model.FilterList
-import okhttp3.Request
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
 
-class YaoiLib : LibGroup("YaoiLib", "https://v2.slashlib.me", "ru") {
+class YaoiLib : LibGroup("YaoiLib", "https://slashlib.me/", "ru") {
```
AwkwardPeak7:
```suggestion
class YaoiLib : LibGroup("YaoiLib", "https://slashlib.me", "ru") {
```

src/ru/yaoilib/build.gradle
```diff
@@ -3,7 +3,7 @@ ext {
     extClass = '.YaoiLib'
     themePkg = 'libgroup'
     baseUrl = 'https://v2.slashlib.me'
```
AwkwardPeak7:
```suggestion
    baseUrl = 'https://slashlib.me'
```

AwkwardPeak7 (changes requested):

AwkwardPeak7:
> I checked and id is not mandatory, here is all 3 sources id automatically generated MangaLib: 6111047689498497237 HentaiLib: 425650164840473547 YaoiLib: 2730544188738947015

you should either change id or add url change exception. changing id can confuse users as the library entries are no longer linked with the source

nedius:
Marking draft until `getToken` situation resolved

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
nedius: @e-shl implementation will not work because api endpoint expects bearer token, obtained by OAuth and saved in `localStorage['auth'].token` on log in.
So getting token from webview inevitable.
The problem is i don't know how to work with coroutines on android (sorry).

Token in `localStorage` are in format `{access_token: String, expires_in: Int, refresh_token: String, token_type: String}`
So i can propose saving `access_token`, `expires_in`, `timestamp` (login time) and `token_type` (incase site will change token type), and get new token when it expires 

I implemented `getConstants()` to prevent issues like this #2660 in future

nedius:
> > I checked and id is not mandatory, here is all 3 sources id automatically generated MangaLib: 6111047689498497237 HentaiLib: 425650164840473547 YaoiLib: 2730544188738947015
> 
> you should either change id or add url change exception. changing id can confuse users as the library entries are no longer linked with the source

TL;DR
```
Removing id definition from sources does not change them, and will not confuse users. Migration still necessary 
```

I am actually not changing id, but removing redundant declaration of extensions id that meses up initialization of `LibGroup.preferences` and extension local `preferences`

If define preferences and extension id's 
For example in LibGroup.kt
```kt
  private val preferences: SharedPreferences by lazy {
      Injekt.get<Application>().getSharedPreferences("source_${id}_2", 0x0000)
  }
```
And this in MangaLib.kt
```kt
    override val id: Long = 6111047689498497237

    private val preferences: SharedPreferences by lazy {
        Injekt.get<Application>().getSharedPreferences("source_${id}_2", 0x0000)
    }
```
We get 2 instances of preferences in `this` scope: 
image
With `LibGroup.preferences` file path set to 
image
And Local instance with file path 
image
And end file in `/data/user/0/app.mihon/shared_prefs/source_6111047689498497237.xml`
image

```
Breakpoint was set in popularMangaParse for testing and making screenshots so it is after classes init
```

There is no `source_${id}_2` only `source_$id`
App saves settings in `source_$id`
So loading preferences back after app restart loads default declared values for variables..

**Is this intended behavior ?**

So for saving variables to file mihon uses it's local `preferences` variable (Is this Mihon bug ?)

I am saying that, explicitly set id for extension does not change they're id's and does not invalidate internal mihon id's for already added mangas.
Checked with my personal phone and "production environment" mihon. Mihon recognized new extension without saying source doesn't installed.
And settings are properly loading from old version of extension.

---

Yes migration is still necessary, but because site changed manga `slug`s to `slug_url`s,
so its not `i-alone-level-up` but `7580--i-alone-level-up`
```
"${slug}" -> "${manga_id}--${slug}"
```

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -716,38 +478,51 @@ abstract class LibGroup(
...
 
-        private val simpleDateFormat by lazy { SimpleDateFormat("yyyy-MM-dd", Locale.US) }
+        val simpleDateFormat by lazy { SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSSSS'Z'", Locale.US) }
     }
 
-    private var isServer: String? = preferences.getString(SERVER_PREF, "fourth")
+    private var isServer: String? = preferences.getString(SERVER_PREF, "main")
     private var isEng: String? = preferences.getString(LANGUAGE_PREF, "eng")
     private var groupTranslates: String = preferences.getString(TRANSLATORS_TITLE, TRANSLATORS_DEFAULT)!!
+    private var isScanUser: Boolean? = preferences.getBoolean(IS_SCAN_USER, false)
```
bapeey: this can cause `preferences` to be initialized before the id (id = 0) if it is overridden in the source class, change it to a function instead

AwkwardPeak7:
> TL;DR
> 
> ```
> Removing id definition from sources does not change them, and will not confuse users. Migration still necessary 
> ```

I was under the impression hard coded id was different than the generated one, sorry all good.

as bapeey said, accessing preference object before id is calculated causes this behaviour you described

AwkwardPeak7:
@nedius about the token thing, I was saying don't run it in `launchIO` in the interceptor, let it run blocking. this way it will add the token to the very first request too.

nedius:
For now webview will be loaded (for getting token) on first network call, if it fails it will not retry until app restart.

Api gives token with expire date so it's possible to check for that. I can implement saving token and its metadata in preferences. Native site js code is needed to update token if it's expires.
Is it safe?

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +22,213 @@ import eu.kanade.tachiyomi.source.model.Page
...
-                    val resBody = response.body.string()
-                    csrfToken = "_token\" content=\"(.*)\"".toRegex().find(resBody)!!.groups[1]!!.value
-                    return@flatMap fetchPopularMangaFromApi(page)
-                }
-        }
-        return fetchPopularMangaFromApi(page)
+    // Latest
+    override fun latestUpdatesRequest(page: Int): Request {
+        val url = "https://api.$apiDomain/api/latest-updates".toHttpUrl().newBuilder()
+            .addQueryParameter("page", page.toString())
```
AwkwardPeak7, line 214: site id not needed here?

AwkwardPeak7:
> For now webview will be loaded (for getting token) on first network call, if it fails it will not retry until app restart.
> 
> Api gives token with expire date so it's possible to check for that. I can implement saving token and its metadata in preferences. Native site js code is needed to update token if it's expires. Is it safe?

you could detect token expire and try to refresh it or throw exception. but you would have to implement it in kotlin, running arbitrary js code should only be reserved for extreme cases. 
But if the extensions work either way then it is fine no?

nedius:
Yes of course i'll do it in kotlin.
Native code needed for site itself to renew token when it expires.
I'll do it on weekends

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
nedius: Not needed in params because already is in headers. 
I think api server has 2 ways to know for what site call is: url params and header field.
I added `site_id` in headers and where site itself uses it as url param, added it too. To be safe and use same network logic as they're site.

nedius:
Is `isUserTokenValid()` a good idea ?
I added it just in case user logged out in webview or token was reset.
This endpoint fetches data about current user and if token invalid gives 401. This data is not used for anything else.

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -18,284 +22,247 @@ import eu.kanade.tachiyomi.source.model.Page
...
+        try {
+            val token = preferences.getString(TOKEN_STORE, "")!!.parseAs<AuthToken>()
+            if (token.isExpired() || !isUserTokenValid(token.getToken())) {
+                val refreshedToken: AuthToken? = refreshToken()
+                if (refreshedToken != null) {
+                    val str = json.encodeToString(refreshedToken)
+                    preferences.edit().putString(TOKEN_STORE, str).commit()
+                }
+            }
+            return token.getToken()
```
AwkwardPeak7, line 143: why return the old token after refreshing it?

AwkwardPeak7:
also seems like an extra slash is being added to page urls
image

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
nedius: Indeed forgot to override it

nedius:
2 slashes is default behavior on the site itself too 🤔
image
CDN just doesn't care about double or even more slashes, as i observed

AwkwardPeak7 (approved):
looks good for now, any quirks can be ironed out in later prs

PR #3245, opened 2024-05-26
Add Nijiero-ch extension 1.0

Moddimation:
This extension is NSFW related, and is less a manga reader, and more an image browser for NSFW content.
The original is "https://nijiero-ch.com/".
Latest is not supported, as i could not get a dedicated link that points to latest content. The website is a little dated anyways.

I tested it multiple times, and it should work. The search is only meant to search for tags or categories. If a genre/tag is specified, search is overriden. If it could not find either category/ or tag/, it searches for the other. I have not tested invalid search results yet, but everything else should work quite decently.

Please ask if needed, my discord is "Moddimation", aka "Moddigi Luigi".


src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,165 @@
...
+            else -> "tag"
+        }
+
+        val keyword = when {
+            query.isBlank() -> nijieroTags[tagIdx]
+            else -> query
+        }.replace(Regex("""\s+"""), "-").lowercase()
+
+        val uniqueParam = System.currentTimeMillis()
+        val url = "$baseUrl/$pageType/$keyword/page/$page?refresh=$uniqueParam"
```
vetleledaal: Need to escape `keyword`, for example: https://github.com/keiyoushi/extensions-source/blob/de5d088129b326b37b2c95031ed36db02f1a10e8/src/en/zahard/src/eu/kanade/tachiyomi/extension/en/zahard/Zahard.kt#L26-L30

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,165 @@
...
+
+        val pageType = when {
+            query.isBlank() -> "category"
+            else -> "tag"
+        }
+
+        val keyword = when {
+            query.isBlank() -> nijieroTags[tagIdx]
+            else -> query
+        }.replace(Regex("""\s+"""), "-").lowercase()
```
vetleledaal: Should move regex object outside so it's not recompiled every time.

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,165 @@
...
+
+    override fun mangaDetailsRequest(manga: SManga) = GET(manga.url, headers)
+    override fun mangaDetailsParse(response: Response): SManga {
+        val document = response.asJsoup()
+        return SManga.create().apply {
+            title = document.selectFirst("div.arrow.mb0 h1.type01_hl")?.text() ?: document.title()
+            thumbnail_url = document.select("meta[property=og:image]").attr("content")
+            status = SManga.COMPLETED
+            val genres = document.select("dl.cf:contains(カテゴリ) a").map { getCategoryOrTagFromLink(it.attr("href")) }
+            genre = genres.joinToString(", ")
```
vetleledaal: `, ` is the default separator and can be removed

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,165 @@
...
+import okhttp3.Request
+import okhttp3.Response
+import org.jsoup.select.Evaluator
+import java.text.SimpleDateFormat
+import java.util.Locale
+
+class Nijiero : HttpSource() {
+    override val name = "Nijiero"
+    override val lang = "all"
+    override val supportsLatest = false
```
vetleledaal: Looks like the latest section is on the home page?

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,165 @@
...
+        val uniqueParam = System.currentTimeMillis()
+        val url = "$baseUrl/ranking.html?refresh=$uniqueParam"
+
+        return GET(url, headers)
+    }
+    override fun popularMangaParse(response: Response): MangasPage {
+        val document = response.asJsoup()
+        val mangas = document.selectFirst("#mainContent .allRunkingArea.tabContent.cf")!!.children().mapNotNull { it ->
+            val link = it.select("a")
+            if (!link.isEmpty()) {
```
vetleledaal: If you use `it.selectFirst("a")` you can use more ideomatic Kotlin here,
```kt
link?.let { link ->
    SManga.create().apply { /* ... */ }
}
```

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,165 @@
...
+                    url = link.attr("href").removePrefix(baseUrl)
+                    title = link.attr("title")
+                    thumbnail_url = it.selectFirst("img")?.attr("data-src")
+                    status = SManga.COMPLETED
+                }
+            } else {
+                null
+            }
+        }
+        var isLastPage = document.selectFirst(Evaluator.Class("next page-numbers")) == null
```
vetleledaal: Consider using `ParsedHttpSource`, it includes convenience functions such as `searchMangaNextPageSelector()`, and you'd skip the `val document = response.asJsoup()` line from most functions

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,165 @@
...
+    }
+    private fun getCategoryOrTagFromLink(link: String): String {
+        val segments = link.trimEnd('/').split("/")
+        return segments.lastOrNull()?.replace("-", " ") ?: ""
+    }
+
+    override fun chapterListRequest(manga: SManga): Request = GET(manga.url, headers)
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val document = response.asJsoup()
+        val dateStr = document.selectFirst("div.postInfo.cf div.postDate.cf time.entry-date.date.published.updated")?.attr("datetime") ?: ""
```
vetleledaal: Can use `.orEmpty()`, here and elsewhere

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,165 @@
...
+        val dateUpload = parseDate(dateStr)
+	
+        return listOf(
+            SChapter.create().apply {
+                name = "Chapter"
+                setUrlWithoutDomain(response.request.url.encodedPath.removePrefix(baseUrl))
```
vetleledaal: `setUrlWithoutDomain()` should remove `baseUrl` for you

src/all/nijiero/res/mipmap-anydpi-v26/ic_launcher.xml
vetleledaal, line 1: Use the icon generator linked in [the contributing guide](https://github.com/keiyoushi/extensions-source/blob/main/CONTRIBUTING.md#tools)

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,165 @@
...
+    override val baseUrl = "https://www.nijiero-ch.com"
+
+    override val client = network.client.newBuilder().followRedirects(true).build()
+    class TagFilter : Filter.Select<String>("Category", nijieroTags)
+
+    override fun pageListParse(response: Response): List<Page> {
+        val document = response.asJsoup()
+        return document.select("#entry > ul > li a[href]").mapIndexed { i, linkElement ->
+            val linkUrl = linkElement.attr("href").removeSuffix(".webp").removePrefix(baseUrl)
+            Page(i, linkUrl, linkUrl)
```
vetleledaal: `Page`'s `url` should be the document where the image is found, not a link to the image itself.

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/NijieroTags.kt
```diff
@@ -0,0 +1,218 @@
+package eu.kanade.tachiyomi.extension.all.nijiero
+
+public val nijieroTags = arrayOf(
```
vetleledaal, line 3: Does this list update often? Consider updating it dynamically.

vetleledaal (changes requested):

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,165 @@
...
+            val currentUrl = response.request.url.toString()
+            val newUrl = if (currentUrl.contains("category/")) {
+                currentUrl.replace("category/", "tag/")
+            } else if (currentUrl.contains("tag/")) {
+                currentUrl.replace("tag/", "category/")
+            } else {
+                currentUrl
+            }
+            val newResponse = client.newCall(GET(newUrl, headers)).execute()
+            return searchMangaParse(newResponse)
```
vetleledaal: So if it doesn't find anything from either endpoint, it will recursively create new requests? You should have a guard against this, or change the implementation so that isn't possible.

Moddimation: is it better?

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/NijieroTags.kt
Moddimation: not really sure if they really added anything... should i make it fetch the tags in some way? but where and how?

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
Moddimation: wdym? it returns the image...

vetleledaal: 2nd argument is the `url`, 3rd argument is the `imageUrl`. If anything you should set the 2nd argument to `document.location()`.

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/NijieroTags.kt
vetleledaal: Seeing as the site hasn't been updated since 2023, I think it should be fine to leave as-is. If you wanted to implement it you could either [piggyback off existing responses](https://github.com/keiyoushi/extensions-source/blob/de5d088129b326b37b2c95031ed36db02f1a10e8/lib-multisrc/mccms/src/eu/kanade/tachiyomi/multisrc/mccms/MCCMSWeb.kt#L38), or [fetch them specifically](https://github.com/keiyoushi/extensions-source/blob/de5d088129b326b37b2c95031ed36db02f1a10e8/lib-multisrc/madara/src/eu/kanade/tachiyomi/multisrc/madara/Madara.kt#L1053)

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
vetleledaal: Other maintainers might have some additional suggestions in regard to the implementation details, but it seems fine - at least it's not recursively creating new requests.

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/NijieroTags.kt
Moddimation: the contents themselves are rather updated slowly, aswell. the last release is from a few months ago...

Moddimation:
I am done with my fixup, requesting review

Moddimation:
oh nvm im dumb

choppeh:
You can use the code below to get the tags. (I had some problems making a PR for your fork)

```kt
    class Tag(val name: String, val ref: String) {
        override fun toString() = name
    }
    class TagFilter(name: String, values: Array<Tag>) :
        Filter.Select<Tag>(name, values) {
        fun selectedValue() = values[state].ref
    }

        val keyword = when {
            query.isBlank() -> filters.filterIsInstance<TagFilter>().firstOrNull()?.selectedValue() ?: "various"
            else -> query
        }.replace(spaceRegex, "-").lowercase()

        val uniqueParam = System.currentTimeMillis()

        var url = "$baseUrl/category/$keyword/page/$page/".toHttpUrl().newBuilder()
            .addQueryParameter("refresh", "$uniqueParam")
            .build()

        // if 404, recheck
        if (client.newCall(GET(url, headers)).execute().code == 404) {
            url = "$baseUrl/tag/$keyword/page/$page".toHttpUrl().newBuilder()
                .addQueryParameter("refresh", uniqueParam.toString())
                .build()
            // if there is a better solution just *tell me* .
        }

        return GET(url, headers)
    }

    override fun getFilterList(): FilterList {
        CoroutineScope(Dispatchers.IO).launch { fetchTags() }

        val filters = mutableListOf<Filter<*>>(
            Filter.Header("You can search for a tag or category, anything else will result in 404(NOT FOUND)."),
            Filter.Header("To use category, make sure the search box is empty."),
        )

        filters += when {
            nijieroTags.isNotEmpty() -> listOf(TagFilter("Category", nijieroTags))
            else -> listOf(Filter.Separator(), Filter.Header("Press 'Reset' to attempt to show the tags"))
        }

        return FilterList(filters)
    }

    private var fetchTagsAttempts: Int = 0

    private fun fetchTags() {
        if (fetchTagsAttempts < 3 && nijieroTags.isEmpty()) {
            try {
                nijieroTags = client.newCall(tagsRequest()).execute().use { parseTags(it.asJsoup()) }
            } catch (_: Exception) {
            } finally {
                fetchTagsAttempts++
            }
        }
    }

    fun tagsRequest(): Request {
        return GET("$baseUrl/allsearch.html", headers)
    }

    private fun parseTags(document: Document): Array<Tag> {
        return document.select("dl.tagList02 dd a")
            .map { element ->
                Tag(
                    element.text().split(" ").first(),
                    element.attr("href").substringAfter("category/"),
                )
            }
            .toTypedArray()
    }
```

Run





### ***This is just a quick copy and paste from other implementations I've done; you need to treat the keyword properly, as I only considered the tag paths***

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,148 @@
...
+		
+        // if 404, recheck
+        if (client.newCall(GET(url, headers)).execute().code == 404) {
+            url = baseUrl.toHttpUrl().newBuilder()
+                .addEncodedPathSegment("tag")
+                .addEncodedPathSegment(keyword)
+                .addEncodedPathSegment("page")
+                .addEncodedPathSegment(page.toString())
+                .addQueryParameter("refresh", uniqueParam.toString())
+                .build()
```
choppeh, line 64: I couldn't find this path in the source

choppeh:
All the content (titles and tags) is in Japanese, I think it would be better to localize it to “jp” instead of “all”.

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,148 @@
...
+import org.jsoup.nodes.Element
+import java.text.SimpleDateFormat
+import java.util.Locale
+
+class Nijiero : ParsedHttpSource() {
+    override val name = "Nijiero"
+    override val lang = "all"
+    override val supportsLatest = true
+
+    override val baseUrl = "https://www.nijiero-ch.com"
```
choppeh: Remove ```www``` to avoid redirects

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,148 @@
...
+            // if there is a better solution just *tell me* .
+        }
+
+        return GET(url, headers)
+    }
+    override fun searchMangaNextPageSelector() = ".next.page-numbers"
+    override fun searchMangaSelector() = ".contentList > div:has(a)"
+    override fun searchMangaFromElement(element: Element) = SManga.create().apply {
+        element.select("a")?.let { link ->
+            thumbnail_url = element.selectFirst("img")?.attr("data-src") ?: element.selectFirst("img")?.attr("src")
```
choppeh: You can simply make a selection and choose the attribute that contains the thumbnail data. Example: https://github.com/keiyoushi/extensions-source/blob/de5d088129b326b37b2c95031ed36db02f1a10e8/src/zh/baozimhorg/src/eu/kanade/tachiyomi/extension/zh/baozimhorg/BaozimhOrg.kt#L206
https://github.com/keiyoushi/extensions-source/blob/de5d088129b326b37b2c95031ed36db02f1a10e8/src/zh/baozimhorg/src/eu/kanade/tachiyomi/extension/zh/baozimhorg/BaozimhOrg.kt#L70

Or

https://github.com/keiyoushi/extensions-source/blob/de5d088129b326b37b2c95031ed36db02f1a10e8/lib-multisrc/madara/src/eu/kanade/tachiyomi/multisrc/madara/Madara.kt#L763

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,148 @@
...
+        }.map {
+            it.replace("-", " ")
+        }
+        return genres.joinToString()
+    }
+
+    override fun chapterListSelector() = "html"
+    override fun chapterFromElement(element: Element) = SChapter.create().apply {
+        setUrlWithoutDomain(element.select("link[rel=canonical]").attr("abs:href"))
+        chapter_number = 0F
```
choppeh: Not need ```chapter_number```

choppeh:
You need to create a new branch from the ```m̀ain```. See: Submitting the changes

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,148 @@
...
+import java.util.Locale
+
+class Nijiero : ParsedHttpSource() {
+    override val name = "Nijiero"
+    override val lang = "all"
+    override val supportsLatest = true
+
+    override val baseUrl = "https://www.nijiero-ch.com"
+
+    override val client = network.client.newBuilder().followRedirects(true).build()
```
AwkwardPeak7:
```suggestion
    override val client = network.cloudflareClient
```
default client already follows redirect

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,148 @@
...
+	
+    class TagFilter : Filter.Select<String>("Category", nijieroTags)
+
+    override fun pageListParse(document: Document): List<Page> {
+        return document.select("#entry > ul > li a[href]").mapIndexed { i, linkElement ->
+            val linkUrl = linkElement.attr("href").removeSuffix(".webp").removePrefix(baseUrl)
```
AwkwardPeak7: why remove baseurl?

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,148 @@
...
+        var url = baseUrl.toHttpUrl().newBuilder()
+            .addEncodedPathSegment("category")
+            .addEncodedPathSegment(keyword)
+            .addEncodedPathSegment("page")
+            .addEncodedPathSegment(page.toString())
+            .addQueryParameter("refresh", uniqueParam.toString())
+            .build()
+		
+        // if 404, recheck
+        if (client.newCall(GET(url, headers)).execute().code == 404) {
```
AwkwardPeak7, line 57: it would be better to do this in an interceptor, so you can simply return the response as is if available or make new request and return it's response

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
```diff
@@ -0,0 +1,148 @@
...
+                .build()
+            // if there is a better solution just *tell me* .
+        }
+
+        return GET(url, headers)
+    }
+    override fun searchMangaNextPageSelector() = ".next.page-numbers"
+    override fun searchMangaSelector() = ".contentList > div:has(a)"
+    override fun searchMangaFromElement(element: Element) = SManga.create().apply {
+        element.select("a")?.let { link ->
```
AwkwardPeak7:
```suggestion
        element.select("a").let { link ->
```
`select` isn't nullable

AwkwardPeak7 (changes requested):

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
Moddimation: This code should actually check if the current page was not found as a tag, so it tries again with category... not sure what you mean by that, sorry.

Moddimation:
> All the content (titles and tags) is in Japanese, I think it would be better to localize it to “jp” instead of “all”.

@ttony2chopper I originally made the extension so that i can read it, so i made some stuff english. Is it required to move it all onto japanese in this case? I am not fluent in that language...

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
Moddimation: ...? You mean, like, file path? If you mean the website, try $baseurl/tag/gif, that should exist.

Moddimation:
The tag and category _links_ are, in reverse, almost all english/romanji, except some tags, thats why i just put english/romanji tags here. This is related to the former comment

Moddimation:
I mean category and tag

Moddimation:
To the branch: I commited to a branch called "pr", how do i set this PR to the branch? Do i need to create a new PR?

AwkwardPeak7:
> To the branch: I commited to a branch called "pr", how do i set this PR to the branch? Do i need to create a new PR?

you should push those commits in this branch, otherwise you would have to make a new pr

src/all/nijiero/src/eu/kanade/tachiyomi/extension/all/nijiero/Nijiero.kt
AwkwardPeak7: I mean instead of make a call just to check if it exists, add an [interceptor](https://square.github.io/okhttp/features/interceptors/) to detect it and make new call there.

basically return the tag request from here, in the interceptor prceed the call and check if it is 404 and url contains tag (this request) and if so, modify the request and call again, otherwise just return the original response. 

something like this:
```
private fun tagCategoryInterceptor(chain: Interceptor.Chain): Response {
        val request = chain.request()
        val url = request.url.toString()

        if (!response.isSuccessful && response.code == 404 && url.contains("/category/")) {
            val newUrl = url.replace("/category/","/tag/")
            val newRequest = request.newBuilder()
                .url(newUrl)
                .build()
            
            return chain.proceed(newRequest)
        }

        return response
    }
```

AwkwardPeak7 (changes requested):
remove extra changes in `common.gradle` and `settings.gradle.kts`

Moddimation:
I actually found several bugs, such as popular not showing images, some non manga entries included in recent sometimes, and tags with / dont work (%2F).

I dont think i have time to update this, though...

PR #3250, opened 2024-05-26
Changed Japscan, fixed the thumbnail problem 

Adam935:
Changed Japscan, fixed the image problem which loaded an obscene image (because of the pop on the screen) now we use a better selector to avoid this kind of problem

Checklist:

### DIFF
--- a/src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
+++ b/src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
@@ -88,11 +88,15 @@ class Japscan : ConfigurableSource, ParsedHttpSource() {
 
     override fun popularMangaFromElement(element: Element): SManga {
         val manga = SManga.create()
-        element.select("a").first()!!.let {
-            manga.setUrlWithoutDomain(it.attr("href"))
-            manga.title = it.text()
-            manga.thumbnail_url = "$baseUrl/imgs/${it.attr("href").replace(Regex("/$"),".jpg").replace("manga","mangas")}".lowercase(Locale.ROOT)
+        element.select("a").first()?.let { anchorElement ->
+            manga.setUrlWithoutDomain(anchorElement.attr("href"))
+            manga.title = anchorElement.select("strong.name").text()
+            anchorElement.select("img").first()?.let { imgElement ->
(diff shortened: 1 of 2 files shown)


src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -88,11 +88,15 @@ class Japscan : ConfigurableSource, ParsedHttpSource() {
 
     override fun popularMangaFromElement(element: Element): SManga {
         val manga = SManga.create()
-        element.select("a").first()!!.let {
-            manga.setUrlWithoutDomain(it.attr("href"))
-            manga.title = it.text()
-            manga.thumbnail_url = "$baseUrl/imgs/${it.attr("href").replace(Regex("/$"),".jpg").replace("manga","mangas")}".lowercase(Locale.ROOT)
+        element.select("a").first()?.let { anchorElement ->
```
AwkwardPeak7:
```suggestion
        element.selectFirst("a")!!.let { anchorElement ->
```

src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -88,11 +88,15 @@ class Japscan : ConfigurableSource, ParsedHttpSource() {
...
+            manga.setUrlWithoutDomain(anchorElement.attr("href"))
+            manga.title = anchorElement.select("strong.name").text()
+            anchorElement.select("img").first()?.let { imgElement ->
```
AwkwardPeak7:
```suggestion
            anchorElement.selectFirst("img")?.let { imgElement ->
```

src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -88,11 +88,15 @@ class Japscan : ConfigurableSource, ParsedHttpSource() {
...
+                val imageUrl = imgElement.attr("abs:data-src").ifEmpty { imgElement.attr("abs:src") }
```
AwkwardPeak7, line 95: use something like this:
```
private fun Element.imgAttr(): String = when {
    hasAttr("data-src") -> absUrl("data-src")
    else -> absUrl("src")
}
```

AwkwardPeak7 (changes requested):

PR #3291, opened 2024-05-29
Mangalink: update domain

### LINKED ISSUE #3261: MangaLink : url change report
### Source information

mangalink 1.4.38

### Source language

Arabic

### Source new URL

https://link-manga.com

### Other details

_No response_

### Acknowledgements

- [X] I have opened WebView and checked that the source URL has not changed yet.

vetleledaal:
Closes #3261

Also made the 'override URL' option not depend on the extension version.

Checklist:

### DIFF
--- a/src/ar/mangalink/src/eu/kanade/tachiyomi/extension/ar/mangalink/Mangalink.kt
+++ b/src/ar/mangalink/src/eu/kanade/tachiyomi/extension/ar/mangalink/Mangalink.kt
@@ -4,7 +4,6 @@ import android.app.Application
 import android.content.SharedPreferences
 import android.widget.Toast
 import androidx.preference.PreferenceScreen
-import eu.kanade.tachiyomi.extension.BuildConfig
 import eu.kanade.tachiyomi.multisrc.madara.Madara
 import eu.kanade.tachiyomi.source.ConfigurableSource
 import uy.kohesive.injekt.Injekt
@@ -15,7 +14,7 @@ import java.util.Locale
 class Mangalink :
     Madara(
         "مانجا لينك",
-        "https://manga-link.com",
+        "https://link-manga.com",
         "ar",
         SimpleDateFormat("MMMM dd, yyyy", Locale("ar")),
     ),
@@ -23,7 +22,7 @@ class Mangalink :
 
     override val chapterUrlSuffix = ""
     override val useLoadMoreRequest = LoadMoreStrategy.Always
-    private val defaultBaseUrl = "https://manga-link.com"
+    private val defaultBaseUrl = "https://link-manga.com"
     override val baseUrl by lazy { getPrefBaseUrl() }
 
     private val preferences: SharedPreferences by lazy {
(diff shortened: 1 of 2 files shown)


src/ar/mangalink/src/eu/kanade/tachiyomi/extension/ar/mangalink/Mangalink.kt
AwkwardPeak7, line 1: change the implementation to update preference when default baseUrl changes

vetleledaal: Not exactly sure how to do that without storing every old URL in the extension. Do you have any example implementation?

AwkwardPeak7: create "defaultBaseUrl" in preference, and on init, check if it is equal to current domain, if not change both "defaultbaseUrl" and the "overrideBaseUrl" to the new domain. this way it doesn't depend on version while also being able to prefer the domain from code when changed

vetleledaal: Still not exactly sure how that would work, wouldn't it just overwrite the user supplied URL on each init?
When calling `preferences.getString(BASE_URL_PREF, null)` it would be `null`, but if the preferences screen has been opened it would actually be set regardless if the user changed it.

So, assuming `BASE_URL_PREF`'s value is set to the old URL, and `DEFAULT_BASE_URL_PREF`'s value is set to the current URL, such as when the user have looked at the preferences screen once and an update was installed, updating the default URL:
```kt
preferences.getString(BASE_URL_PREF, null) // https://example1.com
preferences.getString(DEFAULT_BASE_URL_PREF, null) // https://example2.com
```

You'd see that they differ and update both of them to `https://example2.com`

But then, if the user explicitly updates it:
```kt
preferences.getString(BASE_URL_PREF, null) // https://example4.com
preferences.getString(DEFAULT_BASE_URL_PREF, null) // https://example2.com
```

You'd see that they differ and update both of them to `https://example2.com`?? I don't see how you would differentiate whether it's the old URL or set by the user, and if that's intended, I don't see the point of having an option of changing it if it'll just be changed back.


Maybe I'm just not seeing it, or not getting what you're hinting at

AwkwardPeak7: no you would check `DEFAULT_BASE_URL_PREF` against the hard coded one in the code, if they differ only then would user supplied url be changed. otheriwse use the user supplied url from `BASE_URL_PREF`.
and hardcoded url in code is only expected to change when someone updates the extension

vetleledaal: Ah, now I see. The user should be able to set the baseUrl at any time and it'll stay until an update comes through -- where it would be updated to the new baseUrl as defined in the extension.

AwkwardPeak7 (approved):

PR #3312, opened 2024-05-31
Add filters to Hitomi Extension

GoonGooner:
Added filters to the Hitomi Extension as it was little barebones. Should be able to use the filter menu for both positive and negative filters, filtering on artists, characters, Groups, Series, Female specific tags, male specific tags, and general tags. Based off a brief ~~goon sesh~~ test, everything seems to be working as expected.

Not a Kotlin cooder so apoligies if I failed to follow an expected style or pattern.

Checklist:

- ~~[ ] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions~~
- ~~[ ] Referenced all related issues in the PR body (e.g. "Closes #xyz")~~
- ~~[ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate~~
- ~~[ ] Have explicitly kept the `id` if a source's name or language were changed~~
- ~~[ ] Have removed `web_hi_res_512.png` when adding a new extension~~

### DIFF
--- a/src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/Hitomi.kt
+++ b/src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/Hitomi.kt
@@ -2,7 +2,6 @@ package eu.kanade.tachiyomi.extension.all.hitomi
 
 import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.network.await
-import eu.kanade.tachiyomi.source.model.Filter
 import eu.kanade.tachiyomi.source.model.FilterList
 import eu.kanade.tachiyomi.source.model.MangasPage
 import eu.kanade.tachiyomi.source.model.Page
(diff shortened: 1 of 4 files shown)


src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/Hitomi.kt
```diff
@@ -93,12 +137,53 @@ class Hitomi(
         }
     }
 
-    private class SortFilter : Filter.Select<String>("Sort By", arrayOf("Updated", "Popularity"))
+    private class OrderFilter(val name: String, val order: OrderType) {
```
AwkwardPeak7: since hitomi class is already too big, move the filter classes to a seperate file.

src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/Hitomi.kt
```diff
@@ -126,9 +211,6 @@ class Hitomi(
                 .replace(Regex("""^\?"""), "")
                 .lowercase()
                 .split(Regex("\\s+"))
-                .map {
-                    it.replace('_', ' ')
```
AwkwardPeak7, line 130: why remove 🤔

src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/Hitomi.kt
```diff
@@ -75,12 +75,56 @@ class Hitomi(
...
     private lateinit var searchResponse: List<Int>
 
+    private fun filterToTag(query: StringBuilder, tag: String, filterState: String) {
+        filterState
+            .trim()
+            .split(',')
+            .filter { it.isNotBlank() }
+            .forEach {
+                val trimmed = it.trim()
+                query.append(" ${if (trimmed.startsWith("-")) "-" else ""}$tag:${trimmed.removePrefix("-").replace(" ", "_")}")
```
AwkwardPeak7: consider moving the if condition out for cleaner code

AwkwardPeak7 (changes requested):

src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/Hitomi.kt
GoonGooner: Moved to another file(`Filters.kt`), also moved filter parsing there as well as it felt relevant. Condensed text filters into class as all shared same usage.

src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/Hitomi.kt
GoonGooner: `getGalleryIDsForQuery` function also calls `.replace`, but on the individual query terms, no need to do it twice. Could do it the other way around, keeping the `.map` and not doing it in `getGalleryIDsForQuery`.

Hitomi themselves do it twice (`results.js` L29 and `search.js` L605), which I guess is why we do it twice, but I don't see the reason why.

src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/Hitomi.kt
GoonGooner: Just moved the if to store in a val instead of doing it in string template. Function moved to `Filters.kt`.

GoonGooner:
Would there be any interest in a language filter option? Idea being able to search in other languages without having to switch to `all` source and manually typing out `language:<language>` or swapping to the specific language's source. Maybe get rid of factory and add preference screen where user can set default language(defaults to `all` on first install). Probably best to save this for future PR.

I know currently language is gotten from `nozomiLang` or parsed from query input. I would be thinking of a select filter set to default with whatever `nozomiLang` is. I don't think support for multi language input as I couldn't find any on Hitomi, but maybe I'm not searching with the right combination of languages. And if someone needs multi-lang they can still use searchbar and manually make the query.

If a language filter option is wanted and we go with the select filter, is there a way to have it so languages used in factory are shared with languages used by filter. Create array of pairs `<SourceLang: String, nozomiLang: String>`, and use first and second in factory and only use second for language filter? Where to store array as well, in Dto file or hitomi.kt file itself or factory, new file? If we get rid of factory we don't really need to worry about this.

AwkwardPeak7:
don't add language filter to language specific source instance. you can add them to only the multi one (`lang = "all"`) if you want

GoonGooner:
Okay, not going to implement language filter then.

AwkwardPeak7 (approved):

PR #3384, opened 2024-06-03
Hitomi: Fix language filter in searching

ZIDOUZI:
As my test, Hitomi extension can filter by language in popular and recent, but failed in search. But that can be solved by adding `langauge:specific` to query to solve. That what I do.

Checklist:

### DIFF
--- a/src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/Hitomi.kt
+++ b/src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/Hitomi.kt
@@ -229,7 +229,10 @@ class Hitomi(
         query: String,
         language: String = "all",
     ): MutableList<Int> {
-        query.replace("_", " ").let {
+        query
+            .let { if (language != "all") query + " language:$nozomiLang" else query }
+            .replace("_", " ")
+            .let {
             if (it.indexOf(':') > -1) {
                 val sides = it.split(":")
                 val ns = sides[0]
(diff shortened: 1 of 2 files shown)


src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/Hitomi.kt
```diff
@@ -229,7 +229,7 @@ class Hitomi(
         query: String,
         language: String = "all",
     ): MutableList<Int> {
-        query.replace("_", " ").let {
+        query.let { if (language != "all") query + " language:$nozomiLang" else query }.replace("_", " ").let {
```
AwkwardPeak7: it won't work here, you need to add this to `hitomiSearch`. add `language:$nozomiLang` to the `terms` list there

ZIDOUZI: well, they will finally go to `getGalleryIDsForQuery` from the code of `hitomiSearch`, but move it to `hitomiSearch` deed more logical.

AwkwardPeak7:
ended up doing it as part of #3389, since there were some things that I wanted to address, sorry for eating up your pr

src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/Hitomi.kt
AwkwardPeak7: well `getGalleryIDsForQuery` is used for individual query terms, so it can't handle two terms together

ZIDOUZI:
It's ok, the best thing is to solve the problem

src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/Hitomi.kt
ZIDOUZI: Right. My fault.

PR #3435, opened 2024-06-05
add a Dark Science extension

Evrey:
Hello there!

This one adds a new extension for the webcomic [Dark Science](https://dresdencodak.com/category/darkscience/). The Dresden Codak website has a few more comics, but i’m only reading Dark Science there. So if demand arises, this could instead turn into a Dresden Codak extension in the future.

Checklist:

  - Not a multisrc.
  - No open issues.
  - SFW
  - Ran the extension in private for a long time, so it’s battle-tested.


src/en/darkscience/build.gradle
```diff
@@ -0,0 +1,7 @@
+ext {
+    extName = 'Dark Science'
+    extClass = '.DarkScience'
+    extVersionCode = 2
```
AwkwardPeak7:
```suggestion
    extVersionCode = 0
```

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
```diff
@@ -0,0 +1,156 @@
...
+        "her money’s gone and a bank exploded her house. " +
+        "With no place else to go, she travels to " +
+        "Nephilopolis, the city of giants – built from the " +
+        "ruins of an ancient war and a fading memory of " +
+        "tomorrow.\n" +
+        "Follow our cyborg hero as she attempts to survive " +
+        "the bureaucratic behemoth with a little “help” " +
+        "from her “friends.” And what exactly is " +
+        "Dark Science anyway?\n" +
+        "Support the comic on Patreon: $patreonUrl"
```
AwkwardPeak7: you can inline most of these

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
```diff
@@ -0,0 +1,156 @@
...
+    private fun initTheManga(manga: SManga): SManga = manga.apply {
+        url = archiveUrl
+        thumbnail_url = thumbnailUrl
+        title = seriesTitle
+        author = authorName
+        artist = authorName
+        description = seriesDescription
+        genre = seriesGenre
+        status = SManga.ONGOING
+        update_strategy = UpdateStrategy.ALWAYS_UPDATE
```
AwkwardPeak7:
```suggestion
```
it is already default

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
AwkwardPeak7, line 1: if the site have other comics than you should try to support them too, seems like it won't be too dificult, get the available comics and their description and what not from`https://dresdencodak.com/archives/`

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
```diff
@@ -0,0 +1,156 @@
...
+        // so far, that whole endeavour seems not worth the effort.
+        // So here it is: Each »chapter« having just 1 page.
+        return Observable.just(listOf(Page(0, chapter.url)))
+    }
+
+    override fun fetchImageUrl(page: Page): Observable<String> {
+        val comicPage = client.newCall(GET(baseUrl + page.url, headers)).execute().asJsoup()
+        val imageUrl = comicPage.selectFirst("article.post img.aligncenter")!!.attr("src")
+        return Observable.just(imageUrl)
+    }
```
AwkwardPeak7: override imageUrlParse instead

AwkwardPeak7 (changes requested):

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
Evrey: True. I took inspiration from another extension there, which also kept this metadata out of the way from the »do stuff« code, and liked that. I’ll inline if you prefer that, though.

src/en/darkscience/build.gradle
Evrey: Hey there, question regarding that:

What *is* the state of versions here anyway? The docs seemed to imply that that’s the micro version of the extension API used, which would make it `2` because the linked extension API is `1.4.2`. But you setting that to zero now makes me think this is a mere monotonically incrementing counter for this single extension’s version? Which is it?

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
Evrey: I’ll remove that. It’s a habit from systems programming, where you distrust defaults and prefer making intents explicit. Just so you know.

src/en/darkscience/build.gradle
AwkwardPeak7: Yes it is a incrementing version number of the extension, the app prompts for update when this number increases. The major part (1.4) is the major ext-lib version.

And sorry it should be 1 instead of 0. I mistook it for another one when I took a quick look

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
Evrey: I could do that, but i’d prefer if someone else did it as a successor extension. Maybe i’ll get around to it, but no promises.

src/en/darkscience/build.gradle
Evrey: Setting it to zero makes APK compilation fail. I set it to 1 now.

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
Evrey: Didn’t notice that function, for the extension i learned from didn’t use it, either. Thanks, changed.

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
Evrey: Removed in latest commit.

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
Evrey: All properties that were used only once have been moved now, with the exception of the long description.

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
```diff
@@ -0,0 +1,145 @@
...
+        "ruins of an ancient war and a fading memory of " +
+        "tomorrow.\n" +
+        "Follow our cyborg hero as she attempts to survive " +
+        "the bureaucratic behemoth with a little “help” " +
+        "from her “friends.” And what exactly is " +
+        "Dark Science anyway?\n" +
+        "Support the comic on Patreon: https://www.patreon.com/dresdencodak"
+
+    private fun initTheManga(manga: SManga): SManga = manga.apply {
+        url = archiveUrl
```
AwkwardPeak7:
```suggestion
        setUrlWithoutDomain(archiveUrl)
```

Evrey: Fixed in latest commit.

Evrey:
Heyo. Some changes happened after i finally learned a bit of how these Observable things work.

- `fetchChapterList` should no longer be blocking.
- Switched to `pageListParse`, like previously with `imageUrlParse`.

Unfortunately the helper function `accumulateChapters` has to rely on recursion. Can’t see how to pull an Observable from a `suspend fun` at the given minimum API level.

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
```diff
@@ -0,0 +1,155 @@
...
+
+        return SChapter.create().apply {
+            name = chTitle
+            chapter_number = chNum
+            date_upload = chDate
+            setUrlWithoutDomain(chLink)
+        }
+    }
+
+    private fun accumulateChapters(state: Triple<Call?, Float, Observable<SChapter>>): Observable<Triple<Call?, Float, Observable<SChapter>>> {
```
AwkwardPeak7: you don't need to use observables, since the fetch methods are already run in a background thread. 

\+ observables are deprecated and will be removed in the next version of source api, so I would like to avoid new usage of them

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
```diff
@@ -0,0 +1,155 @@
...
+
+    override fun fetchChapterList(manga: SManga): Observable<List<SChapter>> =
+        accumulateChapters(Triple(client.newCall(chapterListRequest(manga)), 0.0F, Observable.empty())).concatMap { it.third.toList() }
+
+    override fun pageListParse(response: Response): List<Page> =
+        listOf(
+            Page(
+                0,
+                "",
+                response
```
AwkwardPeak7: `imageUrl = `

src/en/darkscience/src/eu/kanade/tachiyomi/extension/en/darkscience/DarkScience.kt
Evrey: Oh interesting! And glad to hear observables’ll be nuked. They’re really awkward to use. I’ll revert this, then.

AwkwardPeak7 (approved):

PR #3474, opened 2024-06-08
[RU]LibGroup filtered excess request if branchesCount is only alone

e-shl:
+the same apiDomain in the urls

Checklist:

### DIFF
--- a/lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
+++ b/lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
@@ -311,7 +311,7 @@ abstract class LibGroup(
         }
 
         val sortingList = preferences.getString(SORTING_PREF, "ms_mixing")
-        val defaultBranchId = if (chaptersData.data.getBranchCount()!! > 1) { // excess request if branchesCount is only alone = slow update library witch rateLimitHost(apiDomain.toHttpUrl(), 1)
+        val defaultBranchId = if (chaptersData.data.getBranchCount() > 1) { // excess request if branchesCount is only alone = slow update library witch rateLimitHost(apiDomain.toHttpUrl(), 1)
             runCatching { getDefaultBranch(slugUrl.substringBefore("-")).first().id }.getOrNull()
         } else {
             null
--- a/lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
+++ b/lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
@@ -64,9 +64,19 @@ abstract class LibGroup(
 
     override val supportsLatest = true
 
+    private val userAgentMobile = "Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Mobile Safari/537.3"
+
+    private var bearerToken: String? = null
+
+    private var userId: Int? = null
+
+    abstract val siteId: Int // Important in api calls
+
+    private val apiDomain: String = "https://api.lib.social"
+
     override val client: OkHttpClient = network.cloudflareClient.newBuilder()
         .rateLimit(3)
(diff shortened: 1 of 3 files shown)


lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -311,8 +311,10 @@ abstract class LibGroup(
         }
 
         val sortingList = preferences.getString(SORTING_PREF, "ms_mixing")
-        val defaultBranchId = runCatching { getDefaultBranch(slugUrl.substringBefore("-")).first().id }.getOrNull()
-
+        var defaultBranchId: Int? = null
+        if (chaptersData.data.filter { it.branchesCount > 1 }.isNotEmpty()) { // excess request if branchesCount is only alone = slow update library witch rateLimitHost(apiDomain.toHttpUrl(), 1)
```
nedius: This may break in future ? LibGroup sites itself always making calls to get default branch even if chapters have only one branch.

It would be better to get branch list length, than trust `branchesCount`. Also getting branches count should be implemented in `Chapter` class in `LibGroupDto.kt` file.

No need `defaultBranchId` to be `var`, make it `val` and move if to definition
`val defaultBranchId: Int? = if(...) {...} else null`

Don't use filter use any instead.

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
```diff
@@ -311,8 +311,10 @@ abstract class LibGroup(
...
+            defaultBranchId = runCatching { getDefaultBranch(slugUrl.substringBefore("-")).first().id }.getOrNull()
+        }
         val chapters = mutableListOf<SChapter>()
```
nedius, line 320:
```suggestion
        val defaultBranchId = if (chaptersData.data.getBranchCount()!! > 1) {
            runCatching { getDefaultBranch(slugUrl.substringBefore("-")).first().id }.getOrNull()
        } else {
            null
        }
            
```

nedius: And add this after `class Chapter` in `LibGroupDto.kt`: 
```kt
fun List<Chapter>.getBranchCount(): Int? = this.maxOfOrNull { chapter -> chapter.branches.size }
```

nedius: Better solution in my opinion

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroup.kt
e-shl: > This may break in future ? LibGroup sites itself always making calls to get default branch even if chapters have only one branch.

In the future, we'll see, in this case it's just an extra request for emptiness and on the site, in fact, too

e-shl: but mangas with several branches is few

lib-multisrc/libgroup/src/eu/kanade/tachiyomi/multisrc/libgroup/LibGroupDto.kt
```diff
@@ -245,6 +245,8 @@ class Chapter(
     }
 }
 
+fun List<Chapter>.getBranchCount(): Int? = this.maxOfOrNull { chapter -> chapter.branches.size }
```
AwkwardPeak7: you are using non null assertion where you use this function. so better to just use `maxOf` instead

nedius: This function will not execute if chapter list empty.
There is check for that in `chapterListParse`.

So yes, no need for null assertion.

nedius (commented):
Looks good to me 👍

AwkwardPeak7 (approved):

PR #3689, opened 2024-06-21
LuraToon: Add login warning

choppeh:
Checklist:

### DIFF
--- a/src/pt/randomscan/build.gradle
+++ b/src/pt/randomscan/build.gradle
@@ -3,7 +3,7 @@ ext {
     extClass = '.LuraToon'
     themePkg = 'peachscan'
     baseUrl = 'https://luratoons.com'
-    overrideVersionCode = 44
+    overrideVersionCode = 45
 }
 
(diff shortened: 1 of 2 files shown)


bapeey:
Does the latest tab work? You can simply check the response.request.url in pageListParse, no interceptor required

choppeh:
> Does the latest tab work? You can simply check the response.request.url in pageListParse, no interceptor required

I disabled the latest tab, I think I had seen the old cache

bapeey:
I asked because the interceptor will be thrown in the latest tab with the old implementation, but should still work. Also keep the `pathSegments` empty check in the `pageListParse`

choppeh:
> I asked because the interceptor will be thrown in the latest tab with the old implementation, but should still work. Also keep the `pathSegments` empty check in the `pageListParse`

You're right, I checked PeachScan, and the latestUpdate URL is in the 'root' of the domain

My bad :donkey: :smile:

AwkwardPeak7 (approved):

PR #3884, opened 2024-07-05
Update debugger instructions

ringosham:
A small change to the Android debugger section. The method mentioned only works on:
* Rooted devices
* Emulated devices without Google play (Google play enabled devices do not have su)

Reference:
https://developer.android.com/studio/debug#attach-debugger

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -720,6 +720,9 @@ And for a release build of Tachiyomi:
 
 ### Android Debugger
 
+> [!IMPORTANT]
+> Rooted device required. If you are using an emulator instead, make sure you choose a profile without Google Play.
+
 You can leverage the Android Debugger to step through your extension while debugging.
 
 You *cannot* simply use Android Studio's `Debug 'module.name'` -> this will most likely result in an
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -721,7 +721,7 @@ And for a release build of Tachiyomi:
 ### Android Debugger
 
 > [!IMPORTANT]
-> Rooted device required. If you are using an emulator instead, make sure you choose a profile without Google Play.
+> Rooted device required. If you are using an emulator instead, make sure you choose a profile **without** Google Play.
 
 You can leverage the Android Debugger to step through your extension while debugging.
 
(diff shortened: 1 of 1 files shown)


vetleledaal:
What version of Android Studio are you using? I previously had issues with debugging on Iguana/Jellyfish, but it started working again after switching to Koala. Works for both non-rooted physical devices and non-rooted emulated devices with and without GMS.

(Downgrading to Hedgehog also works, but I don't think that version supports the Gradle version currently used in this project)

ringosham:
I'm using the latest release of Android studio Koala with SDK platform 35.

I initially had this problem when I was writing the nicovideo extension (which got pushed a few hours ago). I got a new computer so all I have is default configs, ended up figuring out the issue is because the device is not rooted.

I'm guessing if you build the main app from source you can debug it just fine. But since I'm downloading the main app from their website, the process is not debuggable without root.

The main reason root is needed is because the main app needs to be a debug build (with ~~debug signature~~ `debuggable` enabled in the build, not pre-release). Otherwise, you'll need root to let adb `ps` all processes

ringosham:
It's also quite confusing as "debug" build can be easily mixed up with the pre-release build instead of a ~~debug signature~~ debuggable build. The package name ends with `.debug` afterall, but it doesn't seem to have the debuggable flag enabled either.

CONTRIBUTING.md
```diff
@@ -720,6 +720,10 @@ And for a release build of Tachiyomi:
 
 ### Android Debugger
 
+> [!IMPORTANT]
+> If you didn't build Tachiyomi from source with debug enabled and are using a release/beta APK, you **need** a rooted device.
```
AwkwardPeak7:
```suggestion
> If you didn't build Mihon from source with debug enabled and are using a release/beta APK, you **need** a rooted device.
```

ringosham: Yeah I am conflicted about this. The entire document references "Tachiyomi" rather than "Mihon" so I was avoiding that to keep things uniform.

AwkwardPeak7: yeah it is kinda outdated in some places too... you can refer to "the app" instead of name droping

ringosham: I'm just gonna call it "the main app"

AwkwardPeak7 (approved):

PR #3909, opened 2024-07-07
Add Asura Scans TR

vetleledaal:
Checklist:


AwkwardPeak7:
is it same or different from the (original) Asura Scan TR which rebranded to Armoni Scan...?

vetleledaal:
Different Cloudflare NS, different registrar, different set of plugins, probably not run by the same people. They do link to Armoni Scan on the front page, but that seems more like a citation than anything else. 

Should I rename it to something else?

AwkwardPeak7 (approved):

PR #3973, opened 2024-07-11
Add Pijamalı Koi

vetleledaal:
Probably overcomplicating the `fixupUrl()` function a bit.

Checklist:


src/tr/pijamalikoi/src/eu/kanade/tachiyomi/extension/tr/pijamalikoi/PijamaliKoi.kt
```diff
@@ -0,0 +1,35 @@
...
+        return super.mangaDetailsRequest(manga).fixupUrl()
+    }
+
+    override fun pageListRequest(chapter: SChapter): Request {
+        return super.pageListRequest(chapter).fixupUrl()
+    }
+
+    // Fixes the `/m` in baseUrl
+    private fun Request.fixupUrl(): Request {
+        return takeIf { url.pathSize >= 1 }
```
AwkwardPeak7: when does it remove the /m/ ? The site seems to just have it for manga and chapter urls

vetleledaal: If everything is working as expected, it will remove it every time. There's something in `*Request()` that makes the URL look like `https://example.com/m/m/seri/regina-rena-to-the-unforgiven/`.

On second thought, I guess it would fail only if the URL is something like `https://example.com`, but at that point you have bigger problems.

I'll change it so it removes the extra segment unconditionally.

AwkwardPeak7 (approved):

PR #4042, opened 2024-07-16
LibGroup: fix crash on null token

AwkwardPeak7:
can't really test without account...

Checklist:


bapeey:
From crashlog

```
kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 9: Expected start of the object '{', but had 'n' instead at path: $.token
JSON input: {"token":null,"prevUrl":"","timestamp":1718454747252}
```

`token` should be nullable too, and add default `null` value to `auth`

AwkwardPeak7:
superseded by #4071

PR #4044, opened 2024-07-16
Koharu: Add More Filters, fix popular

KenjieDec:
Checklist:

### DIFF
--- a/src/en/koharu/src/eu/kanade/tachiyomi/extension/en/koharu/Koharu.kt
+++ b/src/en/koharu/src/eu/kanade/tachiyomi/extension/en/koharu/Koharu.kt
@@ -48,7 +48,7 @@ class Koharu : HttpSource(), ConfigurableSource {
 
     private val json: Json by injectLazy()
 
-    private val B_REGEX = Regex("""\s*\[.*?]\s*|\s*\(.*?\)\s*|\s*\w+[\)\]]""")
+    private val B_REGEX = Regex("""\s*[([{].*[)\]}]\s*""")
 
     private val preferences: SharedPreferences by lazy {
         Injekt.get<Application>().getSharedPreferences("source_$id", 0x0000)
--- a/src/en/koharu/src/eu/kanade/tachiyomi/extension/en/koharu/Koharu.kt
+++ b/src/en/koharu/src/eu/kanade/tachiyomi/extension/en/koharu/Koharu.kt
@@ -4,6 +4,7 @@ import android.app.Application
 import android.content.SharedPreferences
 import androidx.preference.ListPreference
 import androidx.preference.PreferenceScreen
+import androidx.preference.SwitchPreferenceCompat
 import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.network.POST
 import eu.kanade.tachiyomi.network.interceptor.rateLimit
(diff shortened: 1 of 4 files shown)


src/en/koharu/src/eu/kanade/tachiyomi/extension/en/koharu/KoharuDto.kt
```diff
@@ -60,8 +60,16 @@ class Data(
 @Serializable
 class DataKey(
     val id: Int,
+    val size: Double = 0.0,
     val public_key: String,
-)
+) {
+    fun readableSize() = when {
```
vetleledaal, line 66: Use GiB, MiB, KiB


or if you want to keep using GB, MB, kB you should calculate that (powers-of-10/decimal/SI prefix). Wouldn't recommend it, as those units are ambiguous.

src/en/koharu/src/eu/kanade/tachiyomi/extension/en/koharu/Koharu.kt
```diff
@@ -47,19 +48,23 @@ class Koharu : HttpSource(), ConfigurableSource {
 
     private val json: Json by injectLazy()
 
+    private val B_REGEX = Regex("\\s*\\[.*?]\\s*|\\s*\\(.*?\\)\\s*")
```
vetleledaal: IIRC, you would run into less issues with escaping if you use a triple quoted string

vetleledaal: Should also be more greedy, otherwise it will fail on `123 (456 (789) abc)`. Or are you specifically doing that to avoid another issue?

vetleledaal (changes requested):

src/en/koharu/src/eu/kanade/tachiyomi/extension/en/koharu/KoharuDto.kt
KenjieDec: Hello, thank you for pointing that out. The website seems to convert Bytes to MiB but writes it as MB. 
I'll still use the MB and GB format as I think it's more commonly understood. Thank you

src/en/koharu/src/eu/kanade/tachiyomi/extension/en/koharu/Koharu.kt
vetleledaal: I mean- you should remove the lazy token (`?`). Currently it fails on `123 (456 (789) ab c)`.

I would go for something like:
```
\s*[([{].*[)\]}]\s*
```

src/en/koharu/src/eu/kanade/tachiyomi/extension/en/koharu/KoharuDto.kt
```diff
@@ -66,7 +66,7 @@ class DataKey(
     fun readableSize() = when {
         size >= 300 * 1024 * 1024 -> "${"%.2f".format(size / (1024.0 * 1024.0 * 1024.0))} GB"
         size >= 100 * 1024 -> "${"%.2f".format(size / (1024.0 * 1024.0))} MB"
-        size >= 1024 -> "${"%.2f".format(size / (1024.0))} KB"
+        size >= 1024 -> "${"%.2f".format(size / (1024.0))} kB"
```
vetleledaal: Using `kB` here is wrong. Should be `KB` according to https://en.wikipedia.org/wiki/Kilobyte#Definitions_and_usage

>  [...] one kilobyte is 1000 bytes. The unit symbol is kB.
>
> The binary meaning of the kilobyte for 1024 bytes typically uses the symbol KB, with an uppercase letter _K_. 

You should be following the latter standard, since you're working with powers-of-2/binary/IEC prefixes

(strictly it should be KiB, but you already rejected that ¯\\_(ツ)\_\/¯)

KenjieDec: Sorry if it seemed like I rejected your idea, that was not my intention. I just forgot to upload my dto file when commiting [fab3393](https://github.com/keiyoushi/extensions-source/pull/4044/commits/fab33932038ec5f92c9230da8a553264cc9fcda5) 😅.

AwkwardPeak7 (approved):

PR #4157, opened 2024-07-22
add a Darths & Droids extension

Evrey:
Checklist:

  - Starting at 1
  - Not multisrc
  - No issues yet
  - ’tis SFW
  - No prior extension yet
  - Also tested the extension for months

Hi, it’s me again!

I’m back to having FOSS time. Thanks for finishing and merging Dark Science, btw.!

Anyway, here’s an extension for the campaign comic series »Darths & Droids«. I have finished reading the first two books with it, and i have sampled a few pages of the other books with success. For all i know the extension works fine and *should* continue working once the authors begin working on Star Wars 9.

The `dndThumbnailUrlForTitle` function is the only piece that requires maintenance. It’ll pick a default thumbnail for not yet known books, though, so the extension won’t break. Further more the extension won’t break assuming the authors continue following the largely same-y archive structure they have so far.

The extension could be made faster and less stressing for the comic website, if the extension API received some future modifications. For example, all their finished comic books can be downloaded as a single Zip archive of their pages. (Not CBZ, though, it’s lacking the XML metadata.)


src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+        // is ignored for almost all archives anyway. Only those not pointing at a `.html`
+        // page matter.
+        return Observable.just(fetchChaptersForDatedAndDatelessArchives(archiveResponse.asJsoup(), manga.url))
+    }
+
+    override fun fetchPageList(chapter: SChapter): Observable<List<Page>> =
+        Observable.just(listOf(Page(0, chapter.url)))
+
+    override fun fetchImageUrl(page: Page): Observable<String> {
+        val comicPage = client.newCall(GET(baseUrl + page.url, headers)).execute().asJsoup()
```
vetleledaal: In general you should avoid using the `fetch*` methods unless necessary. The HttpSource (or ParsedHttpSource) class has several functions you can override, assuming the site follows largely the same format as other extensions.

~~When using `client.newCall(...).execute()` you should use `.use()` so the stream is closed. You mostly don't need to worry about this if you implement the suggestion in the first paragraph.~~ Not a thing in OkHttp 5 when using `.string()` or `.bytes()`, ignore that.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+        // Careful. For almost all images it’s `div.center>p>img`, except for pages released on
+        // April’s Fools day, when it’s `div.center>p>a>img`.
+        val imageUrl = comicPage.selectFirst("div.center img")!!.attr("src")
```
vetleledaal: If you know the only valid paths are as described in the comment, why not `div.center > p img`

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+                    )
+                    i += 1
+                }
+            }
+        }
+        chapters.reverse()
+        return chapters
+    }
+
+    private fun absolutePageUrl(pageUrl: String, maybeRelativeTo: String): String =
```
vetleledaal: Normally you would just store the relative URL, that way not every entry becomes invalid if the domain changes.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+                    // the Solo book was published, because currently Solo’s the only one such book.
+                    val pageDate = 1660435200000L + (i * 1000 * 60)
+                    val a = pageAnchor
+
+                    chapters.add(
+                        SChapter.create().apply {
+                            name = a.text()
+                            chapter_number = i.toFloat()
+                            date_upload = pageDate
+                            url = absolutePageUrl(a.attr("href"), archivePath)
```
vetleledaal: `setUrlWithoutDomain()` is preferred
```suggestion
                            setUrlWithoutDomain(a.absUrl("href"))
```

(I have not fully unraveled the significance of `archivePath`)

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+                    },
+                )
+                i += 1
+            } else if (!pageData.hasAttr("colspan")) {
+                // Are we in a dateless archive?
+                pageAnchor = pageData.getOrNull(0)?.selectFirst("a")
+                if (pageAnchor != null) {
+                    // For now we assume all dateless archives were published the same day
+                    // the Solo book was published, because currently Solo’s the only one such book.
+                    val pageDate = 1660435200000L + (i * 1000 * 60)
```
vetleledaal: What is the date? And is it better than the default behavior (date set to current time)?

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+            var pageAnchor = pageData.getOrNull(2)?.selectFirst("a")
+            // null for »Intermission«, main archive, dateless archive,…
+            if (pageAnchor != null) {
+                val a = pageAnchor
+
+                chapters.add(
+                    SChapter.create().apply {
+                        name = a.text()
+                        chapter_number = i.toFloat()
+                        date_upload = DATE_FMT.parse(pageData[0].text().trim())?.time ?: 0L
```
vetleledaal: Wrap in `runCatching`

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+import okhttp3.Request
+import okhttp3.Response
+import org.jsoup.nodes.Document
+import rx.Observable
+import java.text.SimpleDateFormat
+import java.util.Locale
+
+// Dear Darths & Droids creators:
+// I’m sorry if this extension causes too much traffic for your site.
+// If possible, i would use the Zip downloads to source finished
```
vetleledaal: Reading directly from ZIP has been done before in PandaChaika, but that was out of necessity. I would recommend against implementing that.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+import eu.kanade.tachiyomi.util.asJsoup
+import okhttp3.Request
+import okhttp3.Response
+import org.jsoup.nodes.Document
+import rx.Observable
+import java.text.SimpleDateFormat
+import java.util.Locale
+
+// Dear Darths & Droids creators:
+// I’m sorry if this extension causes too much traffic for your site.
```
vetleledaal, line 22: Add a ratelimit to `client` instead, if the site owner complains it can be tweaked.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+// If possible, i would use the Zip downloads to source finished
+// chapters. However, as far as i can see, there is no way to hack
+// this into the existing extension API.
+class DarthsDroids : HttpSource() {
+    override val name = "Darths & Droids"
+    override val baseUrl = "https://www.darthsanddroids.net/"
```
vetleledaal: Remove trailing slash

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+        10 -> "https://www.darthsanddroids.net/cast/Han4.jpg" // Solo
+        11 -> "https://www.darthsanddroids.net/cast/Hux8.jpg" // D&D8
+        // Just some nonsense fallback that screams »Star Wars« but is also so recognisably
+        // OT that one can understand it’s a mere fallback. Better thumbnails require an
+        // extension update.
+        else -> "https://www.darthsanddroids.net/cast/Vader4.jpg"
+    }
+
+    private fun dndManga(archiveUrl: String, mangaTitle: String, mangaStatus: Int, nthManga: Int): SManga = SManga.create().apply {
+        url = archiveUrl
```
vetleledaal: use `setUrlWithoutDomain()`

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+            else -> UpdateStrategy.ALWAYS_UPDATE
+        }
+        initialized = true
+    }
+
+    override fun fetchPopularManga(page: Int): Observable<MangasPage> {
+        val mainArchive = client.newCall(GET(mainArchiveUrl, headers)).execute().asJsoup()
+        val archiveData = mainArchive.select("div.text > table.text > tbody > tr")
+
+        val mangas = mutableListOf<SManga>()
```
vetleledaal, line 126: Avoid mutable structures if possible (probably easier if you migrate away from the `fetch*` methods).

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+                    i += 1
+                }
+            }
+        }
+        chapters.reverse()
+        return chapters
+    }
+
+    private fun absolutePageUrl(pageUrl: String, maybeRelativeTo: String): String =
+        if (pageUrl.startsWith('/')) {
```
vetleledaal: This should be resolved for you if you use `link.absUrl("href")` or `link.attr("abs:href")`, it will resolve the relative path based on the document's `.location()` (set when using `.asJsoup()` (ParsedHttpSource does this for you))

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+            maybeRelativeTo + pageUrl
+        }
+
+    override fun fetchChapterList(manga: SManga): Observable<List<SChapter>> {
+        var archiveResponse = client.newCall(GET(baseUrl + manga.url, headers)).execute()
+        if (!archiveResponse.isSuccessful) {
+            archiveResponse = client.newCall(GET(mainArchiveUrl, headers)).execute()
+        }
+
+        // Fingers crossed they don’t come up with yet another different archive structure.
```
vetleledaal: If you expect it to break easily it might be worth throwing a custom exception.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+        return Observable.just(fetchChaptersForDatedAndDatelessArchives(archiveResponse.asJsoup(), manga.url))
+    }
+
+    override fun fetchPageList(chapter: SChapter): Observable<List<Page>> =
+        Observable.just(listOf(Page(0, chapter.url)))
+
+    override fun fetchImageUrl(page: Page): Observable<String> {
+        val comicPage = client.newCall(GET(baseUrl + page.url, headers)).execute().asJsoup()
+        // Careful. For almost all images it’s `div.center>p>img`, except for pages released on
+        // April’s Fools day, when it’s `div.center>p>a>img`.
```
vetleledaal: Since you're documenting this behavior anyways, please also include a link to where the default assumption breaks.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+        2 -> "https://www.darthsanddroids.net/cast/ObiWan3.jpg" // D&D3
+        3 -> "https://www.darthsanddroids.net/cast/JarJar2.jpg" // JJ
+        4 -> "https://www.darthsanddroids.net/cast/Leia4.jpg" // D&D4
+        5 -> "https://www.darthsanddroids.net/cast/Han5.jpg" // D&D5
+        6 -> "https://www.darthsanddroids.net/cast/Luke6.jpg" // D&D6
+        7 -> "https://www.darthsanddroids.net/cast/Cassian.jpg" // R1
+        8 -> "https://www.darthsanddroids.net/cast/C3PO4.jpg" // Muppets
+        9 -> "https://www.darthsanddroids.net/cast/Finn7.jpg" // D&D7
+        10 -> "https://www.darthsanddroids.net/cast/Han4.jpg" // Solo
+        11 -> "https://www.darthsanddroids.net/cast/Hux8.jpg" // D&D8
```
vetleledaal: From what page are these images linked from? If they're just random images it should be mentioned in a comment.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+    override fun chapterListRequest(manga: SManga): Request = throw UnsupportedOperationException()
+
+    override fun chapterListParse(response: Response): List<SChapter> = throw UnsupportedOperationException()
+
+    override fun pageListParse(response: Response): List<Page> = throw UnsupportedOperationException()
+
+    override fun imageUrlParse(response: Response): String = throw UnsupportedOperationException()
+
+    companion object {
+        private val DATE_FMT = SimpleDateFormat("EEE d MMM, yyyy", Locale.US)
```
vetleledaal, line 243: Since the timezone is available(?) you can add that in the format string or by setting it to PDT/PST/UTC, whichever one it uses for the date field you're reading from.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
vetleledaal: To not repeat the baseUrl you should use `$baseUrl` within the string

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+        val mainArchive = client.newCall(GET(mainArchiveUrl, headers)).execute().asJsoup()
+        val archiveData = mainArchive.select("div.text > table.text > tbody > tr")
+
+        val mangas = mutableListOf<SManga>()
+        var nextMangaTitle = name
+        var nthManga = 0
+
+        run stop@{
+            archiveData.forEach {
+                val maybeTitle = it.selectFirst("th")?.text()?.trim()
```
vetleledaal: `Element.text()` is already trimmed, according to Jsoup's documentation.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+    // deal with.
+    private fun fetchChaptersForDatedAndDatelessArchives(archivePages: Document, archivePath: String): List<SChapter> {
+        val chapters = mutableListOf<SChapter>()
+        var i = 0
+        archivePages.select("""div.text > table.text > tbody > tr""").forEach {
+            val pageData = it.select("""td""")
+            var pageAnchor = pageData.getOrNull(2)?.selectFirst("a")
+            // null for »Intermission«, main archive, dateless archive,…
+            if (pageAnchor != null) {
+                val a = pageAnchor
```
vetleledaal: Can't `pageAnchor` be used directly? Doesn't seem like you're mutating `pageAnchor`, so behavior should be the same.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
vetleledaal: `Element.text()` is already trimmed (assuming `pageData[0]` is of type `Element`), so `.trim()` can be removed

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,222 @@
...
+            // This happens because for whatever reason, the `solo` index isn’t
+            // a `.html` page, but a directory path with an `index.html`. So
+            // we’re returning `solo/$pageUrl`.
+            maybeRelativeTo + pageUrl
+        }
+
+    override fun fetchChapterList(manga: SManga): Observable<List<SChapter>> {
+        var archiveResponse = client.newCall(GET(baseUrl + manga.url, headers)).execute()
+        if (!archiveResponse.isSuccessful) {
+            archiveResponse = client.newCall(GET(mainArchiveUrl, headers)).execute()
```
vetleledaal: Instead of trying 2 URLs, you might be able to set `manga.url` to `/archive.html` for those entries. If you need to pass additional data you can use the fragment part of the URL.

vetleledaal (changes requested):

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: Just to be cautious in case the website has more surprises waiting in the future. I can add it if that’s preferred.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: Will do. It’s a left-over from when i learned my way around with Dark Science. I made these two extensions in parallel.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: Oh it’s just an unfortunate name. Absolute in the sense of starting at the site root, not being relative to the current page. A quirk with *one* of the books on that site, where every book’s archive is a `.html` page having root paths to their comic pages, except `/solo/`, which uses an `index.html` and has relative comic page paths pointing inside `/solo/`. This method appends the `/solo/` part if needed.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: ~~You misunderstood the purpose and committed a bug there. `archivePath` is `/solo/`, or for everything else stuff like `/archive4.html`. When the `href` doesn’t start with a `/`, i.e. when it’s not relative to the website root, the `archivePath` needs to be prepended for links to work.~~

~~This method really needs a better name, am open for ideas.~~

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: The date is »Sun 14 Aug, 2022«. This is the only book for which the date of all pages is known, for all pages were released the same day. For all others there’s but a date range. I added the date just because in this one case i actually have this information, but i can remove it for consistency’s sake.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: Sure. Can you point me at some documentation? I don’t know what that method does or which type implements it.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: Good idea, will try. Any extension that already rate-limits itself i can learn from?

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: Sure, can do. Though the way that method is called, `archiveUrl` is already just a path without `baseUrl`.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: Oh, good to know. Will use that, much cleaner.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
vetleledaal: On mobile, so responses are short.

https://github.com/search?q=repo%3Akeiyoushi%2Fextensions-source%20date_upload%20%3D%20runcatching&type=code

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
vetleledaal: https://github.com/search?q=repo%3Akeiyoushi%2Fextensions-source+client+%3D+network.cloudflareclient+ratelimit&type=code

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: I’ll document it.

They’re from the cast member pages, like here: <https://www.darthsanddroids.net/cast/cast1.html> I’ve tried picking those characters one would intuitively right away associate with the corresponding film, while having each character be used only once as thumbnail. Like Qui-Gon for Phantom Menace. I had to improvise for Solo, though, for that book has no cast member page.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
vetleledaal: If you know the entry that the date applies to you should hardcode that, and basically what you wrote earlier. Assuming it may occur later it's better to default to 0 / 'today'

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: Uhh, i don’t know the timezone, actually. All dates really just match the format string, like: »Sun 13 Jan, 2019«, no additional info. The `Locale.US` only exists to correctly handle day and month names.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: I feel stupid for not having thought of that, thanks. x)

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
vetleledaal: I need to re-review it in context, not able to currently. I don't currently have any name suggestions.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
vetleledaal: I think either is fine, this was mostly a nitpick

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
vetleledaal: The site's footer has a timestamp with a timezone

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: I forgot why exactly i made it that way, but it may have something to do with the `/archive.html` eventually changing. Will investigate a better way. 

I *think* i did it that way because at one point i had the code special-case all comic book initialisation of `SManga`, and book 8 preëmptively had its archive page set to `/archive8.html`. I.e. it’s legacy code.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: Huh, never noticed. Thanks, will add!

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: In the next commit that date is sourced from the site footer, which is the date the corresponding page was published.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: It’s not a CloudFlare site, though. Or is that rate limiter usable anywhere?

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: I didn’t manage to come up with an easy way to get rid of `mutableList` in `popularMangasParse`, for it has to run a state machine that returns early.

I did get rid of `mutableList` in the `chapterList`, though. The next commit also gets rid of almost all `fetch`es.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: Hmm, didn’t see a good opportunity to inject a good custom exception. The existing code’s just too careful to not assume too much.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: The individual page dates have no timezone info, and i’m hesitant to just `+ " PDT"` them. However, i added timezone decoding to »dateless« archives in the next commit.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: Method nuked, only `chapterListParse` now remains.

Evrey:
Pushed an improved version.

There’s a bug i don’t understand, for i neither spot anything in the app’s logs, nor does an exception message pop up in the app. When i try to read the newest chapter released today, <https://www.darthsanddroids.net/episodes/2495.html>, it fails to load the image. If i tell the app to download the image, it works. I couldn’t replicate that behaviour for any other chapter. There’s no difference in the page layout, no difference in URLs, no difference in time formats.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
vetleledaal: Not all apps have added the Cloudflare interceptor by default,
Mihon has it enabled: https://github.com/mihonapp/mihon/blob/22e8050fff273a189d2fdfae6e66ec622f04f1dd/core/src/main/java/eu/kanade/tachiyomi/network/NetworkHelper.kt#L47
J2K does not: https://github.com/Jays2Kings/tachiyomiJ2K/blob/8946395f78d22e44471497d3b76de661e9b4166e/app/src/main/java/eu/kanade/tachiyomi/network/NetworkHelper.kt#L54

You should use `network.cloudflareClient` for consistency. At least that's what the norm is, AFAIK.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,218 @@
...
+
+    override fun popularMangaParse(response: Response): MangasPage {
+        val mainArchive = response.asJsoup()
+        val archiveData = mainArchive.select("div.text > table.text > tbody > tr")
+
+        val mangas = mutableListOf<SManga>()
+        var nextMangaTitle = name
+        var nthManga = 0
+
+        run stop@{
```
AwkwardPeak7, line 130: why `run @stop`?
just loop through it directly

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,218 @@
...
+    // We also can’t really use the `mangaDetailsRequest + mangaDetailsParse`
+    // approach, for we actually expect one of the books’ `url`s to change.
+    override fun fetchMangaDetails(manga: SManga): Observable<SManga> =
+        fetchPopularManga(0)
+            .map { mangasPage ->
+                mangasPage
+                    .mangas
+                    // Do not test for URL-equality, for the last book will always
+                    // eventually migrate its archive page from `/archive.html` to
+                    // its own page.
```
AwkwardPeak7, line 163: can you clarify? manga's url can't be changed once set

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
```diff
@@ -0,0 +1,218 @@
...
+
+    override fun searchMangaParse(response: Response): MangasPage = throw UnsupportedOperationException("searchMangaParse")
+
+    override fun latestUpdatesRequest(page: Int): Request = throw UnsupportedOperationException("latestUpdatesRequest")
+
+    override fun latestUpdatesParse(response: Response): MangasPage = throw UnsupportedOperationException("latestUpdatesParse")
+
+    override fun imageUrlRequest(page: Page): Request = throw UnsupportedOperationException("imageUrlRequest")
+
+    override fun imageUrlParse(response: Response): String = throw UnsupportedOperationException("imageUrlParse")
```
AwkwardPeak7: remove extra new lines and the label from the exception

AwkwardPeak7 (changes requested):

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: I can’t, or rather, i’d have to awkwardly set a boolean to turn the rest of the loop into a no-op.

It’s one big HTML table that at first is an awkward list of previous books, then all that follows is a list of pages of the currently releasing book. So the `return@stop` skips all the current book pages.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: It’s a weird site. Their main archive is `/archive.html`. But that archive is a combined feed of previously released and completed books, as well as *all* pages of the currently releasing book. Once the current book is done, all the pages disappear from that feed, the last book gets its own complete archive page (in this case likely `/archive8.html`), and `/archive.html` is repurposed for their next book (likely Star Wars 9), until the end.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
AwkwardPeak7: a regular for loop can be broken using `break`. or am I missing something?
also `return@forEach` should work too

Evrey: I vaguely remember `return@forEach` not having behaved like `break` but like `continue` in other languages, hence the `run`. I *think* i also tried `break`, but i don’t remember. Kotlin’s weird, coming from other langs. I’ll try the `break`.

AwkwardPeak7: I might be misremembering then..., should be fine then

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: Added to the next commit.

src/en/darthsdroids/src/eu/kanade/tachiyomi/extension/en/darthsdroids/DarthsDroids.kt
Evrey: I documented this weirdness, so that other devs won’t be as surprised when picking up this extension or trying to learn from it.

Evrey:
The bug mentioned in my previous comment https://github.com/keiyoushi/extensions-source/pull/4157#issuecomment-2244903945 still persists, i still don’t understand it. The other issues should be resolved now, though.

AwkwardPeak7 (approved):

PR #4325, opened 2024-07-29
AnimeSama: Add support for multiple scan versions (#4156)

### LINKED ISSUE #4156: AnimeSama default source -> colorless
### Source name

AnimeSama

### Source language

French

It seems that by default AnimeSama selects the color version of the manga (I only check One Piece). Thus it is not necessarily the most up to date version of the manga.
Would it be possible to take the colorless version instead?

### Other details

No

### Acknowledgements

CriosChan:
Closes #4156

Checklist:

### DIFF
--- a/src/fr/animesama/src/eu/kanade/tachiyomi/extension/fr/animesama/AnimeSama.kt
+++ b/src/fr/animesama/src/eu/kanade/tachiyomi/extension/fr/animesama/AnimeSama.kt
@@ -101,35 +101,10 @@ class AnimeSama : ParsedHttpSource() {
         return count > 1
     }
 
-    override fun chapterListRequest(manga: SManga): Request {
-        var url = "$baseUrl${manga.url}".toHttpUrl().newBuilder()
+    private fun parseChapterFromResponse(response: Response, translation_name: String): List<SChapter> {
+        val document = response.asJsoup()
 
-        val fetchExistentSubMangas = GET(url.build(), headers)
-        val page = client.newCall(fetchExistentSubMangas).execute()
-        val document = page.asJsoup()
-        val scriptContent = document.select("script:containsData(panneauScan(\"nom\", \"url\"))").toString()
-        var splitedContent = scriptContent.split(";").toMutableList()
-        // Remove exemple
-        splitedContent.removeAt(0)
-
-        url.addPathSegments("scan/vf")
-
-        splitedContent.forEach { line ->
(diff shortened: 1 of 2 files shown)


src/fr/animesama/src/eu/kanade/tachiyomi/extension/fr/animesama/AnimeSama.kt
```diff
@@ -62,19 +63,52 @@ class AnimeSama : ParsedHttpSource() {
...
-    override fun searchMangaSelector() = "a[href*='catalogue']"
+
+    override fun searchMangaParse(response: Response): MangasPage {
+        val entries: MutableList<SManga> = mutableListOf()
+
+        val document = response.asJsoup()
+
+        document.select(searchMangaSelector()).forEach { element ->
+            val fetchExistentSubMangas = GET(element.select("a").attr("href"), headers)
+            val page = client.newCall(fetchExistentSubMangas).execute()
```
AwkwardPeak7: this will result in n calls for n manga, please avoid this.

AwkwardPeak7 (changes requested):

src/fr/animesama/src/eu/kanade/tachiyomi/extension/fr/animesama/AnimeSama.kt
CriosChan: Would it be ok to limit the number of manga using the page system? I was also thinking about the possibility of making scanlators?

AwkwardPeak7: Pagination is an option but please avoid making individual calls for each manga in browse. 
I don't follow what you mean by scanlators? You can add them to chapters of that what you are asking

src/fr/animesama/src/eu/kanade/tachiyomi/extension/fr/animesama/AnimeSama.kt
```diff
@@ -100,16 +102,41 @@ class AnimeSama : ParsedHttpSource() {
     }
 
     override fun chapterListRequest(manga: SManga): Request {
```
AwkwardPeak7: request methods should ideally only return a request, you can make more requests in the parse method

AwkwardPeak7 (changes requested):

AwkwardPeak7 (approved):

PR #4398, opened 2024-08-03
More scanlator stuff for batoto

### LINKED ISSUE #3957: Batoto empty groups
### Source name

Batoto 1.4.36

### Source language

Multi

Some batoto entries doesn't have a group in it, but only an user. [example](https://battwo.com/series/70745/renchin)
It would be really good if you guys added something like:
if the group is empty, use the username instead as scanlator, so it would be possible to filter out those kind of chapters that doesn't contain a group

### Other details

_No response_

### Acknowledgements

kana-shii:
closes #3957 
Checklist:

### DIFF
--- a/src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
+++ b/src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
@@ -394,7 +394,7 @@ open class BatoTo(
         if (group.isNullOrBlank()) {
             chapter.scanlator = user
         }
-        if (user.isNullOrBlank()) {
+        else if (user.isNullOrBlank()) {
             chapter.scanlator = "Unknown"
         }
 
--- a/src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
+++ b/src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
@@ -388,16 +388,11 @@ open class BatoTo(
         val time = element.select("div.extra > i.ps-3").text()
         chapter.setUrlWithoutDomain(urlElement.attr("href"))
         chapter.name = urlElement.text()
-        if (group != "") {
-            chapter.scanlator = group
+        chapter.scanlator = when {
+            group.isNotBlank() -> group
+            user.isNotBlank() -> user
+            else -> "Unknown"
         }
-        if (group.isNullOrBlank()) {
-            chapter.scanlator = user
-        }
-        else if (user.isNullOrBlank()) {
-            chapter.scanlator = "Unknown"
-        }
-
         if (time != "") {
             chapter.date_upload = parseChapterDate(time)
         }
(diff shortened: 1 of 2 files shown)


src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
```diff
@@ -384,12 +384,20 @@ open class BatoTo(
...
         if (group != "") {
             chapter.scanlator = group
         }
+        if (group.isNullOrBlank()) {
+            chapter.scanlator = user
+        }
+        if (user.isNullOrBlank()) {
+            chapter.scanlator = "Unknown"
+        }
+
```
AwkwardPeak7: use else if

kana-shii:
I'm gonna look for what went wrong...  (i'm a dumbass mb. figured it out)

src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
kana-shii: I tried using `else if` on both, it didn't work... (not in the third commit because I'm a dumbass sry)
I tried using `else` instead of `else if` in the `user.isNullOrBlank` and it also didn't work. (both builds were successful though)
The most recent commit was the only way I found to make everything work... although, if there's a better way, please do tell me and I'll try to fix it again

src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
```diff
@@ -384,11 +384,14 @@ open class BatoTo(
...
+        val user = element.select("div.extra > a.ps-3").text()
         val time = element.select("div.extra > i.ps-3").text()
         chapter.setUrlWithoutDomain(urlElement.attr("href"))
         chapter.name = urlElement.text()
-        if (group != "") {
-            chapter.scanlator = group
+        chapter.scanlator = when {
+            group.isNotBlank() -> group
+            user.isNotBlank() -> user
+            else -> "Unknown"
```
FourTOne5, line 394:
```suggestion
            else -> null
```

kana-shii: If I do this, there will be no way to filter out the chapter in the app.

FourTOne5: Make a feature request on the app then

kana-shii: Shouldn't it be like this for now as a workaround?
I did the same for the mangago extension

kana-shii: > Make a feature request on the app then

Actually, the app dev himself said that this feature wouldn't be added https://github.com/mihonapp/mihon/issues/170

AwkwardPeak7 (approved):

PR #4642, opened 2024-08-16
[NHentai] Fix language filter

ZIDOUZI:
Do not use namespace `language:` will search in all place and could matched in title or other place, which is not we want.

Checklist:

### DIFF
--- a/src/all/nhentai/build.gradle
+++ b/src/all/nhentai/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'NHentai'
     extClass = '.NHFactory'
-    extVersionCode = 42
+    extVersionCode = 43
     isNsfw = true
 }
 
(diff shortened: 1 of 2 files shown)


ZIDOUZI:
Also I would like to refer to some opinion, is whether it is possible to make the following changes:
```kotlin
        val nhLangSearch = if (nhLang.isBlank() || ("language:" in fixedQuery && "language:translated" !in fixedQuery)) "" else "+language:$nhLang "
```
that would be helpful to switch language filter when people want.

AwkwardPeak7:
if people want multiple language results they should use the `all` language variant and apply filters manually

ZIDOUZI:
ok, that would not be applied

AwkwardPeak7 (approved):

PR #4657, opened 2024-08-17
[NHentai] fix several bugs about search

ZIDOUZI:
1. remove `+` before language since they have different semantics (add `+` before means this is not concern as a namespace, that time `+` will be encoded to `%2B`).
2. change fallback value of `fixedQuery` to empty string rather than a quote, since they may have different semantics
3. change search url since there have a redirection.
4. add colon before tag. I forgotten in last pr, sorry about that.

Checklist:


towermom9:
yo tested it and the search is now completely broken?

towermom9:
apparently if I search using the "English" version it doesn't work

towermom9:
but if I search with the "multi" one it works

ZIDOUZI:
> yo tested it and the search is now completely broken?

before #4642 the language filter behave as `+language: specificLanguage`, that is not correct but work, I througt there is only one bug and fix it, now the remains bug destroyed the search

ZIDOUZI:
> but if I search with the "multi" one it works

since the multi language not apply `+lanugage:xxx`, it will work.

towermom9:
example:
Screenshot_2024-08-17-13-23-14-313_app mihon-edit
Screenshot_2024-08-17-13-23-41-459_app mihon-edit


Screenshot_2024-08-17-13-22-53-703_app mihon-edit
Screenshot_2024-08-17-13-24-24-135_app mihon

towermom9:
> > yo tested it and the search is now completely broken?
> 
> before #4642 the language filter behavior as `+language: specificLanguage`, that is not correct but work, I througt there is only one bug and fix it, now the remains bug destroyed the search

oh Ic 😅

AwkwardPeak7:
it seems you are not testing your prs... Please do test them

ZIDOUZI:
> it seems you are not testing your prs... Please do test them

Already test. This will work well.

src/all/nhentai/src/eu/kanade/tachiyomi/extension/all/nhentai/NHentai.kt
```diff
@@ -136,23 +136,23 @@ open class NHentai(
     }
 
     override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
-        val fixedQuery = query.ifEmpty { "\"\"" }
+        val fixedQuery = query.ifEmpty { "" }
```
vetleledaal: `String.ifEmpty { "" }` doesn't make sense.

AwkwardPeak7 (approved):

PR #4863, opened 2024-08-31
Comics Valley: fix latest and popular

### LINKED ISSUE #2575: Comics Valley: Latest & Popular page broken
### Source information

Comics Valley 1.4.36

### Source language

Hindi

### Steps to reproduce

Open Latest & Popular page, no results found

### Expected behavior

Results found

### Actual behavior

No results found

### Mihon/Tachiyomi version

0.16.5

### Android version

Android 14

### Other details

Searching still works, I think it needs a /home-3/ after base URL

### Acknowledgements

dngonz:
Closes #2575

Checklist:

### DIFF
--- a/src/hi/comicsvalley/build.gradle
+++ b/src/hi/comicsvalley/build.gradle
@@ -3,7 +3,7 @@ ext {
     extClass = '.ComicsValley'
     themePkg = 'madara'
     baseUrl = 'https://comicsvalley.com'
-    overrideVersionCode = 1
+    overrideVersionCode = 2
     isNsfw = true
 }
 
--- a/src/hi/comicsvalley/src/eu/kanade/tachiyomi/extension/hi/comicsvalley/ComicsValley.kt
+++ b/src/hi/comicsvalley/src/eu/kanade/tachiyomi/extension/hi/comicsvalley/ComicsValley.kt
@@ -2,4 +2,10 @@ package eu.kanade.tachiyomi.extension.hi.comicsvalley
 
 import eu.kanade.tachiyomi.multisrc.madara.Madara
 
-class ComicsValley : Madara("Comics Valley", "https://comicsvalley.com", "hi")
+class ComicsValley : Madara(
+    "Comics Valley",
+    "https://comicsvalley.com",
+    "hi",
+) {
+    override val mangaSubString = "comics-new"
+}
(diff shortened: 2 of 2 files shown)


vetleledaal:
You could add this to skip a request or two:
```kt
override val useNewChapterEndpoint = true
override val useLoadMoreRequest = LoadMoreStrategy.Always
```

Verify that it actually works, though.

dngonz:
Checked, but unfortunately not, it still does the same requests.

vetleledaal:
> Checked, but unfortunately not, it still does the same requests.

You need to restart the app completely, otherwise it will remember what endpoint to use. If you already did do that there is some flaw in my testing methodology.

dngonz:
Same result maybe I missing something, tried multiple times in different ways but nothing. Apart from the initial requests to get the `.jpg` it always does the same three calls.
Most probably I was doing the tests wrong, could you please tell me how you do it?

vetleledaal:
Not sure about the best way to annotate this, but...:

#### Popular page
1. GET /?s=genre&post_type=wp-manga 403
1. GET /comics-new/?m_orderby=views 200
1. POST /wp-admin/admin-ajax.php (action=madara_load_more&page=1) 200

(1) tries to get the genre and is unrelated to what I mentioned earlier (but should probably still be looked into)
(2) gets page 1, sees that it should use the madara_load_more API instead
(3) is also page 1, but through the madara_load_more API

Here, (2) can be skipped by using `override val useLoadMoreRequest = LoadMoreStrategy.Always`

#### Manga page
1. GET /adult-comics/velamma-latest-hindi-episodes/ 200
1. GET /adult-comics/velamma-latest-hindi-episodes/ 200
1. POST /wp-admin/admin-ajax.php (action=manga_views) 200
1. POST /wp-admin/admin-ajax.php (action=manga_get_chapters) 400
1. GET /adult-comics/velamma-latest-hindi-episodes/ajax/chapters 200

(1) and (2) are for manga details and chapter data. The duplicate request is a limitation of the extension API
(3) is for counting views and can be ignored
(4) tries to get the chapters from the old API since the chapters weren't found in (2)
(5) gets the chapter data from the new API, since it wasn't found in (4)

We can skip (2) and (4) by using `override val useNewChapterEndpoint = true`

#### ---
Madara extensions will save these strategies for subsequent popular/manga requests, so these additional requests will only show up the first time.

Another indication that you should set these explicitly is that it still works as normal. If you're still unsure about some nuances, you would be better off reading the source code of the Madara class.

dngonz:
Got it, now I see the differences. Improvement added.

AwkwardPeak7 (approved):

PR #4874, opened 2024-08-31
Nicovideo: Improve error handle on fetch page list

ringosham:
A small enhancement to check if the user has logged in or not.
Also changed the error text to Japanese.

Checklist:

### DIFF
--- a/src/ja/nicovideoseiga/src/eu/kanade/tachiyomi/extension/ja/nicovideoseiga/NicovideoSeiga.kt
+++ b/src/ja/nicovideoseiga/src/eu/kanade/tachiyomi/extension/ja/nicovideoseiga/NicovideoSeiga.kt
@@ -166,7 +166,7 @@ class NicovideoSeiga : HttpSource() {
                                     302 -> {
                                         // User needs to login via WebView first before accessing the chapter
                                         // "Please login via WebView first"
-                                        Observable.error(SecurityException("WebViewでログインしてください"))
+                                        Observable.error(SecurityException("まず、WebViewでログインしてください"))
                                     }
 
                                     else -> Observable.error(Exception("HTTP error ${login.code}"))
--- a/src/ja/nicovideoseiga/src/eu/kanade/tachiyomi/extension/ja/nicovideoseiga/NicovideoSeiga.kt
+++ b/src/ja/nicovideoseiga/src/eu/kanade/tachiyomi/extension/ja/nicovideoseiga/NicovideoSeiga.kt
@@ -159,7 +159,7 @@ class NicovideoSeiga : HttpSource() {
                                         // User needs to purchase the chapter on the official mobile app
                                         // Sidenote: Chapters can't be purchased on the site
                                         // These paid chapters only show up on the mobile app and are straight up hidden on browsers! Why!?
-                                        // "Please purchase the chapter on the official app"
+                                        // "Please buy from the official app"
                                         Observable.error(SecurityException("公式アプリで購入してください"))
                                     }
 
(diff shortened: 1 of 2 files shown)


src/ja/nicovideoseiga/src/eu/kanade/tachiyomi/extension/ja/nicovideoseiga/NicovideoSeiga.kt
```diff
@@ -140,20 +140,43 @@ class NicovideoSeiga : HttpSource() {
...
+                            .newCall(GET("https://www.nicovideo.jp/my"))
+                            .asObservable()
+                            .flatMap { login ->
+                                when (login.code) {
+                                    200 -> {
+                                        // User needs to purchase the chapter on the official mobile app
+                                        // Sidenote: Chapters can't be purchased on the site
+                                        // These paid chapters only show up on the mobile app and are straight up hidden on browsers! Why!?
+                                        // "Please purchase the chapter on the official app"
+                                        Observable.error(SecurityException("公式アプリで購入してください"))
```
vetleledaal, line 163: Maybe mention that the **chapter** specifically needs to be purchased, so it's consistent with the English comment? Consider (if not directly lifted from the site/app):
```kt
Observable.error(SecurityException("公式アプリで章を購入してください"))
//                                            +++
```

Proposed change might not be accurate as I'm using MTL, I am just pointing out the discrepancy.

src/ja/nicovideoseiga/src/eu/kanade/tachiyomi/extension/ja/nicovideoseiga/NicovideoSeiga.kt
```diff
@@ -140,20 +140,43 @@ class NicovideoSeiga : HttpSource() {
...
+                                    }
+
+                                    302 -> {
+                                        // User needs to login via WebView first before accessing the chapter
+                                        // "Please login via WebView first"
+                                        Observable.error(SecurityException("WebViewでログインしてください"))
```
vetleledaal: JP text doesn't include 'first', is that implied? Consider (if not directly lifted from the site/app):
 ```kt
Observable.error(SecurityException("まず、WebViewでログインしてください"))
//                                  +++++
```

Proposed change might not be accurate as I'm using MTL, I am just pointing out the discrepancy.

src/ja/nicovideoseiga/src/eu/kanade/tachiyomi/extension/ja/nicovideoseiga/NicovideoSeiga.kt
ringosham: About that, I was considering that as well.
I was using GPT for both error messages, and not including "章" seems to flow better as the user is already tapping into a chapter providing a context. 

EDIT: If I'm correct, this sentence would be more like "Please buy from the official app" or something.
If anyone understands Japanese more, feel free to correct me.

src/ja/nicovideoseiga/src/eu/kanade/tachiyomi/extension/ja/nicovideoseiga/NicovideoSeiga.kt
ringosham: Applied the change for this one.

src/ja/nicovideoseiga/src/eu/kanade/tachiyomi/extension/ja/nicovideoseiga/NicovideoSeiga.kt
Norsze: I don't know about "flow", but I would consider "公式アプリで購入するとアクセス得られる" (Purchase on official app to gain access), but it might be a long toast. Just giving options, still only a N3 in japanese

vetleledaal: > EDIT: If I'm correct, this sentence would be more like "Please buy from the official app" or something.

Regardless of what you pick, the English comment above the error message should be derived from the Japanese error message and not simply what it ought to be.

Given the additional context you provided, omitting "章を" should be fine.

I don't have any comment on the alternative provided by @Norsze.

ringosham: I'll keep the current text as is. You only get 2 seconds to read after all.
I changed the original English comment to match the Japanese text instead.

AwkwardPeak7 (approved):

PR #5017, opened 2024-09-12
MangaThemesia: handle NullPointerException

Riztard:
- handle NullPointerException for title selector in mangaDetailsParse

Checklist:

### DIFF
--- a/lib-multisrc/mangathemesia/src/eu/kanade/tachiyomi/multisrc/mangathemesia/MangaThemesia.kt
+++ b/lib-multisrc/mangathemesia/src/eu/kanade/tachiyomi/multisrc/mangathemesia/MangaThemesia.kt
@@ -240,7 +240,7 @@ abstract class MangaThemesia(
 
     override fun mangaDetailsParse(document: Document) = SManga.create().apply {
         document.selectFirst(seriesDetailsSelector)?.let { seriesDetails ->
-            title = seriesDetails.selectFirst(seriesTitleSelector)!!.text()
+            title = seriesDetails.selectFirst(seriesTitleSelector)?.text() ?: ""
             artist = seriesDetails.selectFirst(seriesArtistSelector)?.ownText().removeEmptyPlaceholder()
             author = seriesDetails.selectFirst(seriesAuthorSelector)?.ownText().removeEmptyPlaceholder()
             description = seriesDetails.select(seriesDescriptionSelector).joinToString("\n") { it.text() }.trim()
--- a/lib-multisrc/mangathemesia/build.gradle.kts
+++ b/lib-multisrc/mangathemesia/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 30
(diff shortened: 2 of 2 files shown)


lib-multisrc/mangathemesia/src/eu/kanade/tachiyomi/multisrc/mangathemesia/MangaThemesia.kt
```diff
@@ -240,7 +240,7 @@ abstract class MangaThemesia(
 
     override fun mangaDetailsParse(document: Document) = SManga.create().apply {
         document.selectFirst(seriesDetailsSelector)?.let { seriesDetails ->
-            title = seriesDetails.selectFirst(seriesTitleSelector)!!.text()
+            title = seriesDetails.selectFirst(seriesTitleSelector)?.text() ?: ""
```
AwkwardPeak7, line 243: selector should be updated in the source instead

lib-multisrc/mangathemesia/src/eu/kanade/tachiyomi/multisrc/mangathemesia/MangaThemesia.kt
```diff
@@ -240,7 +240,7 @@ abstract class MangaThemesia(
 
     override fun mangaDetailsParse(document: Document) = SManga.create().apply {
         document.selectFirst(seriesDetailsSelector)?.let { seriesDetails ->
```
AwkwardPeak7, line 242:
```suggestion
        document.selectFirst(seriesDetailsSelector)!!.let { seriesDetails ->
```

AwkwardPeak7 (changes requested):

lib-multisrc/mangathemesia/src/eu/kanade/tachiyomi/multisrc/mangathemesia/MangaThemesia.kt
Riztard: yea but secondary title from manga details is usually is just to update the one from popular. so its not really that important if its suddenly not working and it shouldnt interfere with the other details while its broken

lib-multisrc/mangathemesia/src/eu/kanade/tachiyomi/multisrc/mangathemesia/MangaThemesia.kt
Riztard: still throwing null pointer with only that

AwkwardPeak7: this should be there regardless

lib-multisrc/mangathemesia/src/eu/kanade/tachiyomi/multisrc/mangathemesia/MangaThemesia.kt
AwkwardPeak7: breaks deeplinks

PR #5039, opened 2024-09-13
Mangago regex edit

kana-shii:
Checklist:

### DIFF
--- a/src/en/mangago/src/eu/kanade/tachiyomi/extension/en/mangago/Mangago.kt
+++ b/src/en/mangago/src/eu/kanade/tachiyomi/extension/en/mangago/Mangago.kt
@@ -41,6 +41,7 @@ import java.util.Locale
 import javax.crypto.Cipher
 import javax.crypto.spec.IvParameterSpec
 import javax.crypto.spec.SecretKeySpec
+import androidx.preference.EditTextPreference
 
 class Mangago : ParsedHttpSource(), ConfigurableSource {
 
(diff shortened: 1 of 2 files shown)


src/en/mangago/src/eu/kanade/tachiyomi/extension/en/mangago/Mangago.kt
```diff
@@ -574,9 +586,33 @@ class Mangago : ParsedHttpSource(), ConfigurableSource {
...
+                if (regexPattern.isBlank()) {
+                    text = defaultValue
+                    false
+                } else {
+                    preferences.edit().putString("TITLE_REGEX_PATTERN", regexPattern).apply()
+                    titleRegex = Regex(regexPattern, RegexOption.IGNORE_CASE)
+                    true
+                }
+            }
+        }.let(screen::addPreference)
```
AwkwardPeak7: not needed

AwkwardPeak7 (changes requested):
undo the additional formatting changes

src/en/mangago/src/eu/kanade/tachiyomi/extension/en/mangago/Mangago.kt
kana-shii: Why exactly is it not needed? Without it the regex wouldn't be customizable by the user

AwkwardPeak7 (approved):

PR #5099, opened 2024-09-17
Add hanmanwuzhe source

heagan01:
Closes #5086 

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [ ] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [x] Have removed `web_hi_res_512.png` when adding a new extension

PR #5922, opened 2024-11-07
Revert "Add hanmanwuzhe source"

stevenyomi:
Reverts keiyoushi/extensions-source#5099

The source name clashes with existing one because it's copied.

cc @AwkwardPeak7


AwkwardPeak7:
the name could have be changed...

stevenyomi:
CI is fast so I'd rather first remove it to prevent it from polluting the repository.

PR #6029, opened 2024-11-12
LuraToon: Refactor to use API JSON without scraper

### LINKED ISSUE #5739: LuraToon No pages found
### Source information

Lura Toon 1.4.54

### Source language

Brazilian Portuguese 

### Steps to reproduce

Screenshot_20241028-064102_Mihon

"Nenhuma página encontrada" It means No page found
But, the webview works perfectly, I can read through the webview

### Expected behavior

It was supposed to open the chapter normally, it was working until yesterday

### Actual behavior

When I press to read the chapter, page not found appears. I've tried several ways to solve it, but I believe there was a change on the site. It's the same domain https://luratoons.com but something has changed on the site. Now it has devtool protection and is no longer Peach Scan, I believe it goes something like this: https://github.com/theajack/disable-devtool

### Mihon/Tachiyomi version

App version: 0.17.0-6971 (standard, 38d5fc916, 8, 2024-10-26T21:17:45Z) Android version: 12 (SDK 31; build SP1A.210812.016.N975FXXS9HWG3) Device brand: samsung Device manufacturer: samsung Device name: d2s (d2sxx) Device model: SM-N975F WebView: Android System WebView 131.0.6778.14

### Android version

Android 12

### Other details

This Scan works by subscription, my account is linked to Google so I can't share it, But if there is any information that needs a signature to obtain, please tell me how to do it. I can also pay you a one month subscription if you are a contributor.  

This is the view of the website reader

Screenshot_20241028-065242_Mihon

### Acknowledgements

SMCodesP:
Closes #5739 

Steps:
- [x] Fix list chapters of manga
- [x] Fix list images of chapter
- [x] Fix read image (Difficult process, follow below in the comments....)
- [x] Fix list latest
- [x] Fix request to details manga

Checklist:


SMCodesP:
I started modifying an extension called LuraToons, which extends from PeachScan's multisrc. However, the owners made many changes, and nowadays the only things that work in the extension are the popular listing and search functionalities. So, I rewrote the chapter and page listing methods. But one of the changes the owners made was to implement a kind of encryption on the images, in addition to compressing them into zip files like Peach already supports. They also added encryption, and I've been struggling to reverse-engineer it, but I don't think I'm close to a satisfactory result yet.

SMCodesP:
Below is an example of an link to the JS script that handles the decryption. However, the script is obfuscated. I've discovered a few things using the JavaScript debugger in devtools. The first parameter of the function _0x13c517 appears to be an ArrayBuffer of the encrypted file, and the second parameter is a object. The decryption seems to depend on this second object

script: https://luratoons.com/static/assets/cngv-nBW.js

SMCodesP:
Job finished...

AnonymousPtBr:
My hero!

AwkwardPeak7 (changes requested):
good work 👍, some comments:

- almost nothing is being used from peachscan multisrc, so remove it from the mulstirc
- you may want to move the zip interceptor and related code to a lib, to share between the multisrc and this source (or just copy it for now)
- don't serialize everything in the serializable classes, only required fields

SMCodesP:
> good work 👍, some comments:
> 
> * almost nothing is being used from peachscan multisrc, so remove it from the mulstirc
> * you may want to move the zip interceptor and related code to a lib, to share between the multisrc and this source (or just copy it for now)
> * don't serialize everything in the serializable classes, only required fields

Thank you for the feedback! I agree that not much is being used from Peachscan's multisrc extension, but there are still some utility functions that are being used. I tried rewriting the code locally without multisrc, but it ended up looking a bit messy in my opinion.

To be honest, I'm not very familiar with Android development, Kotlin, or Gradle, so I'm not sure how to refactor it in a clean and proper way. I only learned enough to get it running 😅.

If anyone else wants to take over this pull request, please feel free to do so. For now, I'll just use a local build.

SMCodesP:
I noticed that if I already have some manga saved in my library, only a few of them have their names changed between the main site and the API. Because of this, I had to migrate those specific manga. Should I add an alert for this?

src/pt/randomscan/src/eu/kanade/tachiyomi/extension/pt/randomscan/LuraToon.kt
```diff
@@ -3,61 +3,191 @@ package eu.kanade.tachiyomi.extension.pt.randomscan
...
+        val capitulo = response.parseAs<CapituloPagina>()
         val pathSegments = response.request.url.pathSegments
         if (pathSegments.contains("login") || pathSegments.isEmpty()) {
             throw Exception("Faça o login na WebView para acessar o contéudo")
         }
-        return super.pageListParse(response)
+        val files = (0 until capitulo.files).map { i ->
+            Page(i, baseUrl, "$baseUrl/api/cap-download/${capitulo.obra.id}/${capitulo.id}/$i?obra_id=${capitulo.obra.id}&cap_id=${capitulo.id}&slug=${pathSegments[2]}&cap_slug=${pathSegments[3]}")
+        }
+        return files
```
AwkwardPeak7:
```suggestion
        return (0 until capitulo.files).map { i ->
            Page(i, baseUrl, "$baseUrl/api/cap-download/${capitulo.obra.id}/${capitulo.id}/$i?obra_id=${capitulo.obra.id}&cap_id=${capitulo.id}&slug=${pathSegments[2]}&cap_slug=${pathSegments[3]}")
        }
```

src/pt/randomscan/src/eu/kanade/tachiyomi/extension/pt/randomscan/LuraToon.kt
```diff
@@ -3,61 +3,191 @@ package eu.kanade.tachiyomi.extension.pt.randomscan
...
+
+        return comics.caps.sortedBy {
+            -it.num
+        }.map { chapterFromElement(manga, it) }
+    }
+
+    private fun chapterFromElement(manga: SManga, capitulo: Capitulo) = SChapter.create().apply {
+        val capSlug = capitulo.slug.trimStart('/')
+        val mangaUrl = manga.url.trimEnd('/').trimStart('/')
+        setUrlWithoutDomain("/api/obra/$mangaUrl/$capSlug")
```
AwkwardPeak7, line 131: throw exception in pagelistrequest (to refresh manga) if old urls are different

src/pt/randomscan/src/eu/kanade/tachiyomi/extension/pt/randomscan/LuraToon.kt
```diff
@@ -3,61 +3,191 @@ package eu.kanade.tachiyomi.extension.pt.randomscan
...
+        .addInterceptor(LuraZipInterceptor()::zipImageInterceptor)
         .rateLimit(3)
         .setRandomUserAgent(
             preferences.getPrefUAType(),
             preferences.getPrefCustomUA(),
         )
         .build()
 
+    override fun chapterListRequest(manga: SManga) = GET("$baseUrl/api/obra${manga.url}", headers)
+    override fun mangaDetailsRequest(manga: SManga) = GET("$baseUrl/api/obra/${manga.url.trimStart('/')}", headers)
```
AwkwardPeak7: point one of them to the other, also are old urls compatible?

src/pt/randomscan/src/eu/kanade/tachiyomi/extension/pt/randomscan/LuraToon.kt
```diff
@@ -3,61 +3,191 @@ package eu.kanade.tachiyomi.extension.pt.randomscan
...
+                chapterListParse(manga, response)
+            }
+    }
+
+    fun chapterListParse(manga: SManga, response: Response): List<SChapter> {
+        val comics = response.parseAs<Manga>()
+
+        return comics.caps.sortedBy {
+            -it.num
+        }.map { chapterFromElement(manga, it) }
```
AwkwardPeak7:
```suggestion
        return comics.caps.sortedByDescending {
            it.num
        }.map { chapterFromElement(manga, it) }
```

src/pt/randomscan/src/eu/kanade/tachiyomi/extension/pt/randomscan/LuraToon.kt
```diff
@@ -3,61 +3,191 @@ package eu.kanade.tachiyomi.extension.pt.randomscan
...
+    private val dateFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.getDefault()).apply {
+        timeZone = TimeZone.getTimeZone("America/Sao_Paulo")
+    }
+
+    override fun imageUrlParse(response: Response): String {
+        TODO("Not yet implemented")
+    }
+
+    override fun chapterListParse(response: Response): List<SChapter> {
+        TODO("Not yet implemented")
```
AwkwardPeak7: throw `UnsupportedOperationException`

lib/cryptoaes/src/main/java/eu/kanade/tachiyomi/lib/cryptoaes/CryptoAES.kt
```diff
@@ -130,4 +131,28 @@ object CryptoAES {
...
+
+    /**
+     * Decrypt a file using AES CTR mode
+     *
+     * @param encryptedData the encrypted data
+     * @param keyBytes the key to decrypt the data
+     * @return the decrypted data
+     */
+
+    fun decryptFile(encryptedData: ByteArray, keyBytes: ByteArray, hashCipher: String = HASH_CIPHER): ByteArray {
```
AwkwardPeak7: implement this directly in the source

AwkwardPeak7 (changes requested):

src/pt/randomscan/src/eu/kanade/tachiyomi/extension/pt/randomscan/LuraToon.kt
SMCodesP: I didn't understand the question

lib/cryptoaes/src/main/java/eu/kanade/tachiyomi/lib/cryptoaes/CryptoAES.kt
SMCodesP: > implement this directly in the source

I disagree with this change but I will make it

src/pt/randomscan/src/eu/kanade/tachiyomi/extension/pt/randomscan/LuraToon.kt
vetleledaal: I didn't make the original comment, but I'm pretty sure it's just for deduplication, since they are so similar:
```kt
override fun chapterListRequest(manga: SManga) = GET("$baseUrl/api/obra/${manga.url.trimStart('/')}", headers)
override fun mangaDetailsRequest(manga: SManga) = chapterListRequest(manga)
// or
override fun mangaDetailsRequest(manga: SManga) = GET("$baseUrl/api/obra/${manga.url.trimStart('/')}", headers)
override fun chapterListRequest(manga: SManga) = mangaDetailsRequest(manga)
```

As for the last question, if the URL changed (i.e. `/manga/123` → `/comic/some-title-1609520326`), old library entries will be broken. You can bump `versionId` to force migration instead.

If the URL format is exactly the same (i.e. `/manga/123` → `/manga/123`) existing library entries will continue to work, and `versionId` should be kept.

More of an advanced topic, but if the URL changed in a predictable manner (i.e. `/manga/123` → `/mangas/item.aspx?id=123`) you could rewrite it to avoid migration. This might cause issues from the browse screen, where entries are shown as being not in the library if the URL is different.

SMCodesP: > I didn't make the original comment, but I'm pretty sure it's just for deduplication, since they are so similar:
> 
> ```kotlin
> override fun chapterListRequest(manga: SManga) = GET("$baseUrl/api/obra/${manga.url.trimStart('/')}", headers)
> override fun mangaDetailsRequest(manga: SManga) = chapterListRequest(manga)
> // or
> override fun mangaDetailsRequest(manga: SManga) = GET("$baseUrl/api/obra/${manga.url.trimStart('/')}", headers)
> override fun chapterListRequest(manga: SManga) = mangaDetailsRequest(manga)
> ```
> 
> As for the last question, if the URL changed (i.e. `/manga/123` → `/comic/some-title-1609520326`), old library entries will be broken. You can bump `versionId` to force migration instead.
> 
> If the URL format is exactly the same (i.e. `/manga/123` → `/manga/123`) existing library entries will continue to work, and `versionId` should be kept.
> 
> More of an advanced topic, but if the URL changed in a predictable manner (i.e. `/manga/123` → `/mangas/item.aspx?id=123`) you could rewrite it to avoid migration. This might cause issues from the browse screen, where entries are shown as being not in the library if the URL is different.

Regarding deduplication, I understand the intention behind it, but personally, I feel it adds a bit more verbosity to the code. This is just my opinion, but while these functions are essentially the same, they serve different purposes. If, in the future, someone needs to maintain them separately, it could potentially lead to some confusion. However, I understand where you all are coming from and appreciate your approach. I'll go ahead and make the changes as you suggested.

It would be the first case, /manga/123 → /comic/some-title-1609520326. The change is not predictable, and I also don't know which URLs have changed—only some of them did, not all.

vetleledaal: In that case you should also bump `versionId` to force migration. 

I don't really care for deduplication, it tends to increase coupling, so there's always a tradeoff.

SMCodesP: > `versionId`para

Ok I will update the versionId

SMCodesP: > In that case you should also bump `versionId` to force migration.
> 
> I don't really care for deduplication, it tends to increase coupling, so there's always a tradeoff.

Ok I will update the versionId

AwkwardPeak7 (approved):

PR #6134, opened 2024-11-18
AsuraScans: Fix pages not found

### LINKED ISSUE #6135: Asura scans "no pages found"
### Source information

Asura Scans 1.4.41

### Source language

English 

### Steps to reproduce

1. Open the extension 
2. Try opening any series in the extension 
3. When we open the ch it says no pages found 

### Expected behavior

It should the pages in chapter of manhwa 

### Actual behavior

When we try opening it says no pages found 

### Mihon/Tachiyomi version

App version: 0.17.0 (standard, 38d5fc9, 8, 2024-10-26T17:39:14Z) Android version: 15 (SDK 35; build BP11.241025.006) Device brand: google Device manufacturer: Google Device name: panther (panther_beta) Device model: Pixel 7 WebView: Android System WebView 130.0.6723.107

### Android version

Android 15

### Other details

It is working fine in webview the problem is in extension.
Also the chapter I was reading a little while ago are still opening but the new chapters or I have read them like yesterday or in past are also not opening 

### Acknowledgements

### 

---

bapeey:
Closes #6135 

Checklist:

### DIFF
--- a/src/en/asurascans/build.gradle
+++ b/src/en/asurascans/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Asura Scans'
     extClass = '.AsuraScans'
-    extVersionCode = 41
+    extVersionCode = 42
 }
 
 apply from: "$rootDir/common.gradle"
--- a/src/en/asurascans/src/eu/kanade/tachiyomi/extension/en/asurascans/AsuraScans.kt
+++ b/src/en/asurascans/src/eu/kanade/tachiyomi/extension/en/asurascans/AsuraScans.kt
@@ -267,9 +267,10 @@ class AsuraScans : ParsedHttpSource(), ConfigurableSource {
     }
 
     override fun pageListParse(document: Document): List<Page> {
-        val scripts = document.select("script:containsData(pages)").joinToString { it.data() }
-        val data = PAGES_REGEX.find(scripts)?.groupValues?.get(1) ?: throw Exception("Failed to find chapter pages")
-        val pageList = json.decodeFromString<List<PageDto>>(data.unescape()).sortedBy { it.order }
+        val scriptData = document.select("script:containsData(self.__next_f.push)")
+            .joinToString("") { it.data().substringAfter("\"").substringBeforeLast("\"") }
+        val pagesData = PAGES_REGEX.find(scriptData)?.groupValues?.get(1) ?: throw Exception("Failed to find chapter pages")
+        val pageList = json.decodeFromString<List<PageDto>>(pagesData.unescape()).sortedBy { it.order }
         return pageList.mapIndexed { i, page -> Page(i, imageUrl = page.url) }
     }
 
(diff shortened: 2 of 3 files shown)


Seyden:
Altough the idea is good, i can tell you by experience that parsing stuff from the next.js script content by relying on property starts and stuff is dangerous and can fail any time.

Next.js is putting this content into chunks and on site render its getting concatted again to a full script.. so what i'm trying to say is, yes it could work for the most time but it can also happen that your pages section gets split into 2 scripts and then your parsing fails.

I relied on parsing this json in the past and it worked for a long time and then broke, honestly i dont even know what factor decides wether the chunks are getting split or not.

bapeey:
> Next.js is putting this content into chunks and on site render its getting concatted again to a full script

should be fixed

Seyden:
> > Next.js is putting this content into chunks and on site render its getting concatted again to a full script
> 
> should be fixed

The next thing that you should consider and honestly not sure what the rule for that is, they have placeholders and reference field values to other places, so its possible that they have a value as reference to another property, not sure if its to reduce duplication of values or what the rule there is. But you can see it when you search for things like $30 or any other value.

bapeey:
> The next thing that you should consider and honestly not sure what the rule for that is, they have placeholders and reference field values to other places, so its possible that they have a value as reference to another property, not sure if its to reduce duplication of values or what the rule there is. But you can see it when you search for things like $30 or any other value.

References shouldn't be a problem because would be weird that the regex can catch them, it's already happening 
image

Seyden:
> > The next thing that you should consider and honestly not sure what the rule for that is, they have placeholders and reference field values to other places, so its possible that they have a value as reference to another property, not sure if its to reduce duplication of values or what the rule there is. But you can see it when you search for things like $30 or any other value.
> 
> References shouldn't be a problem because would be weird that the regex can catch them, it's already happening image

wdym? i actually meant that a image url could be referenced to another place

bapeey:
> wdym? i actually meant that a image url could be referenced to another place

it is
{E6F27787-CFFF-490B-B1D5-4A5AF16302E1}

In the website code there are the 2 versions, but the regex can't catch the version with references

Seyden:
Yes the regex can't catch the version with references, so whats happening if the url of the image is a reference?

I'm not saying that its always a reference, i'm just saying IT CAN be, depending on what the nextjs compiler decides to optimize and how.

bapeey:
> Yes the regex can't catch the version with references, so whats happening if the url of the image is a reference?

When that happens it's like a chain of references not just the url, like this:

```
"pages\":\"$6d\"
6d:[\"$6e\",\"$6f\",\"$70\",\"$71\",\"$72\",\"$73\",\"$74\",\"$75\",\"$76\",\"$77\",\"$78\",\"$79\"
6e:{\"order\":1,\"url\":\"https://gg.asuracomic.net/storage/media/251620/conversions/00-optimized.webp\"}
```
It would also be strange if the compiler created a reference just for a string. In any case, it's a rare case that I will ignore. We have other extensions that parse JSON from Nextjs chunks and have never encountered that problem

AwkwardPeak7 (approved):

PR #6164, opened 2024-11-19
ResetScan: change theme to Madara

### LINKED ISSUE #5890: reset scans updates on two domains
### Source name

Reset Scans

### Source link

https://rspro.xyz/

### Source language

English

### Other details

Reset scans is updating different series on two different domains, one is https://reset-scans.xyz/ and another is https://rspro.xyz/. Current extension pulls from https://reset-scans.xyz/ which is only updating one series
suggest having two extensions 

### Acknowledgements

- [X] I have checked that the extension does not already exist by searching the [extension listing](https://keiyoushi.github.io/extensions/) and verified it does not appear there.

dngonz:
Closes #5890

Checklist:

### DIFF
--- a/src/en/resetscans/src/eu/kanade/tachiyomi/extension/en/resetscans/ResetScans.kt
+++ b/src/en/resetscans/src/eu/kanade/tachiyomi/extension/en/resetscans/ResetScans.kt
@@ -1,14 +1,30 @@
 package eu.kanade.tachiyomi.extension.en.resetscans
 
 import eu.kanade.tachiyomi.multisrc.madara.Madara
+import eu.kanade.tachiyomi.source.model.SChapter
+import org.jsoup.nodes.Element
+import java.text.SimpleDateFormat
+import java.util.Calendar
+import java.util.Locale
 
 class ResetScans : Madara(
     "Reset Scans",
     "https://rspro.xyz",
     "en",
+    dateFormat = SimpleDateFormat("MMM dd yyyy", Locale.US),
 ) {
     // Moved from FuzzyDoodle to Madara
     override val versionId = 3
 
+    override val useNewChapterEndpoint = true
+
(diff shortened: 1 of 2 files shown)


src/en/resetscans/src/eu/kanade/tachiyomi/extension/en/resetscans/ResetScans.kt
```diff
@@ -1,13 +1,14 @@
...
 
-class ResetScans : FuzzyDoodle(
+import eu.kanade.tachiyomi.multisrc.madara.Madara
+
+class ResetScans : Madara(
     "Reset Scans",
-    "https://reset-scans.xyz",
+    "https://rspro.xyz",
     "en",
 ) {
```
vetleledaal, line 12: Missing `dateFormat`

vetleledaal, line 12: You can add this to skip a request:
```kt
override val useNewChapterEndpoint = true
```

dngonz: Added 👍

dngonz: Added but year is not defined so is the current one by default

vetleledaal: The convention is to leave out the `yyyy` if it is not on the site, making it default to 1970.

I'd argue that it should be obvious that the year is unknown, rather than guessing the most recent one. Then you would run into issues with the year wrapping around multiple times if chapters span that long.

dngonz: Removed, now the year is by default 1970

vetleledaal (approved):

PR #6177, opened 2024-11-20
Add LikeMangaIn

dngonz:
Checklist:


src/en/likemangain/src/eu/kanade/tachiyomi/extension/en/likemangain/LikeMangaIn.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import eu.kanade.tachiyomi.multisrc.madara.Madara
+import java.text.SimpleDateFormat
+import java.util.Locale
+
+class LikeMangaIn : Madara(
+    "LikeMangaIn",
+    "https://likemanga.in",
+    "en",
+    dateFormat = SimpleDateFormat("dd MMMM, yyyy", Locale.getDefault()),
```
vetleledaal: This will change behavior based on the locale used by the user. Changing the locale in Android will not change how the site displays this and will break.

dngonz: Changed to US

vetleledaal (approved):

PR #6195, opened 2024-11-20
Add Snowmtl

### LINKED ISSUE #4884: Please add snowmtl extension
### Source name

Snow machine translations

### Source link

https://snowmtl.ru/

### Source language

English

### Other details

_No response_

### Acknowledgements

- [X] I have checked that the extension does not already exist by searching the [extension listing](https://keiyoushi.github.io/extensions/) and verified it does not appear there.

choppeh:
Closes #4884

Checklist:


src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
```diff
@@ -0,0 +1,480 @@
...
+                is GenreList -> {
+                    filter.state.filter(GenreCheckBox::state).forEach { genre ->
+                        url.addQueryParameter("genres", genre.id)
+                    }
+                }
+                else -> {}
+            }
+        }
+
+        return GET(url.build(), headers)
```
vetleledaal, line 97: Since extensions-lib 1.4 you should be able to remove `.build()`

```suggestion
        return GET(url, headers)
```

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
```diff
@@ -0,0 +1,480 @@
...
+        date_upload = parseChapterDate(element.selectFirst("span")?.text())
+    }
+
+    // =============================== Pages ================================
+
+    override fun pageListParse(document: Document): List<Page> {
+        captions.clear()
+
+        val jsonArray = document.selectFirst("div#json-data")?.let {
+            Json.parseToJsonElement(it.ownText()).jsonArray
```
vetleledaal: Consider using `kotlinx.serialization` (DTOs).

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
```diff
@@ -0,0 +1,480 @@
...
+            val root = jsonElement.jsonObject
+            val imageUrl = """https://${root["img_url"]?.jsonPrimitive?.content}"""
+
+            captionsCache(root, imageUrl)
+
+            Page(index, imageUrl = imageUrl)
+        }
+    }
+
+    private fun captionsCache(root: JsonObject, imageUrl: String) {
```
vetleledaal: Will the whole captions logic work for downloads, when you do operations out of order, or after a restart?

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
```diff
@@ -0,0 +1,480 @@
...
+    private fun parseRelativeDate(date: String): Long {
+        val number = Regex("""(\d+)""").find(date)?.value?.toIntOrNull() ?: return 0
+        val cal = Calendar.getInstance()
+
+        return when {
+            date.contains("day") -> cal.apply { add(Calendar.DAY_OF_MONTH, -number) }.timeInMillis
+            date.contains("hour") -> cal.apply { add(Calendar.HOUR, -number) }.timeInMillis
+            date.contains("minute") -> cal.apply { add(Calendar.MINUTE, -number) }.timeInMillis
+            date.contains("second") -> cal.apply { add(Calendar.SECOND, -number) }.timeInMillis
+            date.contains("week") -> cal.apply { add(Calendar.DAY_OF_MONTH, -number * 7) }.timeInMillis
```
vetleledaal: Consider making this case insensitive for resiliency.

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
```diff
@@ -0,0 +1,480 @@
...
+            return chain.proceed(request)
+        }
+
+        val response = chain.proceed(request)
+
+        val bitmap = BitmapFactory.decodeStream(response.body.byteStream())!!
+            .copy(Bitmap.Config.ARGB_8888, true)
+
+        val canvas = Canvas(bitmap)
+        val defaultTextSize = 22.sp // random
```
vetleledaal: Should this information not be derived from the input data?

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
```diff
@@ -0,0 +1,480 @@
...
+    getTextWidths(text.first().toString(), fontWidth)
+    return fontWidth.first()
+}
+
+fun Paint.getCharHeight(): Float {
+    val text = "A"
+    val bounds = Rect()
+    getTextBounds(text, 0, text.length, bounds)
+    return bounds.height().toFloat()
+}
```
vetleledaal: What does `A` represent here? The average size? The max size? I'm assuming you're not using a monospaced font, and want to approximate the size of some text. Why not get the exact size of the actual text you're using?

There are so many caveats if you want to implement it yourself, so I'd prefer brute forcing with the exact dimensions over this.

If there's a reason for doing it manually, (i.e. simplified logic elsewhere) there should be a comment somewhere.

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
```diff
@@ -0,0 +1,480 @@
...
+            color = Color.BLACK
+            style = Paint.Style.FILL_AND_STROKE
+            textSize = defaultTextSize
+            isAntiAlias = true
+            typeface = Typeface.SANS_SERIF
+        }
+
+        val marginTop = 30 // random
+        val marginLeft = 30 // random
+        val spaceBetween = 2 // random
```
vetleledaal: The word you are looking for is 'arbitrary'. It is in any case a good idea to highlight :+1:

vetleledaal (changes requested):

curly45:
As the creator of snowmtl, thank you for your work. I still don't know if I'll ever make scraping more complicated, I don't think so because it shouldn't impact my income too much.
But the data json may change when I update the translator

vetleledaal:
> I still don't know if I'll ever make scraping more complicated, I don't think so because it shouldn't impact my income too much.

Rate limiting could be added if this becomes a problem. Some users do end up downloading 1000s of chapters at a time, for some reason, which other site operators have complained about in the past.

> But the data json may change when I update the translator

This is to be expected. You have no obligation to notify us if you make changes. If users care enough they will report it so it can be fixed on our side.

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
choppeh: ```url``` is a builder, and the extensions-lib accept ```HttpUrl``` and ```String```:  [eu/kanade/tachiyomi/network/Requests](https://github.com/mihonapp/mihon/blob/addb4ae9ad5f9294c70bce8b5eebd806115158b2/core/common/src/main/kotlin/eu/kanade/tachiyomi/network/Requests.kt#L18)

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
choppeh: it's an irregular json, there are differences in the structure of the captions

vetleledaal: Would using a custom deserializer work for you?
```kt
@Serializable
class Page(
    @SerialName("img_url")
    val imageUrl: String,
    @Serializable(with = TranslationsListSerializer::class)
    val translations: List<Translation> = emptyList()
)

@Serializable
class Translation(
    val x1: Int,
    val y1: Int,
    val x2: Int,
    val y2: Int,
    val text: String,
)

private object TranslationsListSerializer :
    JsonTransformingSerializer<List<Translation>>(ListSerializer(Translation.serializer())) {
    override fun transformDeserialize(element: JsonElement): JsonElement =
        JsonArray(element.jsonArray.map { array ->
            val coordinates = array.jsonArray[0].jsonArray
            buildJsonObject {
                put("x1", coordinates[0])
                put("y1", coordinates[1])
                put("x2", coordinates[2])
                put("y2", coordinates[3])
                put("text", array.jsonArray[1])
            }
        })
}
```

 I tested with:
```kt
val pages = Json.decodeFromString<List<Page>>(jsonString)
pages.forEach { page ->
    println("== ${page.imageUrl} ==")
    page.translations.forEach { translation ->
        println("Coords: (${translation.x1}, ${translation.y1}) to (${translation.x2}, ${translation.y2})")
        println("Text..: ${translation.text}")
    }
    println()
}
```

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
```diff
@@ -0,0 +1,480 @@
...
+
+    override val lang = "en"
+
+    override val supportsLatest = true
+
+    override val client = network.cloudflareClient.newBuilder()
+        .addInterceptor(::composedImageIntercept)
+        .build()
+
+    private val captions = mutableMapOf<String, PageMapping>()
```
AwkwardPeak7: considering this is a map of image url to translations, consider save them in the url fragments, which you can easily then access from the interceptor. (encode and decode the object to string and put it in the fragment)

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
choppeh: Changed based on suggestion [#6195 (comment)](https://github.com/keiyoushi/extensions-source/pull/6195#discussion_r1852613138)

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
choppeh: I need the dimensions (width x height) of the char to carry out the following process (I've added comments to the functions)

### Given

* Caption: Manga is Good
* Dialog Box:
```
   16px
+--------+
|        | 
|        | 
|        |
|        | 20px
|        |
|        |
+--------+

```

* Char width = 2px
* Char heigth = 3px


### When

Max char per line  : 16px / 2px  = 8 
Max number of lines: 20px / 3px    =  6

Breaking text into lines

```
  2px   2px   2px   2px   2px   2px   2px   2px  
|  M  |  a  |  n  |  g  |  a  |     |  i  |  s  | 3px
|  G  |  o  |  o  |  d  |     |     |     |     | 3px
|     |     |     |     |     |     |     |     | 3px
|     |     |     |     |     |     |     |     | 3px
|     |     |     |     |     |     |     |     | 3px
|     |     |     |     |     |     |     |     | 3px

```

### Then

Dialog box

```
   16px
+--------+
|Manga is| 
|Good    | 
|        |
|        | 20px
|        |
|        |
+--------+
```

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
choppeh: I didn't see this info in the json

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
vetleledaal: That only works for monospace fonts, otherwise it will look like:
>    16px
> +--------+
> |Manga is| 
> |Good    | 
> |        |
> |        | 20px
> |        |
> |        |
> +--------+


Someone made a `Canvas.drawMultilineText()` function that uses `StaticLayout` under the hood- https://medium.com/over-engineering/drawing-multiline-text-to-canvas-on-android-9b98f0bfa16a

https://stackoverflow.com/a/41779935 might also be a useful reference

If you can't use it directly, you could treat that function as a black box and decrease the font size until it fits the text.

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
vetleledaal: I mean, there must be some logic on the site where the font size is calculated. This shouldn't be a value you should be able to pick in the first place.

I can see something similar being useful to clamp the font size to make sure it's neither too small nor too big. It's a bit difficult to make definitive statements on current usage at a glance, but it seems like it is used as the lowest font size?

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/ComposedImageInterceptor.kt
```diff
@@ -0,0 +1,135 @@
...
+            .url(url)
+            .build()
+
+        val response = chain.proceed(imageRequest)
+
+        if (response.isSuccessful.not()) {
+            return response
+        }
+
+        val bitmap = BitmapFactory.decodeStream(response.body.byteStream())!!
```
vetleledaal, line 71: I don't expect you do do anything about it, but Suwayomi breaks here. Throwing:

https://github.com/Suwayomi/Suwayomi-Server/blob/fb51834153b33060779e93d3c460f55878a712cb/AndroidCompat/src/main/java/android/graphics/BitmapFactory.java#L21

(It will most likely break on `StaticLayout` too, so probably not even worth looking into.)

vetleledaal (dismissed):

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/ComposedImageInterceptor.kt
```diff
@@ -0,0 +1,135 @@
...
+import okhttp3.MediaType.Companion.toMediaType
+import okhttp3.Response
+import okhttp3.ResponseBody.Companion.toResponseBody
+import uy.kohesive.injekt.injectLazy
+import java.io.ByteArrayOutputStream
+import java.io.IOException
+import kotlin.math.min
+
+// The Interceptor joins the captions and pages of the manga.
+@RequiresApi(Build.VERSION_CODES.Q)
```
AwkwardPeak7: why version Q?

choppeh: the min api required is 'M' (Android 6) to use StaticLayout. 'Q' was automatically added by the IDE, I changed it to 'M'

src/en/snowmtl/src/eu/kanade/tachiyomi/extension/en/snowmtl/Snowmtl.kt
vetleledaal: This is mostly addressed by the `StaticLayout` change. It is clear that this variable is being used as the max font size.

vetleledaal (dismissed):

YatoHMK:
hello, i want to signal that the text in the tachiyomi version in comparison of the webversion doesnt fit the bubble.
Screenshot_2024-11-24-01-28-11-640_com android chrome
Screenshot_2024-11-24-01-28-24-487_xyz jmir tachiyomi mi debug

vetleledaal (approved):

PR #6241, opened 2024-11-23
Removed FlameComics Multisrc/Added individuall solution

### LINKED ISSUE #6230: flame comic back
### Source information

flame comic 1.4.34

### Source language

English 

### Steps to reproduce

flame comic is back as per their recent announcement 
Screenshot_20241123_090551
but extension is still down despite the website is on
Screenshot_20241123_090355
so hope you can fix the extension error soon .

### Expected behavior

extention showing an 404 error while website is up 
Screenshot_20241123_093358

### Actual behavior

it should be up with both extensions and webview too but it's down on extension .

### Mihon/Tachiyomi version

0.17.0

### Android version

Android 14

### Other details

Screenshot_20241123_090551
Screenshot_20241123_090551
Screenshot_20241123_093358

Hope it get fix soon .

### Acknowledgements

### 

---

Creepler13:
I removed the Multisrc since they changed their website and it seemed custom

The only problem that i have is that the version is now  lower than before, because its no longer using a Multisrc

Would close #6230

Checklist:

- [?] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [?] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions

### DIFF
--- a/src/en/flamecomics/build.gradle
+++ b/src/en/flamecomics/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Flame Comics'
     extClass = '.FlameComics'
-    extVersionCode = 5
+    extVersionCode = 35
 }
 
 apply from: "$rootDir/common.gradle"
--- a/src/en/flamecomics/build.gradle
+++ b/src/en/flamecomics/build.gradle
@@ -1,9 +1,7 @@
 ext {
     extName = 'Flame Comics'
     extClass = '.FlameComics'
-    themePkg = 'mangathemesia'
-    baseUrl = 'https://flamecomics.xyz'
-    overrideVersionCode = 4
+    extVersionCode = 5
 }
 
 apply from: "$rootDir/common.gradle"
--- a/src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
+++ b/src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
@@ -26,7 +26,6 @@ import java.io.ByteArrayOutputStream
 import java.net.URLDecoder
 
 class FlameComics : ParsedHttpSource() {
-    
     override val name = "Flame Comics"
     private val host = "flamecomics.xyz"
     override val baseUrl = "https://$host"
@@ -58,9 +57,7 @@ class FlameComics : ParsedHttpSource() {
     override fun imageUrlParse(document: Document): String = ""
 
     override fun searchMangaParse(response: Response): MangasPage {
-        val query = removeSpecialCharsregex.replace(
-            response.request.url.queryParameter("search").toString().lowercase(), "",
-        )
+        val query = removeSpecialCharsregex.replace(response.request.url.queryParameter("search").toString().lowercase(), "")
 
         var page = 1
         if (response.request.url.queryParameter("page") != null) {
--- a/src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
+++ b/src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
@@ -109,10 +109,10 @@ class FlameComics : HttpSource() {
                     setUrlWithoutDomain(
                         dataApiReqBuilder().apply {
                             val seriesID =
-                                seriesData.series_id // manga.url.toHttpUrl().pathSegments.last()
+                                seriesData.series_id
                             addPathSegment("series")
                             addPathSegment("$seriesID.json")
-                            addQueryParameter("id", seriesData.series_id.toString()) // seriesID)
+                            addQueryParameter("id", seriesData.series_id.toString())
                         }.build().toString(),
                     )
                     thumbnail_url = imageApiUrlBuilder(
--- a/src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
+++ b/src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
@@ -15,7 +15,6 @@ import eu.kanade.tachiyomi.source.model.SManga
 import eu.kanade.tachiyomi.source.online.ParsedHttpSource
 import eu.kanade.tachiyomi.util.asJsoup
 import kotlinx.serialization.decodeFromString
-import kotlinx.serialization.json.Json
 import okhttp3.Interceptor
 import okhttp3.MediaType.Companion.toMediaType
 import okhttp3.Protocol
@@ -24,7 +23,6 @@ import okhttp3.Response
 import okhttp3.ResponseBody.Companion.toResponseBody
 import org.jsoup.nodes.Document
 import org.jsoup.nodes.Element
-import uy.kohesive.injekt.injectLazy
 import java.io.ByteArrayOutputStream
 import java.net.URLDecoder
 
@@ -35,8 +33,10 @@ class FlameComics : ParsedHttpSource() {
         .addInterceptor(::composedImageIntercept)
         .build()
 
+    override val versionId: Int = 2
+
     // Flame Scans -> Flame Comics
-    override val id = 6350607071566689772
+    // override val id = 6350607071566689772
 
     override val name = "Flame Comics"
 
@@ -49,8 +49,6 @@ class FlameComics : ParsedHttpSource() {
     override val lang = "en"
 
     override val supportsLatest = true
-
-    private val json: Json by injectLazy()
 
     private fun getRealUrl(string: String?): String {
         if (string == null) return ""
@@ -67,23 +65,24 @@ class FlameComics : ParsedHttpSource() {
     override fun imageUrlParse(document: Document): String = ""
 
     override fun searchMangaParse(response: Response): MangasPage {
-        val query = response.request.url.queryParameter("search")
+        val query = removeSpecialCharsregex.replace(response.request.url.queryParameter("search").toString().lowercase(), "")
 
         var page = 1
         if (response.request.url.queryParameter("page") != null) {
-            page = Integer.parseInt(response.request.url.queryParameter("page"))
+            page = Integer.parseInt(response.request.url.queryParameter("page") + "")
         }
 
         val doc = response.asJsoup()
-        val jsonData = doc.getElementById("__NEXT_DATA__")?.data() ?: return MangasPage(listOf(), false)
-        val searchData = json.decodeFromString<SearchPageData>(jsonData)
+        val searchedSeriesData = getJsonData<SearchPageData>(doc)?.props?.pageProps?.series ?: return MangasPage(listOf(), false)
 
-        val manga = searchData.props.pageProps.series.filter { it -> removeSpecialCharsregex.replace(query.toString().lowercase(), "")  in  removeSpecialCharsregex.replace(it.title.lowercase(), "")
-        }.map { it ->
+        val manga = searchedSeriesData.filter { series ->
+            val titles = json.decodeFromString<List<String>>(series.altTitles) + series.title
+            titles.any { title -> query in removeSpecialCharsregex.replace(title.lowercase(), "") }
+        }.map { seriesData ->
             SManga.create().apply {
-                title = it.title
-                setUrlWithoutDomain("$baseUrl/series/${it.series_id}")
-                thumbnail_url = "$cdn/series/${it.series_id}/${it.cover}"
+                title = seriesData.title
+                setUrlWithoutDomain("$baseUrl/series/${seriesData.series_id}")
+                thumbnail_url = "$cdn/series/${seriesData.series_id}/${seriesData.cover}"
             }
         }
         Log.i("flamecomics", "" + manga.size)
@@ -125,29 +124,30 @@ class FlameComics : ParsedHttpSource() {
     override fun searchMangaFromElement(element: Element): SManga = popularMangaFromElement(element)
 
     override fun mangaDetailsParse(document: Document): SManga = SManga.create().apply {
-        thumbnail_url = getRealUrl(document.selectFirst("img")?.attr("abs:src"))
+        val seriesData = getJsonData<PageData>(document)?.props?.pageProps?.series
 
-        val descElement = document.selectFirst("h3:contains(Description)")
+        title = seriesData?.title + ""
 
-        description =
-            descElement?.parent()?.selectFirst("p")?.text()
+        thumbnail_url = "$cdn/series/${seriesData?.series_id}/${seriesData?.cover}"
 
-        status = if (document.selectFirst("span:contains(Ongoing)") != null) {
-            SManga.ONGOING
-        } else if (document.selectFirst("span:contains(Dropped)") != null) {
-            SManga.CANCELLED
-        } else {
-            SManga.UNKNOWN
+        description = seriesData?.description
+
+        when (seriesData?.status) {
+            "Ongoing" -> status = SManga.ONGOING
+            "Dropped" -> status = SManga.CANCELLED
+            "Hiatus" -> status = SManga.ON_HIATUS
+            "Completed" -> status = SManga.COMPLETED
+            else -> SManga.UNKNOWN
         }
 
-        author = document.selectFirst("p:contains(Author)")?.nextElementSibling()?.text()
+        author = seriesData?.author
     }
 
     override fun chapterListParse(response: Response): List<SChapter> {
         val doc = response.asJsoup()
         val jsonData = doc.getElementById("__NEXT_DATA__")?.data() ?: return listOf<SChapter>()
         val pageData = json.decodeFromString<PageData>(jsonData)
-        return pageData.props.pageProps.chapters.mapIndexed { idx, chapter ->
+        return pageData.props.pageProps.chapters.map { chapter ->
             SChapter.create().apply {
                 setUrlWithoutDomain("/series/${pageData.props.pageProps.series.series_id}/${chapter.token}")
                 date_upload = chapter.release_date * 1000
@@ -159,7 +159,7 @@ class FlameComics : ParsedHttpSource() {
     }
 
     override fun pageListParse(document: Document): List<Page> {
-        return document.select("img").mapIndexed { idx, img ->
+        return document.select("div[class$=mantine-Stack-root] img").mapIndexed { idx, img ->
             var url = img.attr("abs:src")
             url = URLDecoder.decode(url.substringAfter("url="), "utf8")
             Page(idx, imageUrl = url.substringBefore("?"))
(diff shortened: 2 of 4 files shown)


src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,166 @@ import android.graphics.Bitmap
...
+class FlameComics : ParsedHttpSource() {
 
     override val client = super.client.newBuilder()
         .rateLimit(2, 7)
         .addInterceptor(::composedImageIntercept)
         .build()
 
-    override val pageSelector = "div#readerarea img:not(noscript img)[class*=wp-image]"
+    // Flame Scans -> Flame Comics
+    override val id = 6350607071566689772
```
bapeey: if old URLs aren't compatible with the new version remove this and bump `versionId` to force the migration

src/en/flamecomics/build.gradle
```diff
@@ -1,9 +1,7 @@
 ext {
     extName = 'Flame Comics'
     extClass = '.FlameComics'
-    themePkg = 'mangathemesia'
-    baseUrl = 'https://flamecomics.xyz'
-    overrideVersionCode = 4
+    extVersionCode = 5
```
vetleledaal:
```suggestion
    extVersionCode = 35
```

This will prompt existing users to update the extension. The value is +1 from the previous version:
```
baseVersionCode + overrideVersionCode =
30 + 4 = 34
// ext-lib is 1.4, so extension version is 1.4.34
```

You could also find the current version from https://keiyoushi.github.io/extensions/

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
+    override val name = "Flame Comics"
+    private val host = "flamecomics.xyz"
+    override val baseUrl = "https://$host"
+    override val lang = "en"
+    override val supportsLatest = true
+    override val versionId: Int = 2
 
-    // Flame Scans -> Flame Comics
-    override val id = 6350607071566689772
+    private val cdn = "https:cdn.$host"
```
vetleledaal: Missing `//`

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
+    private val removeSpecialCharsregex = Regex("[^A-Za-z0-9 ]")
 
-    // Split Image Fixer Start
-    private val composedSelector: String = "#readerarea div.figure_container div.composed_figure"
+    private fun getRealUrl(string: String?): String {
+        if (string == null) return ""
+        var url = URLDecoder.decode(string.substringAfter("url="), "utf8")
+        url = url.substringBefore("?")
+        url = url.substringBefore("&")
+        return url
```
vetleledaal: `String.toHttpUrlOrNull()` could be useful here.

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
+        if (lastPage < 0) lastPage = 0
+        return MangasPage(manga.subList(page * 20, lastPage), lastPage < manga.size)
+    }
+
+    override fun searchMangaNextPageSelector(): String? = null
+    override fun popularMangaNextPageSelector(): String? = null
+    override fun latestUpdatesNextPageSelector(): String? = null
+
+    override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request =
+        GET("$baseUrl/browse?search=$query&page=$page")
```
vetleledaal: Use `String.toHttp().newBuilder()`, that way you can properly escape queries.

vetleledaal: Missing `headers` passed to `GET`.

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
+    override fun popularMangaRequest(page: Int): Request = GET(baseUrl, headers)
+    override fun latestUpdatesRequest(page: Int): Request = GET(baseUrl, headers)
+
+    override fun searchMangaSelector(): String = throw UnsupportedOperationException()
+    override fun popularMangaSelector(): String = "div[class^=SeriesCard_imageContainer] img"
+    override fun latestUpdatesSelector(): String =
+        "div[class^=SeriesCard_chapterImageContainer] img"
+
+    override fun popularMangaFromElement(element: Element): SManga = SManga.create().apply {
+        title = element.attr("abs:alt").substringAfter("$baseUrl/").replaceFirst("/", "")
```
vetleledaal: Using `abs:` for the `alt` attribute seems wrong.

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
+        val url = element.attr("abs:src")
+        thumbnail_url = getRealUrl(url).substringBefore("&")
+        setUrlWithoutDomain(getRealUrl(url).substringBefore("/thumb"))
```
vetleledaal: Use `HttpUrl`

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
+
+    override fun latestUpdatesFromElement(element: Element): SManga =
+        popularMangaFromElement(element)
+
+    override fun searchMangaFromElement(element: Element): SManga = popularMangaFromElement(element)
+
+    override fun mangaDetailsParse(document: Document): SManga = SManga.create().apply {
+        val seriesData = getJsonData<PageData>(document)?.props?.pageProps?.series
+
+        title = seriesData?.title + ""
```
vetleledaal: Should only set title if it's available.

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
+
+        thumbnail_url = "$cdn/series/${seriesData?.series_id}/${seriesData?.cover}"
```
vetleledaal: Since `seriesData` might be null you should handle this differently.

```suggestion
        seriesData?.let {
            thumbnail_url = "$cdn/series/${it.series_id}/${it.cover}"
        }
```

(although you can wrap more of the surrounding code into this `let` block, maybe use a different name than `it` in that case)

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
+
+        description = seriesData?.description
+
+        when (seriesData?.status) {
+            "Ongoing" -> status = SManga.ONGOING
+            "Dropped" -> status = SManga.CANCELLED
+            "Hiatus" -> status = SManga.ON_HIATUS
+            "Completed" -> status = SManga.COMPLETED
```
vetleledaal: Make case insensitive.

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
+                name = n
             }
+        }
+    }
+
+    override fun pageListParse(document: Document): List<Page> {
+        return document.select("div[class$=mantine-Stack-root] img").mapIndexed { idx, img ->
+            var url = img.attr("abs:src")
+            url = URLDecoder.decode(url.substringAfter("url="), "utf8")
+            Page(idx, imageUrl = url.substringBefore("?"))
```
vetleledaal: Use `HttpUrl`

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
+        val doc = response.asJsoup()
+        val jsonData = doc.getElementById("__NEXT_DATA__")?.data() ?: return listOf<SChapter>()
+        val pageData = json.decodeFromString<PageData>(jsonData)
+        return pageData.props.pageProps.chapters.map { chapter ->
+            SChapter.create().apply {
+                setUrlWithoutDomain("/series/${pageData.props.pageProps.series.series_id}/${chapter.token}")
+                date_upload = chapter.release_date * 1000
+                var n = "Chapter ${chapter.chapter.toInt()} - "
+                if (chapter.title != null) n += chapter.title
+                name = n
```
vetleledaal:
```suggestion
                name = buildString {
                    append("Chapter ${chapter.chapter.toInt()}")
                    chapter.title?.let { append(" - $it") }
                }
```

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
+        setUrlWithoutDomain(getRealUrl(url).substringBefore("/thumb"))
+    }
+
+    override fun latestUpdatesFromElement(element: Element): SManga =
+        popularMangaFromElement(element)
+
+    override fun searchMangaFromElement(element: Element): SManga = popularMangaFromElement(element)
+
+    override fun mangaDetailsParse(document: Document): SManga = SManga.create().apply {
+        val seriesData = getJsonData<PageData>(document)?.props?.pageProps?.series
```
vetleledaal: You should return early if it doesn't find `seriesData`, as all the fields rely on this being set. (or see other comment for `thumbnail_url`)

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
+            "Completed" -> status = SManga.COMPLETED
+            else -> SManga.UNKNOWN
+        }
+
+        author = seriesData?.author
+    }
+
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val doc = response.asJsoup()
+        val jsonData = doc.getElementById("__NEXT_DATA__")?.data() ?: return listOf<SChapter>()
```
vetleledaal: Can `.getJsonData()` be reused here?

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
+        val pageData = json.decodeFromString<PageData>(jsonData)
+        return pageData.props.pageProps.chapters.map { chapter ->
+            SChapter.create().apply {
+                setUrlWithoutDomain("/series/${pageData.props.pageProps.series.series_id}/${chapter.token}")
```
vetleledaal: If `token` is regularly updated it will cause entries to be marked as unread, at least when it can't detect the chapter number.

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
-class FlameComics : MangaThemesia(
-    "Flame Comics",
-    "https://flamecomics.xyz",
-    "en",
-    mangaUrlDirectory = "/series",
-) {
+class FlameComics : ParsedHttpSource() {
+    override val name = "Flame Comics"
+    private val host = "flamecomics.xyz"
+    override val baseUrl = "https://$host"
```
vetleledaal: It should be fine to repeat the `host` here and in `cdn`. Semantically you're saying that the CDN must be under the main domain, which isn't always the case.

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,158 @@ import android.graphics.Bitmap
...
-            .select(composedSelector)
-            .firstOrNull() != null
+    override fun chapterFromElement(element: Element): SChapter =
+        throw UnsupportedOperationException()
+
+    override fun chapterListSelector(): String = throw UnsupportedOperationException()
+    override fun imageUrlParse(document: Document): String = ""
+
+    override fun searchMangaParse(response: Response): MangasPage {
+        val query = removeSpecialCharsregex.replace(response.request.url.queryParameter("search").toString().lowercase(), "")
```
vetleledaal: Are you removing special characters since you're not using `HttpUrl` elsewhere, or because the site doesn't support it?

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/JsonClasses.kt
vetleledaal, line 1: Should be named something like `FlameComicsDto.kt`

vetleledaal (changes requested):

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
Creepler13: I had a problem while Migrating, because the apostrophe in the old title and in the new were different.
And its easier to search, because it will find the manga since it wont matter if you add special chars or not

Creepler13:
I also changed to HttpSource and only use the JSON data, because there are only 12 popular entries and now i can simply sort all by views

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
```diff
@@ -4,62 +4,172 @@ import android.graphics.Bitmap
...
+            "completed" -> SManga.COMPLETED
+            else -> SManga.UNKNOWN
+        }
+    }
+
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val mangaPageData = getJsonData<MangaPageData>(response.asJsoup()) ?: return listOf()
+        return mangaPageData.props.pageProps.chapters.map { chapter ->
+            SChapter.create().apply {
+                setUrlWithoutDomain(
```
vetleledaal, line 216: URL should be in format:
```kt
setUrlWithoutDomain("/series/$seriesId/$token")
```
Sending `chapterNum` is in any case a bad idea, as the remote can use it to detect keiyoushi users.

This is of course assuming that the token will be static for that chapter.

---

For the actual resource you request it would be (with `HttpUrl` for proper escaping?):
```kt
val x = "/_next/data/$buildId/series/$seriesId.json?id=$seriesId"
val x = "/_next/data/$buildId/series/$seriesId/$token.json?id=$seriesId&token=$token"
```

You don't want to use this format (`_next/...` + `getMangaUrl()` + `getChapterUrl()`) saved in in the internal database, as that would move the entry/chapter on each new `buildId`.

You can check the "Hachi" extension on how to get the `buildId` for NextJS, although I may have over-complicated it a bit.

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComicsDto.kt
```diff
@@ -0,0 +1,90 @@
...
+    @Serializable(with = keystoListSerializer::class)
+    val images: List<Page>,
+)
+
+@Serializable
+class Page(
+    val name: String,
+)
+
+class keystoListSerializer : KSerializer<List<Page>> {
```
vetleledaal: What is `keysto`?

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
vetleledaal: Regex you can use to get seriesid/token from user-facing URLs:
```kt
private val patternMangaUrl = """/series/(?<seriesId>[0-9]+)(?:/(?<token>[0-9A-Za-z]+))?""".toRegex()
```

Creepler13: The chapters appear to each have an `id`. I could try to use them since they seem more permanent than the tokens or would that have the same problem as with `chapterNum`?

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComicsDto.kt
Creepler13: It just means keys to List, because the Images are saved in a wierd format in the json

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComics.kt
vetleledaal: I don't see `chapterNum` as unique enough. You have special chapters, multiple groups, etc. For `tokens`- I think it should be safe to rely on since the public URL is shown with this token, and people might bookmark it. Meaning they should maintain compatibility with it in the future.

src/en/flamecomics/src/eu/kanade/tachiyomi/extension/en/flamecomics/FlameComicsDto.kt
vetleledaal: Then, following camelCase, you should use:
```suggestion
class keysToListSerializer : KSerializer<List<Page>> {
```

Creepler13:
I implemented the way hachi does it

vetleledaal:
The chapter URL saved in the database is wrong:
image
This will cause issues when `buildId` updates.

---

For comparison, Hachi saves the 'friendly' URL:
image

Creepler13:
This should fix it. I dont know to to look

vetleledaal:
> This should fix it. I dont know to to look

If your host app (Mihon) is a debug build, you can use 'App Inspection' in Android Studio. You can search for it with Ctrl + Shift + A.

vetleledaal (approved):

PR #6269, opened 2024-11-24
Fix InfinityScans

### LINKED ISSUE #5561: InfinityScans Theme Change
### Source information

InfinityScans 1.4.4

### Source language

English

### Steps to reproduce

1. Go to Sources
2. Select InfinityScans

### Expected behavior

1. Go to Sources
2. Select InfinityScans
3. Shows Manga lists

### Actual behavior

Not Showing Manga Lists

### Mihon/Tachiyomi version

Yokai 1.8.5.9 / Mihon 0.16.5

### Android version

Android 14

### Other details

Before the Theme Changed InfinityScans Extension, Mangas did not Show Thumbnails (return null)

### Acknowledgements

1aam2am1:
Checklist:

Closes #5561

### DIFF
--- a/src/en/infinityscans/src/eu/kanade/tachiyomi/extension/en/infinityscans/InfinityScans.kt
+++ b/src/en/infinityscans/src/eu/kanade/tachiyomi/extension/en/infinityscans/InfinityScans.kt
@@ -2,6 +2,7 @@ package eu.kanade.tachiyomi.extension.en.infinityscans
 
 import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.network.POST
+import eu.kanade.tachiyomi.network.asObservableSuccess
 import eu.kanade.tachiyomi.network.interceptor.rateLimit
 import eu.kanade.tachiyomi.source.model.Filter
 import eu.kanade.tachiyomi.source.model.FilterList
@@ -13,15 +14,12 @@ import eu.kanade.tachiyomi.source.online.HttpSource
 import eu.kanade.tachiyomi.util.asJsoup
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.decodeFromStream
-import okhttp3.FormBody
-import okhttp3.Headers
 import okhttp3.HttpUrl.Companion.toHttpUrl
-import okhttp3.MultipartBody
 import okhttp3.OkHttpClient
 import okhttp3.Request
-import okhttp3.RequestBody.Companion.toRequestBody
 import okhttp3.Response
 import org.jsoup.nodes.Element
+import rx.Observable
 import uy.kohesive.injekt.injectLazy
 
 class InfinityScans : HttpSource() {
@@ -29,7 +27,7 @@ class InfinityScans : HttpSource() {
     override val name = "InfinityScans"
 
     override val baseUrl = "https://infinityscans.net"
-    private val cdnHost = "cdn.infinityscans.xyz"
+    private val cdnHost = "cdn.infinityscans.net"
 
     override val lang = "en"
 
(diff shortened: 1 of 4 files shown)


src/en/infinityscans/src/eu/kanade/tachiyomi/extension/en/infinityscans/InfinityScans.kt
```diff
@@ -49,59 +47,114 @@ class InfinityScans : HttpSource() {
...
         add("Sec-Fetch-Mode", "cors")
         add("Sec-Fetch-Site", "same-origin")
-        add("X-Requested-With", "XMLHttpRequest")
+        add("X-requested-with", "XMLHttpRequest")
     }.build()
 
     private val json: Json by injectLazy()
 
     // Popular
+    override fun fetchPopularManga(page: Int): Observable<MangasPage> {
```
bapeey: Use the request and parse methods instead

1aam2am1: Ack,
fetchPopularManga
fetchLatestUpdates
reverted to use request and parse methods

fetchSearchManga is impossible to revert as in parse method we don't know about query or filters.

src/en/infinityscans/src/eu/kanade/tachiyomi/extension/en/infinityscans/InfinityScans.kt
```diff
@@ -49,59 +47,106 @@ class InfinityScans : HttpSource() {
         add("Sec-Fetch-Dest", "empty")
         add("Sec-Fetch-Mode", "cors")
         add("Sec-Fetch-Site", "same-origin")
-        add("X-Requested-With", "XMLHttpRequest")
+        add("X-requested-with", "XMLHttpRequest")
```
vetleledaal, line 50: Why change? Casing doesn't matter for HTTP/2 in any case

src/en/infinityscans/src/eu/kanade/tachiyomi/extension/en/infinityscans/InfinityScans.kt
```diff
@@ -49,59 +47,106 @@ class InfinityScans : HttpSource() {
...
+
+    override fun fetchSearchManga(
+        page: Int,
+        query: String,
+        filters: FilterList,
+    ): Observable<MangasPage> {
+        return client.newCall(fetchJson("api/comics"))
+            .asObservableSuccess()
+            .map { response ->
+                val data = response.parseAs<ResponseDto<SearchResultDto>>().result
```
vetleledaal, line 91: Consider extracting this scope into another function, so you don't have too deep indentation.

src/en/infinityscans/src/eu/kanade/tachiyomi/extension/en/infinityscans/WebviewInterceptor.kt
```diff
@@ -55,7 +55,7 @@ class WebviewInterceptor(private val baseUrl: String) : Interceptor {
 
             webview.webViewClient = object : WebViewClient() {
                 override fun shouldInterceptRequest(view: WebView?, request: WebResourceRequest?): WebResourceResponse? {
-                    if (request?.url.toString().contains("$baseUrl/ajax/turnstile")) {
+                    if (request?.url.toString().contains("$baseUrl/api/verification")) {
```
vetleledaal, line 58: Can you do this without converting to string? I'm assuming `contains` is used to detect the path regardless of query parameters.

```kt
request?.url.encodedPath == "/api/verification"
```

src/en/infinityscans/src/eu/kanade/tachiyomi/extension/en/infinityscans/InfinityScans.kt
1aam2am1: Hm...
I don't see any meaning to extracting this part, as it doesn't duplicate between any function and it doesn't add any clarity to what the code does.

From my point of view this function is somewhat big, and maybe need to be broken up, but it is one whole entity.

src/en/infinityscans/src/eu/kanade/tachiyomi/extension/en/infinityscans/WebviewInterceptor.kt
1aam2am1: I don't know why exactly it is checked to that way.
I only updated the requested url to match the reality.

Should I change it, as it now works.

src/en/infinityscans/src/eu/kanade/tachiyomi/extension/en/infinityscans/InfinityScans.kt
vetleledaal: This is of course subjective, but extracting it would make it [follow the parent function more closely](https://github.com/mihonapp/mihon/blob/3bd8d3ecb7023d1b01930ab0f91482c23e89c946/source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/online/HttpSource.kt#L143). It almost seem like you could remove the custom `fetch*` function here entirely. If that is possible that would be the ideal. Mostly because of the deprecation of RxJava.

src/en/infinityscans/src/eu/kanade/tachiyomi/extension/en/infinityscans/WebviewInterceptor.kt
vetleledaal: It's not critical, so leaving it as-is is also valid.

src/en/infinityscans/src/eu/kanade/tachiyomi/extension/en/infinityscans/InfinityScans.kt
vetleledaal: I just saw your other comment. In that case there's no straight forward way to do it.

> fetchSearchManga is impossible to revert as in parse method we don't know about query or filters.

vetleledaal (approved):

src/en/infinityscans/src/eu/kanade/tachiyomi/extension/en/infinityscans/InfinityScans.kt
1aam2am1: No thoughts. I didn't know http specification, and I was seeing that it sends small. So I changed it.

PR #6389, opened 2024-11-29
Nhentai | Fixed NPE in regex parser

FoxeiZ:
Checklist:

Fixed NPE cause by regex because it can't parse the `JSON.parse` properly
For example: 264032

### DIFF
--- a/src/all/nhentai/src/eu/kanade/tachiyomi/extension/all/nhentai/NHentai.kt
+++ b/src/all/nhentai/src/eu/kanade/tachiyomi/extension/all/nhentai/NHentai.kt
@@ -74,7 +74,7 @@ open class NHentai(
     }
 
     private val shortenTitleRegex = Regex("""(\[[^]]*]|[({][^)}]*[)}])""")
-    private val dataRegex = Regex("""JSON.parse\("([^*]*)"\)""")
+    private val dataRegex = Regex("""JSON\.parse\(\s*"([^"\\]*(?:\\.[^"\\]*)*)"\s*\)""")
     private val hentaiSelector = "script:containsData(JSON.parse):not(:containsData(media_server))"
     private fun String.shortenTitle() = this.replace(shortenTitleRegex, "").trim()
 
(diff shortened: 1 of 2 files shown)


src/all/nhentai/src/eu/kanade/tachiyomi/extension/all/nhentai/NHentai.kt
```diff
@@ -74,7 +74,7 @@ open class NHentai(
     }
 
     private val shortenTitleRegex = Regex("""(\[[^]]*]|[({][^)}]*[)}])""")
-    private val dataRegex = Regex("""JSON.parse\("([^*]*)"\)""")
+    private val dataRegex = Regex("""JSON\.parse\(\s*"([^"\\]*(?:\\.[^"\\]*)*)"\s*\)""")
```
vetleledaal: It's usually a bad idea to parse escape sequences with regex. Consider using something more greedy, like:
```kt
 private val dataRegex = Regex("""JSON\.parse\(\s*"(.*)"\s*\)""")
```

This completes in 26 steps, as opposed to ~2200 for some arbitrary entry.

FoxeiZ: Yeah, that's made more sense. Thanks for the suggestion

src/all/nhentai/src/eu/kanade/tachiyomi/extension/all/nhentai/NHentai.kt
```diff
@@ -241,8 +242,10 @@ open class NHentai(
     override fun chapterListParse(response: Response): List<SChapter> {
         val document = response.asJsoup()
         val script = document.selectFirst(hentaiSelector)!!.data()
+        Log.e("chapterListParse.script", "script: ${script.isEmpty()}")
 
         val json = dataRegex.find(script)?.groupValues!![1]
+        Log.e("chapterListParse.json", "json: ${json.isEmpty()}")
```
vetleledaal: Remove debugging remnants.

vetleledaal (approved):

PR #6399, opened 2024-11-30
Comick: Better first cover selection

BrutuZ:
- Use `startsWith` to better handle compound languages
- Additional fallback to a cover on the entry's original language if none a re found in the extension's selected language. Handles weird edge-cases like the example below where the Ukranian cover would be picked for Japanese content regardless of the selected language since it's the first on the list.
image

Checklist:

### DIFF
--- a/src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
+++ b/src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
@@ -396,23 +396,16 @@ abstract class Comick(
             val coversUrl =
                 "$apiUrl/comic/${mangaData.comic.slug ?: mangaData.comic.hid}/covers?tachiyomi=true"
             val covers = client.newCall(GET(coversUrl)).execute()
-                .parseAs<Covers>().mdCovers.reversed().toMutableList()
-            if (covers.any { it.vol == "1" }) covers.retainAll { it.vol == "1" }
-            if (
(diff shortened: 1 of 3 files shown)


src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
```diff
@@ -399,9 +399,15 @@ abstract class Comick(
...
             if (
-                covers.any { it.locale == comickLang.split('-').first() }
+                covers.any { comickLang.startsWith(it.locale.orEmpty()) }
             ) {
-                covers.retainAll { it.locale == comickLang.split('-').first() }
+                covers.retainAll { comickLang.startsWith(it.locale.orEmpty()) }
+            } else if (
+                covers.any { mangaData.comic.isoLang.orEmpty().startsWith(it.locale.orEmpty()) }
+            ) {
+                covers.retainAll {
```
vetleledaal: Might be better to use immutable list + when.

BrutuZ: I couldn't figure out how a `when` would fit in this situation, but I wasn't happy with all the duplicated code so I split each into its own filter.

vetleledaal: Nice, the new solution seems overall better than what you would achieve with `when`.

vetleledaal (approved):

PR #6405, opened 2024-11-30
Add Manhwalike

### LINKED ISSUE #6037: manhwalike
### Source name

manhwalike

### Source language

English

Would you add this site to new repo version? 

https://manhwalike.com/ 

### Other details

_No response_

### Acknowledgements

TeflonOfJoy:
Closes #6037

Checklist:


src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,223 @@
...
+    override fun popularMangaNextPageSelector() = null
+
+    override fun popularMangaFromElement(element: Element): SManga {
+        return SManga.create().apply {
+            title = element.selectFirst("h3.title a")!!.text()
+            setUrlWithoutDomain(element.selectFirst("a")!!.attr("href"))
+            thumbnail_url = when {
+                element.selectFirst("img")!!.hasAttr("data-original") -> element.selectFirst("img")!!.attr("data-original")
+                else -> element.selectFirst("img")!!.attr("src")
+            }
```
vetleledaal: Do not use non-null assert for thumbnail URL. It should not hard-fail if it can't find a cover.

There are other places where you should remove non-null asserts, author, date, description, genre, status should all be optional.

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,223 @@
...
+    override fun getFilterList() = FilterList(
+        Filter.Header("NOTE: Ignored if using text search!"),
+        Filter.Separator(),
+        GenreFilter(),
+    )
+
+    private class GenreFilter : UriPartFilter(
+        "Genre",
+        arrayOf(
+            Pair("Action", "manga-genre-action"),
```
vetleledaal, line 155: Generally you should also include an 'all' genre. You should also add a comment on how to update this list, for example:
```js
copy($$("#glo_gnb .sub-menu a").map(el => `Pair("${el.innerText.trim()}", "${el.pathname.substr(1)}"),`).join("\n")) 
```

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/ManhwalikeHelper.kt
```diff
@@ -0,0 +1,39 @@
...
+        return when {
+            this == null -> SManga.UNKNOWN
+            this.contains("Ongoing", true) -> SManga.ONGOING
+            this.contains("Finish", true) -> SManga.COMPLETED
+            else -> SManga.UNKNOWN
+        }
+    }
+
+    fun String?.toDate(): Long {
+        val dateFormat = SimpleDateFormat("MMMM d, yyyy", Locale.ENGLISH)
```
vetleledaal: Looks like the timezone is UTC-5, most likely EST. You can find the timezone in the sitemap (`/sitemap.xml`). I have cross-referenced this with the relative timestamps, which appears to use the same offset.

You can set it with for example:
```kt
val dateFormat = SimpleDateFormat(...).apply {
    timeZone = Timezone.getTimeZone("Europe/New_York")
}

File: src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt (hunk 1 above)
Comment (bapeey) [inline src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt]: Consider extracting this to a function
```kt
private fun Element.imgAttr(): String = when {
    hasAttr("data-original") -> absUrl("data-original")
    else -> absUrl("src")
}
```
```kt
thumbail_url = element.selectFirst("img")?.imgAttr()
```

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,223 @@
...
+
+    override fun latestUpdatesSelector() = "ul.slick_item div.visual"
+
+    override fun latestUpdatesNextPageSelector() = popularMangaNextPageSelector()
+
+    override fun latestUpdatesFromElement(element: Element): SManga = popularMangaFromElement(element)
+
+    // search
+    override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
+        val filterList = filters.let { if (it.isEmpty()) getFilterList() else it }
```
bapeey: not needed, you can use `filters.forEach` directly

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,223 @@
...
+            }
+            setUrlWithoutDomain(element.selectFirst("a")!!.attr("href"))
+        }
+    }
+
+    // details
+    override fun mangaDetailsParse(document: Document): SManga {
+        return SManga.create().apply {
+            author = document.selectFirst("div.author a")!!.text()
+            status = document.selectFirst("small:contains(Status) + strong")!!.text().trim().toStatus()
```
bapeey: Non-critical fields
```suggestion
            author = document.selectFirst("div.author a")?.text()
            status = document.selectFirst("small:contains(Status) + strong")?.text()?.trim()?.toStatus()
```

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,223 @@
...
+            genre = document.select("div.categories a").joinToString { it.text() }
+            description = document.select("div.summary-block p.about").text()
+            thumbnail_url = document.select("div.fixed-img img").attr("src")
```
bapeey:
```suggestion
            thumbnail_url = document.select("div.fixed-img img").attr("abs:src")
```

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,223 @@
...
+            setUrlWithoutDomain(element.select("a").attr("href"))
+            name = element.select("a").text()
+            date_upload = element.select(".time").text().toDate()
+        }
+    }
+
+    // pages
+    override fun pageListParse(document: Document): List<Page> {
+        return document.select(".chapter-content .page-chapter img").mapIndexed { i, it ->
+            Page(i, imageUrl = it.attr("src"))
```
bapeey:
```suggestion
            Page(i, imageUrl = it.attr("abs:src"))
```

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/ManhwalikeHelper.kt
```diff
@@ -0,0 +1,39 @@
...
+    fun String?.toStatus(): Int {
+        return when {
+            this == null -> SManga.UNKNOWN
+            this.contains("Ongoing", true) -> SManga.ONGOING
+            this.contains("Finish", true) -> SManga.COMPLETED
+            else -> SManga.UNKNOWN
+        }
+    }
+
+    fun String?.toDate(): Long {
```
bapeey, line 37: Extract `dateFormat` to avoid creating a new instance on every call.

Wrap `dateFormat.parse(...)` in a try-catch block that returns 0, the app will use the current date in that case, so there's no need to use `Calendar`

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
vetleledaal: Should use `.absUrl()`, but that only works for `Element` not `Elements`. You can get `Element` with `.selectFirst()` (or by iteration, not relevant here)

vetleledaal (changes requested):
Comments were pending. Might overlap with what @bapeey mentioned elsewhere

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
vetleledaal: I would say `.absUrl()` is preferred.

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
TeflonOfJoy: The reason I haven't included the 'All' genre in the list is that there isn't a proper one on the site, the 'All' genre is just a redirect to the front page:

`<a class="navbar-menu-item cathot" href="/">All</a>`

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,217 @@
...
+            title = element.selectFirst("img")!!.attr("alt")
+            thumbnail_url = element.selectFirst("img")?.toOriginal()
+            setUrlWithoutDomain(element.selectFirst("a")!!.attr("href"))
+        }
+    }
+
+    // details
+    override fun mangaDetailsParse(document: Document): SManga {
+        return SManga.create().apply {
+            author = document.selectFirst("div.author a")!!.text()
```
vetleledaal: Should be optional

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,217 @@
...
+        val mangas = when {
+            document.select(searchMangaSelector()).isEmpty() -> document.select("ul li").map { element ->
+                searchMangaFromElement(element)
+            }
+            else -> document.select(searchMangaSelector()).map { element ->
+                searchMangaFromElement(element)
+            }
+        }
+        val hasNextPage = searchMangaNextPageSelector().let { selector ->
+            document.select(selector).first()
```
vetleledaal: `.selectFirst()`

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,217 @@
...
+        return SChapter.create().apply {
+            setUrlWithoutDomain(element.selectFirst("a")!!.absUrl("href"))
+            name = element.selectFirst("a")!!.text()
+            date_upload = element.selectFirst(".time")!!.text().toDate()
+        }
+    }
+
+    // pages
+    override fun pageListParse(document: Document): List<Page> {
+        return document.select(".chapter-content .page-chapter img").mapIndexed { i, it ->
```
vetleledaal: Don't use `it`. The most common are `i/idx/index` and `img/image`. I would say any combination of that is valid, but I personally like `i, img`

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,217 @@
...
+            Page(i, imageUrl = it.attr("src"))
```
vetleledaal: Missing absUrl

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,217 @@
...
+        }
+    }
+
+    // chapters
+    override fun chapterListSelector() = "ul.chapter-list li"
+    override fun chapterFromElement(element: Element): SChapter {
+        return SChapter.create().apply {
+            setUrlWithoutDomain(element.selectFirst("a")!!.absUrl("href"))
+            name = element.selectFirst("a")!!.text()
+            date_upload = element.selectFirst(".time")!!.text().toDate()
```
vetleledaal: Should be optional

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,217 @@
...
+    override fun imageUrlParse(document: Document): String = throw UnsupportedOperationException()
+
+    // filters
+    override fun getFilterList() = FilterList(
+        Filter.Header("NOTE: Ignored if using text search!"),
+        Filter.Separator(),
+        GenreFilter(),
+    )
+
+    private class GenreFilter : UriPartFilter(    // the list can be updated via copy($$("#glo_gnb .sub-menu a").map(el => `Pair("${el.innerText.trim()}", "${el.pathname.substr(1)}"),`).join("\n"))
```
vetleledaal: Failing on lint

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,217 @@
...
+            thumbnail_url = element.selectFirst("img")?.toOriginal()
+            setUrlWithoutDomain(element.selectFirst("a")!!.attr("href"))
+        }
+    }
+
+    // details
+    override fun mangaDetailsParse(document: Document): SManga {
+        return SManga.create().apply {
+            author = document.selectFirst("div.author a")!!.text()
+            status = document.selectFirst("small:contains(Status) + strong")?.text()?.trim().toStatus()
```
vetleledaal: `.text()` is trimmed already, according to Jsoup docs.

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,217 @@
...
+            document.select(selector).first()
+        } != null
+        return MangasPage(mangas, hasNextPage)
+    }
+
+    override fun searchMangaFromElement(element: Element): SManga {
+        return SManga.create().apply {
+            title = element.selectFirst("img")!!.attr("alt")
+            thumbnail_url = element.selectFirst("img")?.toOriginal()
+            setUrlWithoutDomain(element.selectFirst("a")!!.attr("href"))
```
vetleledaal: use absUrl

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    override fun popularMangaSelector() = "ul.list-hot div.visual"
+
+    override fun popularMangaNextPageSelector() = null
+
+    override fun popularMangaFromElement(element: Element): SManga {
+        return SManga.create().apply {
+            title = element.selectFirst("h3.title a")!!.text()
+            setUrlWithoutDomain(element.selectFirst("a")!!.attr("href"))
```
vetleledaal: Use absUrl

vetleledaal (changes requested):

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,218 @@
...
+        return GET(baseUrl, headers)
+    }
+
+    override fun popularMangaSelector() = "ul.list-hot div.visual"
+
+    override fun popularMangaNextPageSelector() = null
+
+    override fun popularMangaFromElement(element: Element): SManga {
+        return SManga.create().apply {
+            title = element.selectFirst("h3.title a")?.text().toString()
```
vetleledaal: This will set the title to an empty string if the title is not found, equivalent to this, but much harder to see:
```kt
title = element.selectFirst("h3.title a")?.text() ?: ""
```

The correct thing to do if title is not found is to error out (you can use non-null assert) or just not set the variable.

So, what I'm suggesting is either of:
```kt
title = element.selectFirst("h3.title a")?.text()!!
```

```kt
element.selectFirst("h3.title a")?.text()?.also { title = it }
```

vetleledaal: Since you are using the latter pattern for URL, maybe do the same here?

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,218 @@
...
+            element.selectFirst("a")?.absUrl("href")?.let { setUrlWithoutDomain(it) }
```
vetleledaal: Since the lambda block is not modifying anything in the chain, you should use `.also`, which is intended for side effects.

Comment applies to every place you use `.let` (except for `hasNextPage`)

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,218 @@
...
+        }
+        val hasNextPage = searchMangaNextPageSelector().let { selector ->
+            document.selectFirst(selector)
+        } != null
+        return MangasPage(mangas, hasNextPage)
+    }
+
+    override fun searchMangaFromElement(element: Element): SManga {
+        return SManga.create().apply {
+            title = element.selectFirst("img")?.attr("alt").toString()
```
vetleledaal: Hides `?: ""` thing.

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
```diff
@@ -0,0 +1,218 @@
...
+            thumbnail_url = document.selectFirst("div.fixed-img img")?.absUrl("src")
+        }
+    }
+
+    // chapters
+    override fun chapterListSelector() = "ul.chapter-list li"
+    override fun chapterFromElement(element: Element): SChapter {
+        return SChapter.create().apply {
+            element.selectFirst("a")?.absUrl("href")?.let { setUrlWithoutDomain(it) }
+            name = element.selectFirst("a")?.text().toString()
```
vetleledaal: Hides `?: ""` thing.

vetleledaal (dismissed):

src/en/manhwalike/src/eu/kanade/tachiyomi/extension/en/manhwalike/Manhwalike.kt
TeflonOfJoy: Alright, thank you for pointing it out

vetleledaal (approved):

PR #6408, opened 2024-12-01
Fix HentaiRead

### LINKED ISSUE #6386: Hentairead chapters not downloading
### Source information

Hentairead

### Source language

English 

### Steps to reproduce

When I tried to dl any chapters it says 

" Unexpected JSON token at offset 2:

Expected start of the object '{', but had '/' instead at path: $[0]

JSON input: [ /* <CDATA[ */

var chapterExt..... "

### Expected behavior

Is supposed be normal to download any chapters.

### Actual behavior

It giving errors and I think its almost a month now.

![Screenshot_2024-11-29-19-12-13-28_afd66b7916d19a60815f56273408b73e

### Mihon/Tachiyomi version

0.15.3

### Android version

Android 11

### Other details

_No response_

### Acknowledgements

### 

---

Creepler13:
Closes #6386

Checklist:

### DIFF
--- a/src/en/hentairead/src/eu/kanade/tachiyomi/extension/en/hentairead/Hentairead.kt
+++ b/src/en/hentairead/src/eu/kanade/tachiyomi/extension/en/hentairead/Hentairead.kt
@@ -185,8 +185,11 @@ class Hentairead : Madara("HentaiRead", "https://hentairead.com", "en", dateForm
         }
     }
 
-    private val chapterExtraDataRegex = Regex("= (\\{[^;]+)")
-    private val pagesDataRegex = Regex(".(ey\\S+).\\s")
+    // chapterExtraData = ({...});
+    private val chapterExtraDataRegex = Regex("""= (\{[^;]+)""")
+
+    // window.mMjM5MjM2 = '(eyJkYX...);
+    private val pagesDataRegex = Regex(""".(ey\S+).\s""")
 
     // From ManhwaHentai - modified
     override fun pageListParse(document: Document): List<Page> {
--- a/src/en/hentairead/src/eu/kanade/tachiyomi/extension/en/hentairead/Hentairead.kt
+++ b/src/en/hentairead/src/eu/kanade/tachiyomi/extension/en/hentairead/Hentairead.kt
@@ -1,5 +1,6 @@
 package eu.kanade.tachiyomi.extension.en.hentairead
 
+import android.util.Base64
 import eu.kanade.tachiyomi.multisrc.madara.Madara
 import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.source.model.FilterList
@@ -8,7 +9,6 @@ import eu.kanade.tachiyomi.source.model.Page
 import eu.kanade.tachiyomi.source.model.SChapter
 import eu.kanade.tachiyomi.source.model.SManga
 import eu.kanade.tachiyomi.source.model.UpdateStrategy
-import kotlinx.serialization.Serializable
 import kotlinx.serialization.decodeFromString
 import okhttp3.HttpUrl.Companion.toHttpUrl
 import okhttp3.Request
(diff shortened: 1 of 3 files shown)


src/en/hentairead/src/eu/kanade/tachiyomi/extension/en/hentairead/Hentairead.kt
```diff
@@ -189,14 +189,18 @@ class Hentairead : Madara("HentaiRead", "https://hentairead.com", "en", dateForm
     override fun pageListParse(document: Document): List<Page> {
         launchIO { countViews(document) }
 
-        val pages = document.selectFirst("[id=single-chapter-js-extra]")?.data()
-            ?.substringAfter(":[")
-            ?.substringBefore("],")
-            ?.let { json.decodeFromString<List<PageDto>>("[$it]") }
+        val pageBaseUrl = document.selectFirst("[id=single-chapter-js-extra]")?.data()
+            ?.substringAfter("chapterExtraData = ")
+            ?.substringBefore(";")?.let { json.decodeFromString<ImageBaseUrlDto>(it).baseUrl }
```
vetleledaal: Use a regex for this. It is more flexible.

Current implementation is equivalent to:
```regex
chapterExtraData = (.*?);
```
but you probably want:
```regex
\bchapterExtraData\s*=\s*(.*);
```

src/en/hentairead/src/eu/kanade/tachiyomi/extension/en/hentairead/Hentairead.kt
```diff
@@ -189,14 +189,18 @@ class Hentairead : Madara("HentaiRead", "https://hentairead.com", "en", dateForm
...
+
+        val pages = document.selectFirst("[id=single-chapter-js-before]")?.data()
+            ?.substringAfter("= '")
+            ?.substringBefore("'")
```
vetleledaal: Also use regex here.

current is equivalent to:
```regex
= '(.*?)'
```

but you probably want:
```regex
=\s*'(.*)'
```

src/en/hentairead/src/eu/kanade/tachiyomi/extension/en/hentairead/Hentairead.kt
```diff
@@ -189,14 +189,18 @@ class Hentairead : Madara("HentaiRead", "https://hentairead.com", "en", dateForm
...
+            ?.let { json.decodeFromString<PagesDto>(String(Base64.decode(it, Base64.DEFAULT))) }
             ?: throw Exception("Failed to find page list. Non-English entries are not supported.")
 
-        return pages.mapIndexed { idx, page ->
-            Page(idx, document.location(), page.src)
+        return pages.data.chapter.images.mapIndexed { idx, page ->
+            Page(idx, document.location(), "$pageBaseUrl/${page.src}")
```
vetleledaal, line 208: Does this escape URLs properly?

Assuming the least, it would break in this scenario:
- `page.src` is a filename
- the filename is `example#filename.png`
- it will try to load `$pageBaseUrl/example` instead of `$pageBaseUrl/example%23filename.png`

vetleledaal (changes requested):

src/en/hentairead/src/eu/kanade/tachiyomi/extension/en/hentairead/Hentairead.kt
Creepler13: I think its fine the pages are saved like this 
```json
[{"src":"239236\/66227\/hr_0001.jpg","width":1280,"height":1810}...n]
```
and the baseUrl like this
```json
{ "baseUrl": "https:\/\/henread.xyz" }
```

vetleledaal: Thanks for clarifying. If this was escaped, it would turn `/` into `%2F`, which would be wrong.

src/en/hentairead/src/eu/kanade/tachiyomi/extension/en/hentairead/Hentairead.kt
```diff
@@ -185,17 +185,19 @@ class Hentairead : Madara("HentaiRead", "https://hentairead.com", "en", dateForm
         }
     }
 
+    private val chapterExtraDataRegex = Regex("= (\\{[^;]+)")
+    private val pagesDataRegex = Regex(".(ey\\S+).\\s")
```
vetleledaal: Use `"""` to  avoid escaping inside the regex. Since these are sufficiently complex you should show a cut down example of the intended target text.

AwkwardPeak7 (approved):

PR #6471, opened 2024-12-05
mangadex: add option to use source language for title

### LINKED ISSUE #5747: MangaDex: Allow Titles to Be in Desired Language
### Source name

MangaDex

### Source language

Multiple, but English in my case

Manga titles can come in several languages. When it comes to MangaDex, the default title appears to be the transliterated one when available (sometimes it can be pure English, although rarer). Sometimes when refreshing a manga that originally did not have a transliterated title, the title will switch to the transliterated one, causing headache for myself and potentially others when wanting to find the manga in question later down the line.

Tl;DR: My request is to add an option to the extension settings for a preferred language for Manga titles.

Example Manga: https://mangadex.org/title/91503d4c-16d6-4962-9eca-93557667066c/naze-boku-no-sekai-wo-daremo-oboeteinai-no-ka
Title in app should change from `Naze Boku no Sekai wo Daremo Oboeteinai no ka?` to `Why Does Nobody Remember Me in This World?`

### Other details

I have tried with the latest TachiyomiJ2k and Yokai, and I can confirm that this option is not present with either of them.

### Acknowledgements

Kladki:
Closes #5747

I am not too confident about the variable names and English strings. So if someone could look over them and see if they are good, that would be nice.

Some titles still appear in their original title with this setting enabled. This is because MangaDex incorrectly claims that the title is actually in English. There isn't much that can be done about that on our side.
For example, [My Dress-Up Darling](https://mangadex.org/title/aa6c76f7-5f5f-46b6-a800-911145f81b9b/sono-bisque-doll-wa-koi-o-suru) states that the title "Sono Bisque Doll wa Koi o Suru" is English in the following API call: `https://api.mangadex.org/manga/aa6c76f7-5f5f-46b6-a800-911145f81b9b?includes%5B%5D=cover_art&includes%5B%5D=author&includes%5B%5D=artist`
```json5
{
    "data": {
        "attributes": {
            "altTitles": [
                // -- snip --
                // This is the title most people would say is correct, but afaik there is no systematic way to figure that out
                {
                    "en": "My Dress-Up Darling"
                },
                {
                    "en": "The Bisque Doll Is Falling in Love"
                },
                {
                    "en": "The Bisque Doll Falls in Love"
                }
            ],
            // -- snip --
            "title": {
               // This is clearly not English
                "en": "Sono Bisque Doll wa Koi o Suru"
            },
            "updatedAt": "2024-11-20T12:11:31+00:00",
            "version": 278,
            "year": 2018
        },
       // -- snip --
    },
}
```

Checklist:


src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDexHelper.kt
```diff
@@ -283,14 +283,26 @@ class MangaDexHelper(lang: String) {
...
+                    ?: titleMap.values.firstOrNull() // fallback to the original language title
+                    ?: mangaDataDto.attributes.altTitles
+                        .find { it["en"] !== null } // if all else fails, try to find an english title
+                        ?.values?.singleOrNull()
+            } else {
+                titleMap.values.firstOrNull() // use literally anything from title as first resort
+                    ?: mangaDataDto.attributes.altTitles
+                        .find { (it[lang] ?: it["en"]) !== null }
+                        ?.values?.singleOrNull() // find something else from alt titles
+            }
```
vetleledaal, line 305: Consider:
```kt
private fun List<Map<String, String>>.findTitleByLang(lang: String): String? =
    firstOrNull { it[lang] != null }?.values?.singleOrNull()
```

```kt
val dirtyTitle = with(mangaDataDto.attributes) {
    titleMap[lang] ?: altTitles.run {
        val mainTitle = titleMap.values.firstOrNull()
        val langTitle = findTitleByLang(lang)
        val enTitle = findTitleByLang("en")

        if (preferSourceLangTitle) {
            listOf(langTitle, mainTitle, enTitle).firstNotNullOfOrNull { it }
        } else {
            listOf(mainTitle, langTitle, enTitle).firstNotNullOfOrNull { it }
        }
    }
}
```
    
or

```kt
val altTitles = mangaDataDto.attributes.altTitles
val dirtyTitle = titleMap[lang] ?: if (preferSourceLangTitle) {
    listOf(
        altTitles.findTitleByLang(lang),
        titleMap.values.firstOrNull(),
        altTitles.findTitleByLang("en"),
    ).firstNotNullOfOrNull { it }
} else {
    listOf(
        titleMap.values.firstOrNull(),
        altTitles.findTitleByLang(lang),
        altTitles.findTitleByLang("en"),
    ).firstNotNullOfOrNull { it }
}
```

(needs testing)

src/all/mangadex/assets/i18n/messages_en.properties
```diff
@@ -76,6 +76,8 @@ original_language=Original language
 original_language_filter_chinese=%s (Manhua)
 original_language_filter_japanese=%s (Manga)
 original_language_filter_korean=%s (Manhwa)
+prefer_title_in_source_language=Prefer title in source language
+prefer_title_in_source_language_summary=If there is an alternate title available which matches the language set for the source, it will be used
```
vetleledaal: It is not clear what "set for the source" means.

vetleledaal (dismissed):

src/all/mangadex/assets/i18n/messages_en.properties
Kladki: Basically what is in the screenshot, not sure how else to explain it.
Screenshot_2024-12-07-01-04-18-439-edit_app mihon

vetleledaal: While 'source' is used internally, I think something else should be used for user-facing strings. Currently it can be confused for the origin language (manga=Japanese, manhwa=Korean). A bit awkward, but I think calling it the extension language would make it unambiguous.

So:
```ini
prefer_title_in_source_language=Prefer extension language
prefer_title_in_source_language_summary=If there is an alternate title available which matches the extension language, it will be used
```

I think the use of the comma could be better, but I don't have a better suggestion.

Kladki: Thanks! Should I change the variables to match (e.g. `prefer_title_in_extension_language`) or are you indifferent?

vetleledaal: It would be better if it matches. If not it will be unclear why 'source' was used when it's not in the string itself.

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDexHelper.kt
Kladki: Now I think about it, I don't see why there is even a fallback to English alt titles, since I would assume there would always be some sort of "official" title? Unfortunately, `git blame` doesn't give any info about who wrote it and why due to keiyoushi bot being the only committer. Should we keep the final fallback just in case, or is it pointless and should be removed?

vetleledaal: The only time it will fall back to using English is if `altTitles` doesn't have the chosen extension language and `titles` has nothing. This is an existing assumption(?), and if both those are missing it is then assuming English will be available in `altTitles`.

Rather than removing it, I think adding another fallback would be preferable. So, for example:
```kt
titleMap.values.firstOrNull(),
altTitles.findTitleByLang(lang),
altTitles.findTitleByLang("en"),
altTitles.firstOrNull(),
```

This will only fail if both `title` and `altTitles` is missing, but at that point, there must be something seriously wrong.

You can check the old history here: https://github.com/funkyhippo/tachiyomi-extensions/blame/1c8478fcca73ce1c8d36bb6be021253b4be1ccbe/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDexHelper.kt

Kladki: Looking at the history, I pinned it to [this](https://github.com/funkyhippo/tachiyomi-extensions/commit/f92453cdcf7fc514a90d2b2f0f5409bab01895fc) commit. However, the description of the commit doesn't really make sense:
```
Some manga somehow has romaji title tagged as "jp" language. So also use
altTitles list and use extension language as the base language when
searching while keeping english as fallback.
```
The first sentence doesn't really justify the second, especially when you consider that before that commit, the extension only ever tried to read English languages. @ivaniskandar do you remember why you did this, since I have never seen a manga on MangaDex which doesn't have any titles.

Kladki: > `titles` has nothing

That's kind of my point. I am asking whether that is even possible? Cause if not, no other fallbacks are needed,

Kladki: I implemented your first changes, but I'll leave the fallbacks as-is for now.

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
```diff
@@ -757,11 +758,27 @@ abstract class MangaDex(final override val lang: String, private val dexLang: St
             }
         }
 
+        val preferExtensionLangTitlePref = SwitchPreferenceCompat(screen.context).apply {
+            key = MDConstants.getPreferExtensionLangTitlePrefKey(dexLang)
+            title = helper.intl["prefer_title_in_extension_language"]
+            summary = helper.intl["prefer_title_in_extension_language_summary"]
+            setDefaultValue(true)
```
vetleledaal, line 765: It makes sense to enable for new users, but not at the cost of invalidating downloads for existing users. Do you think detecting this would be doable? Otherwise the default should be false.

Kladki: From my testing, items in your library cannot change title. The only way to do so is to remove it, refresh, then add to library again, so this shouldn't be an issue.

AwkwardPeak7 (approved):

PR #6598, opened 2024-12-13
[MangaThemesia] Option to hide paid chapters

### LINKED ISSUE #5908: Thunder Scans: Premium/Paid Chapters Auto detect
### Source name

Thunder Scans

### Source language

English

Currently, early access chapter do not appear in the app as a paid/premium chapter.

If there was a toggle in settings so they would show up as a paid/premium chapter or a way it could auto detect that these are paid/premium chapters, and don't list them till they are free to read, would be great.
The other option would be to have us logging in with our accounts to access the Premium/paid chapters for those who already have a premium account.

### Other details

_No response_

### Acknowledgements

lord-ne:
Two of the sources using the `MangaThemesiaAlt` theme, NightScans and ThunderScans, have paid chapters which cannot be viewed in Mihon without purchasing (attempting to click them results in an error). This PR hides those chapters from the chapter list. It also adds a setting to go back to the old behavior of not hiding them (a user might want to use this, for example, to be notified when there are new paid chapters so that they can go and purchase them on the scanlator's website), but the default setting is to hide them.

Closes #5908

Checklist:

- [x] N/A Updated `extVersionCode` value in `build.gradle` for individual extensions
  - Updated `overrideVersionCode` for extensions that we added functionality to. The `baseVersionCode` does not need to be updated since we only added a new file to the multisource library, we didn't do anything that would affect existing extensions using the library.
- [x] N/A Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] N/A Have explicitly kept the `id` if a source's name or language were changed
- [x] N/A Have removed `web_hi_res_512.png` when adding a new extension

The new setting option:

Screenshot from 2024-12-13 12-41-48

With hiding enabled:

Screenshot from 2024-12-13 12-41-18

With hiding disabled:

Screenshot from 2024-12-13 12-42-13

(Also confirmed that after logging in with web view and purchasing chapters, they do show up even if hiding is enabled)


AwkwardPeak7:
I don't think this needs to be in `MangaThemesiaAlt`, maybe in the base class, but imo just update selector in affected sources.
afaik lock system isn't part of the base wordpress theme, it comes with some addon, so different sources can have different selectors

lord-ne:
Two out of the three sources using the `MangaThemesiaAlt` base class (`NightScans` and `ThunderScans`) have this lock system, whereas none of the ones I checked using the regular `MangaThemesia` base class had it (I checked like 20), so I assumed it was related to the theme. Do you know what addon it comes from? 

To be honest I still don't fully understand what `MangaThemesiaAlt` is, like why is there an Alt version? Is there a second version of the Manga theme on Themesia or something? We should really have documentation

AwkwardPeak7:
Alt version have additional logic for dynamic urls (random part in manga url which changes every so often)

lord-ne:
I see a few ways forward here, so I'm writing them out here to get your thoughts:

1. Manually implement this for Night Scans and Thunder Scans.
  a. This is simple, but we'd be duplicating a lot of code. It would get pretty annoying if any more sources needed this, we'd be copying the code everywhere.
2. Create some helper methods in the `mangathemesia` lib to help implement this
  a. Basically all the sources would have to do is call the helper methods inside their `setupPreferenceScreen` and `chapterListSelector` functions. But I'm not sure if creating helper methods like that is really something we do in this codebase.
3. Keep the functionality in the `MangaThemesiaAlt` class
  a. This is pretty simple. We should probably have a boolean to enable/disable it even if we do keep it here; we don't want the preference showing up for sources that actually don't need it. 
4. Put the functionality in the `MangaThemesia` class
  a. This makes more sense conceptually than (3), but if it's tricky if we won't want the preference option to show up for sources that don't support it (which is most of them, so we probably do want to ensure it doesn't show up) because `MangaThemesia` doesn't currently have any preferences or inherit from `ConfigurableSource`, and we wouldn't want to have a blank preference screen show up.

I'm leaning towards (2), I'll throw together an implementation and see how it looks.

lord-ne:
Updated and bench tested. How does this implementation look @AwkwardPeak7 ?

AwkwardPeak7:
honestly just change selector in those sources to block paid chapters without any preference to toggle it.... that is what we usually do for most of such sources and only add preference if users ask for it.

lord-ne:
I've actually been finding the toggle useful, some of these sites have decent coin rates where it's only like 10 cents per chapter, so on some of them I actually do want to leave the toggle off and buy the chapters.

src/all/thunderscans/src/eu/kanade/tachiyomi/extension/all/thunderscans/ThunderScansFactory.kt
AwkwardPeak7, line 1: make a ThunderScan class here with the changes then inherit both from this class

AwkwardPeak7 (changes requested):

AwkwardPeak7 (approved):

PR #6657, opened 2024-12-17
Asura Scans: support high quality chapter images

vetleledaal:
Might be patched in the future, so disabling it by default.

Have not fully considered how caching will be affected by this. User might expect chapters in the cache to be invalidated, but if that is not a concern it should be fine.

Checklist:

### DIFF
--- a/src/en/asurascans/build.gradle
+++ b/src/en/asurascans/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Asura Scans'
     extClass = '.AsuraScans'
-    extVersionCode = 44
+    extVersionCode = 45
 }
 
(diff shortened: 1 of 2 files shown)


AwkwardPeak7:
this conversion doesn't seem to work for me, it gives 404 and redirects to their admin page(?)
image

vetleledaal:
The issue was that it was trying to rewrite cover URLs as well. Added a fragment for all chapter images so it can be identified in the interceptor.

The logic will be disabled if _any_ 404 is seen during the URL rewrite, as this means a previous assumption is broken. I tried to provide feedback via toast, but that broke the images completely. Instead, additional text was added to the extension settings.

AwkwardPeak7 (approved):

PR #6686, opened 2024-12-18
GalleryAdults & Fox additions

SilverBeamx:
Checklist:

This PR aims achieves 2 goals:
- GalleryAdults multisrc-lib now fetches tags from 5 pages instead of 3. This was necessary due to some sources having very few tags per page
- Implemented search for the 4 types of "currently trending" galleries which appear in the sidebar of HentaiFox's homepage

### DIFF
--- a/src/all/hentaifox/build.gradle
+++ b/src/all/hentaifox/build.gradle
@@ -3,7 +3,7 @@ ext {
     extClass = '.HentaiFoxFactory'
     themePkg = 'galleryadults'
     baseUrl = 'https://hentaifox.com'
-    overrideVersionCode = 6
+    overrideVersionCode = 7
     isNsfw = true
 }
 
--- a/src/all/hentaifox/build.gradle
+++ b/src/all/hentaifox/build.gradle
@@ -3,7 +3,7 @@ ext {
     extClass = '.HentaiFoxFactory'
     themePkg = 'galleryadults'
     baseUrl = 'https://hentaifox.com'
-    overrideVersionCode = 7
+    overrideVersionCode = 6
     isNsfw = true
 }
 
--- a/src/all/hentaifox/src/eu/kanade/tachiyomi/extension/all/hentaifox/HentaiFox.kt
+++ b/src/all/hentaifox/src/eu/kanade/tachiyomi/extension/all/hentaifox/HentaiFox.kt
@@ -146,25 +146,17 @@ class HentaiFox(
 
     override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
         // Sidebar mangas should always override any other search, so they should appear first
-        // in the case/when and only propagate to super when a "normal" search is issued
+        // and only propagate to super when a "normal" search is issued
         val sortOrderFilter = filters.filterIsInstance<SortOrderFilter>().firstOrNull()
 
-        return when {
-            sortOrderFilter?.state == 2 ->
-                sidebarRequest("top_rated")
-
-            sortOrderFilter?.state == 3 ->
-                sidebarRequest("top_faved")
-
-            sortOrderFilter?.state == 4 ->
-                sidebarRequest("top_fapped")
-
-            sortOrderFilter?.state == 5 ->
-                sidebarRequest("top_downloaded")
-
-            else ->
-                super.searchMangaRequest(page, query, filters)
+        sortOrderFilter?.let {
+            if (sortOrderFilter.state > 1) {
+                return sidebarRequest(
+                    sidebarCategoriesFilterStateMap.getValue(sortOrderFilter.state),
+                )
+            }
         }
+        return super.searchMangaRequest(page, query, filters)
     }
 
     private var csrfToken: String? = null
@@ -224,5 +216,14 @@ class HentaiFox(
             Pair("Most Fapped", "mf"),
             Pair("Most Downloaded", "md"),
         )
+    }
+
+    companion object {
+        private val sidebarCategoriesFilterStateMap = mapOf(
+            2 to "top_rated",
+            3 to "top_faved",
+            4 to "top_fapped",
+            5 to "top_downloaded",
+        ).withDefault { "top_rated" }
     }
 }
--- a/src/all/hentaifox/src/eu/kanade/tachiyomi/extension/all/hentaifox/HentaiFox.kt
+++ b/src/all/hentaifox/src/eu/kanade/tachiyomi/extension/all/hentaifox/HentaiFox.kt
@@ -1,17 +1,16 @@
 package eu.kanade.tachiyomi.extension.all.hentaifox
 
 import eu.kanade.tachiyomi.multisrc.galleryadults.GalleryAdults
+import eu.kanade.tachiyomi.multisrc.galleryadults.Genre
 import eu.kanade.tachiyomi.multisrc.galleryadults.SortOrderFilter
 import eu.kanade.tachiyomi.multisrc.galleryadults.imgAttr
 import eu.kanade.tachiyomi.multisrc.galleryadults.toDate
-import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.network.POST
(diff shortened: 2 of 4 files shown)


SilverBeamx:
Apologies in advance for the PR title not respecting standards, i forgot to set it properly before finalizing the PR

src/all/hentaifox/src/eu/kanade/tachiyomi/extension/all/hentaifox/HentaiFox.kt
```diff
@@ -100,4 +114,116 @@ class HentaiFox(
...
+            url.build().toString(),
+            setSidebarHeaders(csrfToken),
+            FormBody.Builder()
+                .add("type", category)
+                .build(),
+        )
+    }
+
+    override fun searchMangaParse(response: Response): MangasPage {
+        if (response.request.url.toString().endsWith(sidebarPath)) {
```
vetleledaal: Maybe compare against `.encodedPath` instead.

src/all/hentaifox/src/eu/kanade/tachiyomi/extension/all/hentaifox/HentaiFox.kt
```diff
@@ -100,4 +114,116 @@ class HentaiFox(
...
+                super.searchMangaRequest(page, query, filters)
+        }
+    }
+
+    private var csrfToken: String? = null
+
+    private fun sidebarRequest(category: String): Request {
+        // Use Elvis operator to fetch CSRF only if it is null
+
+        csrfToken ?: runBlocking {
```
vetleledaal: Can you get the CSRF token via an interceptor instead? `csrfTokenInterceptor` from WebNovel is an example. There are similar interceptors, but they are named something else.

src/all/hentaifox/build.gradle
```diff
@@ -3,7 +3,7 @@ ext {
     extClass = '.HentaiFoxFactory'
     themePkg = 'galleryadults'
     baseUrl = 'https://hentaifox.com'
-    overrideVersionCode = 6
+    overrideVersionCode = 7
```
vetleledaal: You are already bumping the multisrc, so no need to double bump it.

src/all/hentaifox/src/eu/kanade/tachiyomi/extension/all/hentaifox/HentaiFox.kt
```diff
@@ -100,4 +114,116 @@ class HentaiFox(
...
+            launchIO {
+                runCatching {
+                    csrfToken = client.newCall(fetchHomepage())
+                        .execute().asJsoup().let { homepageCsrfParser(it) }
+                }
+            }.join()
+        }
+
+        val url = "$baseUrl/$sidebarPath".toHttpUrl().newBuilder()
```
vetleledaal: Builder is useless here, since you just convert it back to a String.

vetleledaal (changes requested):

src/all/hentaifox/build.gradle
SilverBeamx: Makes sense

src/all/hentaifox/src/eu/kanade/tachiyomi/extension/all/hentaifox/HentaiFox.kt
SilverBeamx: WebNovel intercepts the CSRF token via an header. HentaiFox instead embeds it into the homepage itself, thus i don't think that i can intercept it the same way WebNovel does. I fetch the homepage once and then cache the token, so the traffic rate increase should be negligible

SilverBeamx: The only concern would be the Token's validity period, but in my 15 minutes of testing i have never seen it expire across multiple searches

SilverBeamx: MangaSwat uses an Interceptor with a similar usecase as this one, but all it boils down to is a wrapper around an homepage request, which is the same thing i do here. I cannot intercept it elsewhere as the CSRF token is valid only from the homepage (from what i've seen), which is normally never queried for other extension functionalities.

src/all/hentaifox/src/eu/kanade/tachiyomi/extension/all/hentaifox/HentaiFox.kt
```diff
@@ -100,4 +113,116 @@ class HentaiFox(
...
+    }
+
+    override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
+        // Sidebar mangas should always override any other search, so they should appear first
+        // in the case/when and only propagate to super when a "normal" search is issued
+        val sortOrderFilter = filters.filterIsInstance<SortOrderFilter>().firstOrNull()
+
+        return when {
+            sortOrderFilter?.state == 2 ->
+                sidebarRequest("top_rated")
```
cuong-tran: Could we put this into a map then get the URI part directly instead of checking each case? Also use that map to populate `getSortOrderURIs`

src/all/hentaifox/src/eu/kanade/tachiyomi/extension/all/hentaifox/HentaiFox.kt
cuong-tran: csrf-token is available on other pages as well, the requestTags for example. Maybe we can get that while loading tags?

cuong-tran:
How about this:
```
    override fun getSortOrderURIs(): List<Pair<String, String>> {
        return super.getSortOrderURIs() + sidebarCategoriesFilterStateMap.toList()
    }

    companion object {
        private val sidebarCategoriesFilterStateMap = mapOf(
            "Top Rated" to "top_rated",
            "Most Faved" to "top_faved",
            "Most Fapped" to "top_fapped",
            "Most Downloaded" to "top_downloaded",
        ).withDefault { "top_rated" }
    }
```

Then in `searchMangaRequest`:
```
sortOrderFilter?.state >= super.getSortOrderURIs().size ->
                sidebarCategoriesFilterStateMap[sortOrderFilter.state - super.getSortOrderURIs().size].second
```

Be careful when multisrc might change its `getSortOrderURIs` size in future then using specific index might fail.

SilverBeamx:
> How about this:
> 
> ```
>     override fun getSortOrderURIs(): List<Pair<String, String>> {
>         return super.getSortOrderURIs() + sidebarCategoriesFilterStateMap.toList()
>     }
> 
>     companion object {
>         private val sidebarCategoriesFilterStateMap = mapOf(
>             "Top Rated" to "top_rated",
>             "Most Faved" to "top_faved",
>             "Most Fapped" to "top_fapped",
>             "Most Downloaded" to "top_downloaded",
>         ).withDefault { "top_rated" }
>     }
> ```
> 
> Then in `searchMangaRequest`:
> 
> ```
> sortOrderFilter?.state >= super.getSortOrderURIs().size ->
>                 sidebarCategoriesFilterStateMap[sortOrderFilter.state - super.getSortOrderURIs().size].second
> ```
> 
> Be careful when multisrc might change its `getSortOrderURIs` size in future then using specific index might fail.

Thanks for your comments. Your code gave me an idea. I used what you suggested for the Map, but instead of relying on integers, as you pointed out, i now extract the currently selected sort label from the filter, and use the special sidebar logic only if its key is present in the Map.

SilverBeamx:
Seems ready for merge. Thanks to both of you for your insights!

SilverBeamx:
@vetleledaal PR is waiting for you review approval. Sorry for the ping, i don't wanna pass off as insistent, i justdon't know if you've read or this PR got buried under new ones :)

Smol-Ame:
It'll be reviewed whenever/whoever has the time, it won't get lost

vetleledaal (approved):

PR #6792, opened 2024-12-24
zaimanhua: add ranking filters

zhongfly:
Checklist:

### DIFF
--- a/src/zh/zaimanhua/src/eu/kanade/tachiyomi/extension/zh/zaimanhua/Zaimanhua.kt
+++ b/src/zh/zaimanhua/src/eu/kanade/tachiyomi/extension/zh/zaimanhua/Zaimanhua.kt
@@ -197,8 +197,8 @@ class Zaimanhua : HttpSource(), ConfigurableSource {
 
     // Popular
     private fun rankApiUrl(): HttpUrl.Builder =
-        "$apiUrl/comic/rank/list".toHttpUrl().newBuilder().addQueryParameter("by_time", "3")
-            .addQueryParameter("tag_id", "0").addQueryParameter("rank_type", "0")
+        "$apiUrl/comic/rank/list".toHttpUrl().newBuilder()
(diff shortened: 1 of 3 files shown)


src/zh/zaimanhua/src/eu/kanade/tachiyomi/extension/zh/zaimanhua/Zaimanhua.kt
```diff
@@ -225,16 +226,43 @@ class Zaimanhua : HttpSource(), ConfigurableSource {
     override fun searchMangaParse(response: Response): MangasPage =
         response.parseAs<ResponseDto<PageDto>>().data.toMangasPage()
 
+    override fun fetchSearchManga(
```
vetleledaal: Avoid the `fetch*` methods if possible. Try to replace with `searchMangaRequest` / `searchMangaParse`

zhongfly: As the title says, this change is actually intended to display the ranking list in, not the search, so it doesn't use the same api as the search and returns a completely different data structure.
In the `searchMangaRequest`,`query.isEmpty() && filters.filterIsInstance<RankingGroup>().firstOrNull()! = null` can determine which api to use. But in `searchMangaParse`, I don't know how to know if it's searching or getting the ranking list.

zhongfly: I thought of ways to do it, but it would make the code even uglier.

AwkwardPeak7: you forgot to remove the fetch method override, since the changes were made to search request and parse

AwkwardPeak7 (approved):

PR #6900, opened 2024-12-29
[ci] build only required modules

AwkwardPeak7:
- prepare
  - gets file diff from successful ci run to present
  - for src changes, adds the corrosponding module
  - for lib or multisrc changes, gets it's corrospoinding modules
- build step is similar as before with matrix chunks
- repo commit
  - removes old changed or deleted apks/icons from current repo
  - merges new ones
  - merges the index.min.json

needs (more) testing and review 
- pr workflow is untested yet
- ci workflow should be fine up until repo deploy step, which needs testing
  - tested in act


.github/scripts/merge-repo.py
```diff
@@ -0,0 +1,37 @@
+import sys
+import json
+from pathlib import Path
+import shutil
+
+to_delete = json.loads(sys.argv[1])
+
+for module in to_delete:
+    apk_name = f"tachiyomi-{module}-v1.4.*.apk"
```
vetleledaal:
```suggestion
    apk_name = f"tachiyomi-{module}-v*.*.*.apk"
```

.github/scripts/modules-to-build.py
```diff
@@ -0,0 +1,74 @@
+import os
+import subprocess as sp
+import re
+import json
+import sys
+from pathlib import Path
+from typing import List, Tuple, Dict, Any
```
vetleledaal: List, Tuple, and Dict can be replaced with list, tuple, and dict since [Python 3.9](https://peps.python.org/pep-0585/), which is oldest version that is still supported.

.github/scripts/modules-to-build.py
```diff
@@ -0,0 +1,74 @@
...
+        check=True,
+        capture_output=True,
+        text=True
+    ).stdout
+    
+    modules.update(result.splitlines())
+                
+    return list(modules), list(deleted)
+
+def chunker(iterable: List[str], size: int) -> Dict[str, Any]:
```
vetleledaal: Should replicate the signature of [`itertools.batched`](https://docs.python.org/3/library/itertools.html#itertools.batched), the docs contain an implementation that can be used for Python <3.12.

vetleledaal: Further down you would use something like (can probably be formatted better):
```py
chunked = {"chunk": [{"num": i, "modules": list(batch)} for i, batch in enumerate(batched(modules, int(os.getenv("CI_CHUNK_SIZE", 65))))]}
```

PR #6919, opened 2024-12-31
SlimeReadTheme: remove use of incompatible API

vetleledaal:
Not updating `FUNCTION_REGEX` so it can be merged together with #6918 without merge conflicts. Ideally `FUNCTION_REGEX` should be changed since I'm unsure if Android <9 silently fails to parse the regex. Thoughts?

Due to region blocking it has not been tested in context. Change does in any case match the expected behavior:
https://pl.kotl.in/yfc7GcqEm

Checklist:


AwkwardPeak7 (approved):
idk if we should even bother with such errors...

Issue #6237, opened 2024-11-23
E-hentai wrong url


Wakr21:
### Source information

E-hentai 1.4.22

### Source language

Multi

### Source new URL

https://e-hentai.org

### Other details

Currently the open in web browser takes you to exhentai.org which does not exist.

### Acknowledgements

- [X] I have updated all installed extensions.
- [X] I have opened WebView and checked that the source URL has not changed yet.
- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

### 

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/keiyoushi/extensions-source/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

vetleledaal:
If you fill out `ipb_member_id` and `ipb_pass_hash` in the extension settings (or have it set by logging into WebView) it will use the alternate domain. IIRC, your account needs to be at least 1 week old before you get access.

Wakr21:
> If you fill out `ipb_member_id` and `ipb_pass_hash` in the extension settings (or have it set by logging into WebView) it will use the alternate domain. IIRC, your account needs to be at least 1 week old before you get access.

Do you know if there's any way to log out from the website in Webbiew? Atm it directs to a blank screen, so I can't figure out a way to access the original website and log out.

vetleledaal:
I would try clearing out the values in the extension settings **and** clear cookies in the app settings. Clearing both since I'm not sure if they are synced.

PR #7098, opened 2025-01-10
e-hentai: favorite filtering and manual igneous cookie setting

rebelonion:
Checklist:

# Features:

## Override igneous cookie
(similar to #6237)
Sometimes the igneous cookie won't auto-populate and needs to be set manually. I've come across this problem even after opening webview to set cookies.
Also, some forks of Tachi don't have the ability to open webview (Suwayomi), so it's just eternally broken there.
This shouldn't break previous users, as the default is the same (an empty string).
Idk why this wasn't added originally, as you already need to gather cookies anyways.

## Filter a user's favorites
Pretty self-explanatory. Adds the ability to view and search your favorites by adding a filter option.

## Force e-hentai setting
Useful if someone wants to log in and view favorites or watched tags while avoiding content on exh, or if they just don't have access.


grayankit (commented):
looks good

src/all/ehentai/src/eu/kanade/tachiyomi/extension/all/ehentai/EHentai.kt
```diff
@@ -48,10 +48,12 @@ abstract class EHentai(
...
     private val memberId: String by lazy { getMemberIdPref() }
     private val passHash: String by lazy { getPassHashPref() }
+    private val igneous: String by lazy { getIgneousPref() }
+    private val forceEh: Boolean by lazy { getForceEhPref() }
 
     override val baseUrl: String
         get() = when {
-            System.getenv("CI") == "true" -> "https://e-hentai.org"
+            System.getenv("CI") == "true" || forceEh -> "https://e-hentai.org"
             memberId.isNotEmpty() && passHash.isNotEmpty() -> "https://exhentai.org"
```
AwkwardPeak7:
```suggestion
            System.getenv("CI") == "true"  -> "https://e-hentai.org"
            !forceEh && memberId.isNotEmpty() && passHash.isNotEmpty() -> "https://exhentai.org"
```

src/all/ehentai/src/eu/kanade/tachiyomi/extension/all/ehentai/EHentai.kt
```diff
@@ -561,11 +572,33 @@ abstract class EHentai(
...
+        val forceEhPref = CheckBoxPreference(screen.context).apply {
+            key = FORCE_EH
+            title = FORCE_EH_TITLE
+            summary = FORCE_EH_SUMMARY
+            setDefaultValue(FORCE_EH_DEFAULT_VALUE)
+
+            setOnPreferenceChangeListener { _, newValue ->
+                val checkValue = newValue as Boolean
+                preferences.edit().putBoolean(FORCE_EH, checkValue).commit()
+            }
```
AwkwardPeak7:
```suggestion
```
not needed, app sets the preference itself

src/all/ehentai/src/eu/kanade/tachiyomi/extension/all/ehentai/EHentai.kt
```diff
@@ -561,11 +572,33 @@ abstract class EHentai(
...
+
+        private const val IGNEOUS_PREF_KEY = "IGNEOUS"
+        private const val IGNEOUS_PREF_TITLE = "igneous"
+        private const val IGNEOUS_PREF_SUMMARY = "igneous value override"
+        private const val IGNEOUS_PREF_DEFAULT_VALUE = ""
+
+        private const val FORCE_EH = "FORCE_EH"
+        private const val FORCE_EH_TITLE = "Force e-hentai"
+        private const val FORCE_EH_SUMMARY = "Force e-hentai to avoid content on exhentai"
+        private const val FORCE_EH_DEFAULT_VALUE = false
```
AwkwardPeak7: maybe we should set this to true?

AwkwardPeak7 (changes requested):

src/all/ehentai/src/eu/kanade/tachiyomi/extension/all/ehentai/EHentai.kt
rebelonion: This really only matters if the user signs in, and I assumed that if you are taking the time to find your cookies, you're probably trying to use exh. 

I'm open to switching it if you think it would be better defaulted to true though

AwkwardPeak7 (approved):

PR #7119, opened 2025-01-11
Fix: PhenixScans - Correctly parse chapter dates

Nyantad:
This PR fixes the chapter date parsing for PhenixScans, preventing incorrect dates (like 01/21/1970) and handling future dates properly.

Checklist:


src/fr/phenixscans/src/eu/kanade/tachiyomi/extension/fr/phenixscans/PhenixScans.kt
```diff
@@ -21,4 +24,40 @@ class PhenixScans : MangaThemesia("PhenixScans", "https://phenixscans.fr", "fr",
...
+
+    override fun chapterFromElement(element: Element): SChapter {
+        val chapter = SChapter.create()
+
+        with(element) {
+            selectFirst(chapterUrlSelector)?.let { urlElement: Element ->
+                chapter.url = urlElement.attr("abs:href")
+                chapter.name = urlElement.selectFirst(chapterNameSelector)?.text() ?: ""
+            }
+
```
AwkwardPeak7: this is basically the default implementation on base class so you don't need it here. override `parseChapterDate` if needed

src/fr/phenixscans/src/eu/kanade/tachiyomi/extension/fr/phenixscans/PhenixScans.kt
```diff
@@ -21,4 +24,40 @@ class PhenixScans : MangaThemesia("PhenixScans", "https://phenixscans.fr", "fr",
...
+            if (!dateString.isNullOrEmpty()) {
+                try {
+                    val parsedDate = chapterDateFormat.parse(dateString)?.time ?: 0
+                    val currentDate = System.currentTimeMillis()
+
+                    if (parsedDate > currentDate) {
+                        chapter.date_upload = 0L
+                    } else {
+                        chapter.date_upload = parsedDate
+                    }
```
AwkwardPeak7: why check future dates?

AwkwardPeak7 (changes requested):

src/fr/phenixscans/src/eu/kanade/tachiyomi/extension/fr/phenixscans/PhenixScans.kt
Nyantad: ~~The PhenixScans website occasionally lists chapters as "announced" even though they are not yet available, so these should be disregarded as they are not currently accessible from what I've seen~~

My bad, I saw it on one chapter, but it was a one-off. Usually, this check isn't needed.

src/fr/phenixscans/src/eu/kanade/tachiyomi/extension/fr/phenixscans/PhenixScans.kt
Nyantad: Yeah i just see that im going to cut all of this

Nyantad:
My bad, turned out I didn't have much to change. Should've read the base lib better xD

AwkwardPeak7 (approved):

PR #7140, opened 2025-01-12
Fix: Improved date parsing for Keyoapp sites

Nyantad:
Introduced `DateSelector` variable in `Keyoapp` base class: This allows individual sites to define a custom CSS selector for the date element. The previous fixed selector (`.text-xs`)  failed to accurately capture the date on sites with different HTML structures, for example, it was capturing a "New" tag instead of the actual date.

Updated `ReaperScans` and `StarboundScans` to use specific `DateSelector` values:
These sites required more precise selectors to pinpoint the correct date element within their unique HTML otherwise, it would capture the "New" tag for recently released chapters.

Simplified  `parseRelativeDate`  function: Removed the code that sets the time to midnight because it caused incorrect chapter release times. Now, the function correctly calculates the release time by subtracting the relative time from the current time.
If a chapter was released "3 hours ago" and the current time is 4:00 PM, the old code would incorrectly calculate the release time as 9:00 PM the previous day (midnight minus 3 hours). The new code correctly calculates the release time as 1:00 PM today (4:00 PM minus 3 hours).

Checklist:

### DIFF
--- a/lib-multisrc/keyoapp/src/eu/kanade/tachiyomi/multisrc/keyoapp/Keyoapp.kt
+++ b/lib-multisrc/keyoapp/src/eu/kanade/tachiyomi/multisrc/keyoapp/Keyoapp.kt
@@ -254,7 +254,7 @@ abstract class Keyoapp(
     override fun chapterFromElement(element: Element): SChapter = SChapter.create().apply {
         setUrlWithoutDomain(element.selectFirst("a[href]")!!.attr("href"))
         name = element.selectFirst(".text-sm")!!.text()
-        element.selectFirst(DateSelector)?.run {
+        element.selectFirst(dateSelector)?.run {
             date_upload = text().trim().parseDate()
         }
         if (element.select("img[src*=Coin.svg]").isNotEmpty()) {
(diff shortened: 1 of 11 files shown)


src/fr/reaperscans/build.gradle
```diff
@@ -3,7 +3,7 @@ ext {
     extClass = '.ReaperScans'
     themePkg = 'keyoapp'
     baseUrl = 'https://reaper-scans.fr'
-    overrideVersionCode = 4
+    overrideVersionCode = 5
```
vetleledaal: You don't need to bump `overrideVersionCode` if you are already bumping `baseVersionCode`

lib-multisrc/keyoapp/src/eu/kanade/tachiyomi/multisrc/keyoapp/Keyoapp.kt
```diff
@@ -211,6 +211,7 @@ abstract class Keyoapp(
     protected open val authorSelector: String = "div:has(span:containsOwn(Author)) ~ div"
     protected open val artistSelector: String = "div:has(span:containsOwn(Artist)) ~ div"
     protected open val genreSelector: String = "div:has(span:containsOwn(Type)) ~ div"
+    protected open val DateSelector: String = ".text-xs"
```
vetleledaal: Use camelCase.

```suggestion
    protected open val dateSelector: String = ".text-xs"
```

src/fr/reaperscans/build.gradle
Nyantad: Alright

Nyantad:
Also i fixes relative date parsing, adds French support, and corrects Astral Manga date format.

lib-multisrc/madara/build.gradle.kts
```diff
@@ -2,7 +2,7 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 37
+baseVersionCode = 38
```
vetleledaal: Move the Madara changes so it can be handled in another PR. Bumping Madara needs additional consideration, as it would update ~500 extensions.

Nyantad: Ok

AwkwardPeak7 (approved):

PR #7161, opened 2025-01-13
Madara: Improve French relative date parsing

Nyantad:
Added support for "X jour(s)" format to handle relative dates in French. This fixes cases like "1 jour" or "3 jours" which weren't working before. I've tried to be extra careful with the regex to avoid breaking anything else in the existing date parsing. 

Checklist:


lib-multisrc/madara/src/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
```diff
@@ -920,6 +920,10 @@ abstract class Madara(
             WordSet("hace").startsWith(date) -> {
                 parseRelativeDate(date)
             }
+            // Handle "jour" with a number before it
+            date.contains(Regex("""\b\d+ jour""")) -> {
```
AwkwardPeak7, line 924: shouldn't it be added along "ago" etc in the wordset above?

Nyantad: I wouldn't add "jour" to the "ago" group, "jour" is way too common and would likely cause errors. Keeping the separate condition with the number is a better bet, i think.
For example, sometimes there "mise a jour"...

AwkwardPeak7 (approved):

PR #7244, opened 2025-01-19
VlogTruyen Update Domain & update logo & change name

dejavui:
Checklist:


AwkwardPeak7 (changes requested):
this would essentially create an new extension and users would have to uninstall old one and reinstall new one. 
so keep the old package (keep files in old place) and just change the name in code (extName, name in class etc) 
also override the `id` in the source so users don't have to migrate, unless the manga urls have changed

src/vi/vlogtruyen/src/eu/kanade/tachiyomi/extension/vi/vlogtruyen/VlogTruyen.kt
AwkwardPeak7, line 1: override id to the old one to prevent the need for migration. if manga urls have not changed

dejavui: i added override id

AwkwardPeak7 (approved):

Issue #3976, opened 2024-07-12
MangaBuddy Fetching the wrong series


Achyuth072:
### Source information

MangaBuddy 1.4.16

### Source language

English

### Steps to reproduce

1. Search for 4 cut hero 
2. Check the chapters and view them.
3. Check the chapters in the webview

### Expected behavior

 Both the web view and extensions should've shown the same 242 chapters.

### Actual behavior

The extension for some reason fetches 5 chapters from some random manga instead.

### Mihon/Tachiyomi version

1.7.4

### Android version

Android 11

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.15.3](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

PR #7270, opened 2025-01-20
Madtheme: fix chapters fetching

### LINKED ISSUE #7268: MangaBuddy not fetching some chapters
### Source information

MangaBuddy 1.4.17

### Source language

English

### Steps to reproduce

1. Open Demon King Who Lost His Job.
2. Refresh it.

### Expected behavior

It should show the newest chapters.

### Actual behavior

It doesn't.

### Mihon/Tachiyomi version

Yokai 1.9.8-b1

### Android version

14

### Other details

This is just an example other manga are broken as well.  It seems the newest extension update broke them.
Image

Image
 

### Acknowledgements

### 

---

dngonz:
Closes #7268

Checklist:

### DIFF
--- a/lib-multisrc/madtheme/build.gradle.kts
+++ b/lib-multisrc/madtheme/build.gradle.kts
@@ -2,4 +2,4 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 15
+baseVersionCode = 16
--- a/lib-multisrc/madtheme/src/eu/kanade/tachiyomi/multisrc/madtheme/MadTheme.kt
+++ b/lib-multisrc/madtheme/src/eu/kanade/tachiyomi/multisrc/madtheme/MadTheme.kt
@@ -184,18 +184,19 @@ abstract class MadTheme(
         val bookId = script.data().substringAfter("bookId = ").substringBefore(";")
         val bookSlug = script.data().substringAfter("bookSlug = \"").substringBefore("\";")
 
-        // At this moment we ca not decide which endpoint has the chapters, so we call both.
+        // At this moment we can not decide which endpoint has the chapters, so we call both.
         val idRequest = client.newCall(GET(buildChapterUrl(bookId), headers)).execute()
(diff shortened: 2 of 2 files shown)


vetleledaal:
Seems like the wrong abstraction layer? Is this something that is better scoped for each site? Or entry? Is the only way to check definitely by testing both endpoints?

Also, from the user's point of view this is an implementation detail that is hard to take a position on.

dngonz:
> Is this something that is better scoped for each site? Or entry?

Either I think, I didn't test it too much but depends on the site and/or entry, there are some site that the updated chapter list is by Id and others are by slug

> Is the only way to check definitely by testing both endpoints?

Yes I think so, the original issue #3976 in this case `4 Cut Hero` uses the [slug](https://mangabuddy.com/api/manga/4-cut-hero/chapters?source=detail) fetch and the chapter entries are all wrong but with the [id](https://mangabuddy.com/api/manga/11017/chapters?source=detail) fetch has the real entries.
By other hand in this issue `Demon King Who Lost His Job` if you search by [slug](https://mangabuddy.com/api/manga/the-demon-king-who-lost-his-job/chapters?source=detail) has a more updated chapter list than by [id](https://mangabuddy.com/api/manga/9972/chapters?source=detail).
Also if you notice on the entry site `Demon King Who Lost His Job` if you press "Show More", bellow chapters, the list will change by the id fetch, basically the API is not really in sync.

> Also, from the user's point of view this is an implementation detail that is hard to take a position on.

I could add a new option that includes both, that will compare the two requests and choose the one with more chapters.

vetleledaal:
Could the difference be that when there are less than x chapters using the 'show more' API is only done by non-users and should not be done? If you can't use that heuristic I would fetch both regardless -- the user still can't take a position site-wide, imo.

dngonz:
I think the best approach it's just call both endpoints and fetch the one that has more chapters.

vetleledaal (approved):

PR #7300, opened 2025-01-23
Add TruyenTranh3Q

bhdai:
Checklist:


src/vi/truyentranh3q/src/eu/kanade/tachiyomi/extension/vi/truyentranh3q/TruyenTranh3Q.kt
```diff
@@ -0,0 +1,273 @@
...
+            val relativeImageUrl = element.selectFirst(".book_avatar a img")?.attr("src")
+            thumbnail_url = relativeImageUrl?.let {
+                if (it.startsWith("/_next/image?")) {
+                    Uri.parse(it).getQueryParameter("url")?.let { encodedUrl ->
+                        URLDecoder.decode(encodedUrl, "UTF-8")
+                    }
+                } else {
+                    it
+                }
+            }
```
vetleledaal: Maybe something like this? Not sure if you need the URLDecoder part when using OkHttp's `HttpUrl`. Linter might disagree.
```kt
thumbnail_url = element.selectFirst(".book_avatar a img")
    ?.absUrl("src")
    ?.let { url ->
        url.toHttpUrlOrNull()
        ?.queryParameter("url")
        ?.let { URLDecoder.decode(it, "UTF-8") }
        ?: url
    }
```

src/vi/truyentranh3q/src/eu/kanade/tachiyomi/extension/vi/truyentranh3q/TruyenTranh3Q.kt
```diff
@@ -0,0 +1,273 @@
...
+        }
+    }
+
+    override fun chapterFromElement(element: Element): SChapter {
+        return SChapter.create().apply {
+            element.selectFirst(".name-chap > a")?.let {
+                name = it.text()
+                setUrlWithoutDomain(it.attr("abs:href"))
+            }
+            date_upload = parseChapterDate(element.selectFirst(".time-chap")!!.text() ?: "")
```
vetleledaal: `date_upload` is not critical, remove non-null assert.

src/vi/truyentranh3q/src/eu/kanade/tachiyomi/extension/vi/truyentranh3q/TruyenTranh3Q.kt
```diff
@@ -0,0 +1,273 @@
...
+        }
+    }
+
+    // parse pages
+    private val pageListSelector = ".chapter_content .page-chapter img"
+
+    override fun pageListParse(document: Document): List<Page> {
+        return document.select(pageListSelector).mapIndexed { idx, it ->
+            Page(idx, imageUrl = it.attr("data-src"))
```
vetleledaal:
```suggestion
            Page(idx, imageUrl = it.absUrl("data-src"))
```
```suggestion
            Page(idx, imageUrl = it.attr("abs:data-src"))
```

src/vi/truyentranh3q/src/eu/kanade/tachiyomi/extension/vi/truyentranh3q/TruyenTranh3Q.kt
```diff
@@ -0,0 +1,273 @@
...
+    private fun parseGenres(document: Document): List<Genre> {
+        return document.select(".genre-item").mapIndexed { index, element ->
+            Genre(element.text(), index + 1)
+        }
+    }
+
+    private fun fetchGenres() {
+        if (fetchGenreAttempts < 3 && genreList.isEmpty()) {
+            try {
+                genreList = client.newCall(genresRequest()).execute().asJsoup().let(::parseGenres)
```
vetleledaal, line 260: Consider piggy-backing off of other responses, like with other implementations of the dynamic genre thing. Like with an interceptor.

bhdai: could you please tell me more about this? if i parse genre from other responses, doesn't that mean the genre filter can only apply during the search? Since the the popular and the latest page don't include genre items. I wouldn't able to populate genre list from those pages. Is my understanding correct?

vetleledaal: That's right, it would only add the genres if you search for something first. When `getFilterList()` is called you can force a request (like here), and only get the genres if not set already -- through search or explicitly tapping 'filter'.

Since the times you get the genres for free is so limited it is up to you if you want do anything about it. Leaving it as-is is good as well.

src/vi/truyentranh3q/src/eu/kanade/tachiyomi/extension/vi/truyentranh3q/TruyenTranh3Q.kt
```diff
@@ -0,0 +1,269 @@
...
+                    }
+                    if (excludeGenres.isNotEmpty()) {
+                        url.addQueryParameter("nocategories", excludeGenres.joinToString(","))
+                    }
+                }
+                else -> {} // do nothing for unhandled filters
+            }
+        }
+
+        return GET(url.toString(), headers)
```
AwkwardPeak7:
```suggestion
        return GET(url.build(), headers)
```

src/vi/truyentranh3q/src/eu/kanade/tachiyomi/extension/vi/truyentranh3q/TruyenTranh3Q.kt
```diff
@@ -0,0 +1,269 @@
...
+        now.add(calendarUnit, -number)
+
+        return now.timeInMillis
+    }
+
+    private fun parseChapterDate(dateString: String): Long {
+        return try {
+            // first try to parse as a relative date
+            parseRelativeDate(dateString)
+        } catch (e: ParseException) {
```
AwkwardPeak7: are you sure it will throw `ParseException`?

AwkwardPeak7 (approved):

PR #7317, opened 2025-01-24
Bato.to: update mirrors, pick one at random, deprecated mirror migration

vetleledaal:
Merge together with #7378 to avoid bumping version twice

'Auto' option will pick a random mirror to use, based on the last update timestamp or the current extension version. The experience will be consistent on a user-by-user basis, in case ISP blocks one or more of the domains.

Checklist:


AwkwardPeak7:
imo extension version isn't really a good way to select random mirror, considering we can go months between new extension updates

AwkwardPeak7:
or maybe switch mirror automatically if current mirror is not reachable.

vetleledaal:
It's supposed to be something that doesn't update often. Although it does not really achieve that goal currently, since all users will use the same mirror.

I wanted to do something more user-specific. Install timestamp is Android specific, so maybe something like this (code below) might work?

It will spread users across mirrors, if the user wants to troubleshoot they can reinstall or select a mirror in the extension settings. Other runtime environments (Suwayomi, Tachimanga, ColorOS) will probably fail finding 'lastUpdateTime' and will fall back to using `BuildConfig.VERSION_NAME`.

```kt
private fun getMirrorPref(): String {
    return preferences.getString("${MIRROR_PREF_KEY}_$lang", MIRROR_PREF_DEFAULT_VALUE)
        ?.takeUnless { it == MIRROR_PREF_DEFAULT_VALUE }
        ?: let {
            val seed = runCatching {
                val pm = Injekt.get<Application>().packageManager
                pm.getPackageInfo(BuildConfig.APPLICATION_ID, 0).lastUpdateTime
            }.getOrElse {
                BuildConfig.VERSION_NAME.hashCode().toLong()
            }

            MIRROR_PREF_ENTRY_VALUES[1 + (seed % (MIRROR_PREF_ENTRIES.size - 1)).toInt()]
        }
}
```

vetleledaal:
Should investigate further:

> I am getting this weird error with the bato.to extension. For some of the series in my library, I get the "failed to connect to bato.to/0.0.0.0:443". I know that bato.to no longer works but the other series in the same extension change mirrors just fine. Even when I change mirrors, the ones that didn't work before will still try to connect to bato.to in webview even it the other series go to the proper mirror sites in webview. If I copy the link of  the failed series and in a web browser change 'bato.to' to one of the mirrors such as 'zbato.org', then it works. I was hoping the last update would fix this but it didn't. Even more strange than all of this is that for the series that do work, if I try to migrate the series to the bato extension,  it says it's in my library. If I try to migrate the failed series, the listing in the bato extension does not show up in my library. Migrating in this fashion (to the same extension I guess?) fixes the issue but I do not want to manually update everything

AwkwardPeak7 (approved):

PR #7468, opened 2025-02-02
revert kotlin 2.1.0 and deps for now

AwkwardPeak7:
generated serializers seem to be missing in final apk, need further investigation


src/en/mangamo/build.gradle
```diff
@@ -1,7 +1,7 @@
 ext {
     extName = 'Mangamo'
     extClass = '.Mangamo'
-    extVersionCode = 2
+    extVersionCode = 1
```
vetleledaal, line 4: This should probably be bumped to 3, assuming version 2 was broken.

PR #7586, opened 2025-02-10
weebcentral: add extra info, format manga details

### LINKED ISSUE #7574: Weeb Central feature rework
### Source name

Weeb Central

### Source language

English

Right now, Weeb Central does three things wrong:

1. It displays the status of the translations (“Official translations” if the manga has been updated with at least on English volume) in the manga description.
2. That information along with alternative names is displayed _above_ the manga description, which design-wise doesn’t make much sense. The description should take precedence, no?
3. The extension pulls the translation status of each chapter and displays it in the app. Theoretically, that should be a useful feature. Unfortunately, NepNep didn’t update the translation status of most of the already completed series after migrating to WeebCentral, so we end up with many, many fully officially translated series having “unofficial” chapters.

The solutions are fairly simple:

1. Remove the translation status from the manga description. It’s redundant information, especially after adding the per-chapter translation status.
2. Have the description be above everything else.
3. As for this, I am torn. On one hand, its inclusion isn’t all that offensive, even if it is an actual source of misinformation (for which Nep is at fault). On the other hand, it looks plain bad when you know for a fact that a chapter has official translations yet the app tells you otherwise.

### Other details

_No response_

### Acknowledgements

### 

---

mrtear:
Checklist:
Closes #7574
Supersedes #7576

weeb


src/en/weebcentral/src/eu/kanade/tachiyomi/extension/en/weebcentral/WeebCentral.kt
```diff
@@ -163,7 +167,7 @@ class WeebCentral : ParsedHttpSource() {
         element.selectFirst("svg")?.attr("stroke")?.also { stroke ->
             scanlator = when (stroke) {
                 "#d8b4fe" -> "Official"
-                "#4C4D54" -> "Unofficial"
+                "#4C4D54" -> "unknown"
```
AwkwardPeak7: just remove this, so only "official" appears when necessary

src/en/weebcentral/src/eu/kanade/tachiyomi/extension/en/weebcentral/WeebCentral.kt
```diff
@@ -107,27 +107,31 @@ class WeebCentral : ParsedHttpSource() {
...
             )
+
+            val relatedSeries = select("li:has(strong:contains(Related Series)) li")
+            if (relatedSeries.size > 0) {
+                descBuilder.append("\n\nRelated Series(s):")
+                relatedSeries.forEach { series ->
+                    val name = series.selectFirst("a")?.text()
+                    val relation = series.selectFirst("span")?.text()
+                    descBuilder.append("\n").append("• $name $relation")
+                }
```
AwkwardPeak7:
```suggestion
                relatedSeries.forEach { series ->
                    descBuilder.append("\n").append("• ${series.text()")
                }
```

AwkwardPeak7 (changes requested):

src/en/weebcentral/src/eu/kanade/tachiyomi/extension/en/weebcentral/WeebCentral.kt
Mythos: I'd prefer to have it in, as in the current state, the scanlator filter works on a blacklist basis.

With "Unofficial" or "Unknown" I'd be able to only show official chapters, as I prefer to read official translations if possible. (if the website has them marked as official of course)

Without that info that filtering wouldn't be possible anymore, unless someone makes the scanlator filter support both white- and blacklists.

Norsze: @Mythos rejected

scb261:
Imo, it's better to have "Unknown" or "Unofficial" for the grey checkmarks.

Filtering is quite a good reason for this. Personally, for some manga, I read only official scans. And I'm guessing there are other people who do this.

Another reasoning is that if Nep ever changes the checkmarks to something else, it will be easier to notice. I.e., with the current implementation, if the design of checkmarks ever changes, the source would start showing nothing for official scans, and the user would assume that it's just unofficial/unknown scans.

AwkwardPeak7:
sure

AwkwardPeak7 (approved):

PR #7671, opened 2025-02-16
CBHentai: fix uri path changed & search failed with special characters

cuong-tran:
also fix altNameSelector & fix query string with special characters

Checklist:

### DIFF
--- a/src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
+++ b/src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
@@ -23,10 +23,10 @@ class HentaiCB : Madara("CBHentai", "https://hentaicube.xyz", "vi", SimpleDateFo
 
     override val altNameSelector = ".post-content_item:contains(Tên khác) .summary-content"
 
-    // Changed from 'read' to 'manga'
     @Suppress("OVERRIDE_DEPRECATION")
     override fun fetchSearchManga(page: Int, query: String, filters: FilterList): Observable<MangasPage> {
         if (query.startsWith(URL_SEARCH_PREFIX)) {
+            // Changed from 'read' to 'manga'
             val mangaUrl = "/manga/${query.substringAfter(URL_SEARCH_PREFIX)}/"
             return client.newCall(GET("$baseUrl$mangaUrl", headers))
                 .asObservableSuccess().map { response ->
--- a/src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
+++ b/src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
@@ -25,9 +25,17 @@ class HentaiCB : Madara("CBHentai", "https://hentaicube.xyz", "vi", SimpleDateFo
 
     @Suppress("OVERRIDE_DEPRECATION")
     override fun fetchSearchManga(page: Int, query: String, filters: FilterList): Observable<MangasPage> {
-        if (query.startsWith(URL_SEARCH_PREFIX)) {
+        // Special characters causing search to fail
+        val queryFixed = query
+            .replace("–", "-")
+            .replace("’", "'")
+            .replace("“", "\"")
+            .replace("”", "\"")
+            .replace("…", "...")
+
+        if (queryFixed.startsWith(URL_SEARCH_PREFIX)) {
             // Changed from 'read' to 'manga'
-            val mangaUrl = "/manga/${query.substringAfter(URL_SEARCH_PREFIX)}/"
+            val mangaUrl = "/manga/${queryFixed.substringAfter(URL_SEARCH_PREFIX)}/"
             return client.newCall(GET("$baseUrl$mangaUrl", headers))
                 .asObservableSuccess().map { response ->
                     val manga = mangaDetailsParse(response).apply {
@@ -38,7 +46,7 @@ class HentaiCB : Madara("CBHentai", "https://hentaicube.xyz", "vi", SimpleDateFo
                 }
         }
 
-        return super.fetchSearchManga(page, query, filters)
+        return super.fetchSearchManga(page, queryFixed, filters)
     }
 
     private val oldMangaUrlRegex by lazy { Regex("""^$baseUrl/read/""") }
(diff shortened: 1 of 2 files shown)


src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
```diff
@@ -12,6 +18,35 @@ class HentaiCB : Madara("CBHentai", "https://hentaicube.xyz", "vi", SimpleDateFo
...
+                }
+        }
+
+        return super.fetchSearchManga(page, query, filters)
+    }
+
+    // Change old entries from 'read' to 'manga'
+    override fun getMangaUrl(manga: SManga): String {
+        return super.getMangaUrl(manga)
+            .replace("""^$baseUrl/read/""".toRegex(), "$baseUrl/manga/")
```
vetleledaal: Extract regex out of function.

cuong-tran:
Might as well tell people to migrate old entries to avoid finding duplicate when browsing

src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
```diff
@@ -12,6 +18,45 @@ class HentaiCB : Madara("CBHentai", "https://hentaicube.xyz", "vi", SimpleDateFo
...
     override val filterNonMangaItems = false
 
+    // Changed from 'manga' to 'read'
+    override val mangaSubString = "read"
+
+    override val altNameSelector = ".post-content_item:contains(Tên khác) .summary-content"
+
+    @Suppress("OVERRIDE_DEPRECATION")
+    override fun fetchSearchManga(page: Int, query: String, filters: FilterList): Observable<MangasPage> {
+        // Special characters causing search to fail
```
vetleledaal: That's usually a very good hint that you should use a `HttpUrlBuilder`. You can still do the replacements, but it is not exhaustive.

cuong-tran: By `HttpUrlBuilder`, do you mean the `URL_SEARCH_PREFIX` part? I actually copied it from Madara just to replace the uri path. If you do mean that part I can rewrite it into using `HttpUrlBuilder`.

The replacements are meant for searching when calling super. It should be moved down after the `URL_SEARCH_PREFIX` part. And I just trying to find the actual characters used on the website which causing failed search (because I need actual special characters causing failure to do the replacement)

src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
```diff
@@ -12,6 +18,45 @@ class HentaiCB : Madara("CBHentai", "https://hentaicube.xyz", "vi", SimpleDateFo
 
     override val filterNonMangaItems = false
 
+    // Changed from 'manga' to 'read'
+    override val mangaSubString = "read"
+
+    override val altNameSelector = ".post-content_item:contains(Tên khác) .summary-content"
+
+    @Suppress("OVERRIDE_DEPRECATION")
```
AwkwardPeak7:
```suggestion
```

src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
```diff
@@ -12,6 +18,45 @@ class HentaiCB : Madara("CBHentai", "https://hentaicube.xyz", "vi", SimpleDateFo
...
+                    }
+
+                    MangasPage(listOf(manga), false)
+                }
+        }
+
+        return super.fetchSearchManga(page, queryFixed, filters)
+    }
+
+    private val oldMangaUrlRegex by lazy { Regex("""^$baseUrl/read/""") }
```
AwkwardPeak7: no need for lazy

src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
```diff
@@ -12,6 +18,45 @@ class HentaiCB : Madara("CBHentai", "https://hentaicube.xyz", "vi", SimpleDateFo
...
+        // Special characters causing search to fail
+        val queryFixed = query
+            .replace("–", "-")
+            .replace("’", "'")
+            .replace("“", "\"")
+            .replace("”", "\"")
+            .replace("…", "...")
+
+        if (queryFixed.startsWith(URL_SEARCH_PREFIX)) {
+            // Changed from 'read' to 'manga'
```
AwkwardPeak7: why?

AwkwardPeak7 (changes requested):

src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
cuong-tran: Also copied from inside Madara, to support input string like "slug:[manga-path]".

Because the path changed so I took it outside.

src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
vetleledaal: Yeah, `URL_SEARCH_PREFIX`. Current approach seems good.

src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
AwkwardPeak7: I should have been more clear, why set it explicitly to `/manga` when subdir is `/read`?

cuong-tran: previously, it was using `mangaSubString = "manga"` for both browsing & `slug:[manga-path]`.

Now this site is using `mangaSubString = "read"` for browsing while using "manga" for manga url so this `URL_SEARCH_PREFIX` must set to "manga" for user to open specific manga with `slug:[manga-path]`.

AwkwardPeak7 (approved):

PR #7674, opened 2025-02-16
Madara: Expose genresList

cuong-tran:
Expose genresList so that:
- Extensions be able to read/write the genresList. Some of the extensions here (updated along in this PR) had duplicated code to load genresList in its own way.
- Extensions can also hard-code a list of genres so they have something ready before actually querying website for genres list. Given that genres list rarely changes. If the genres list does change completely then the hard-code list might be useless, but with the former way genres filter is useless before any loading, anyway. This would also help forks like SY which support saved-search based on filters' value.
- Also update the `fetchSearchManga` to using `HttpUrlBuilder`

Checklist:

### DIFF
--- a/lib-multisrc/madara/build.gradle.kts
+++ b/lib-multisrc/madara/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 39
+baseVersionCode = 40
 
 dependencies {
     api(project(":lib:cryptoaes"))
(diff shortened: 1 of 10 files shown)


AwkwardPeak7 (commented):
how would this preload the genres?

cuong-tran:
The "preload" hasn't added yet.
This would allows:
- Extensions to read/write the genresList. Some of the extensions here must duplicate code to load genresList in its own way.
- Extensions can hard-code a list of genres (the preload part) so they have something ready before actually request sites for genres list. Given that genres list rarely changes. If the genres list does change completely then the hard-code list might be useless, but with the former way genres filter is useless before any loading, anyway.

AwkwardPeak7 (approved):

PR #7683, opened 2025-02-17
add helper utils

AwkwardPeak7:
Add some common boilerplate functions in a shared utils module. More functions can be added as needed, suggestions welcome.

Will do  a separate pr to replace current usage.

### DIFF
--- a/buildSrc/src/main/kotlin/lib-multisrc.gradle.kts
+++ b/buildSrc/src/main/kotlin/lib-multisrc.gradle.kts
@@ -38,6 +38,7 @@ kotlinter {
 
 dependencies {
     compileOnly(versionCatalogs.named("libs").findBundle("common").get())
+    implementation(project(":utils"))
 }
 
 tasks {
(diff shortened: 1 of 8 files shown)


vetleledaal (approved):
Looks good- but how do we want to handle versioning?

In my opinion:
For new features (not touching existing interfaces) there will be no functional difference and we can avoid bumping dependents. 

Fixing existing interfaces should obviously come with a bump.

'Refactoring' existing code should also come alongside a version bump. While we want to avoid behavior changes with a refactor, given the many unknown unknowns we can't just assume changes will work on every app, every device, and so on.

There might be some complex combination of versioning that could be implemented, but I don't think the tradeoffs are worth it, which are numerous. I just wanted to raise the concern so we don't dig ourself into some complicated hole further down the line.

---

In any case, the current feature set looks good. Maybe we can deploy it to a subset of extensions first? I'm not personally familiar with all the Kotlin features used here.

AwkwardPeak7:
versioning these would be difficult, but since all of them are independent, we only need to bump versions of extensions if the thing they are using is changed, and I don't think the implementation of existing ones will change that much

PR #7786, opened 2025-02-24
CbHentai: Add override baseUrl, Change baseUrl, Fix Regex

dejavui:
Checklist:
- Fixed Regex bug because previously manga URLs were not just /read/ but could be /manga/ leading to double /manga/ in manga URL and use use mangaSubString instead of "manga" works more accurately
- Change baseUrl to resolve http 404 latestUpdatesRequest & popularMangaRequest
- Add override baseUrl Allow users to freely change the desired url.

### DIFF
--- a/src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
+++ b/src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
@@ -1,25 +1,59 @@
 package eu.kanade.tachiyomi.extension.vi.hentaicube
 
+import android.content.SharedPreferences
+import android.widget.Toast
+import androidx.preference.EditTextPreference
+import androidx.preference.PreferenceScreen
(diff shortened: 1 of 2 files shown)


src/vi/hentaicube/src/eu/kanade/tachiyomi/extension/vi/hentaicube/HentaiCB.kt
```diff
@@ -51,15 +85,40 @@ class HentaiCB : Madara("CBHentai", "https://hentaicube.xyz", "vi", SimpleDateFo
         return super.fetchSearchManga(page, queryFixed, filters)
     }
 
-    private val oldMangaUrlRegex = Regex("""^$baseUrl/read/""")
+    private val oldMangaUrlRegex = Regex("^$baseUrl/\\w+/")
 
-    // Change old entries from 'read' to 'manga'
+    // Change old entries from mangaSubString
```
cuong-tran, line 90: perhap should encourage user to migrate old entries instead. Keeping this will surely leads to duplicate entries

AwkwardPeak7 (approved):

PR #7781, opened 2025-02-23
Mangakakalot, Manganato: Updated Domain, Selectors and Filters

JakeeLuwi:
Closes #7754 

From what I see, Mangakakalot and Manganato are in the phase of migrating resources. Because of that, filters and release does not properly work/inaccurate. In order to receive updates on saved mangas, migration is needed but I believe this will all be fixed when the site is done updating.

Fixes/changes were not done in mangabox since it could break other extensions that depend on it (Mangabat and Mangairo)

Also fixed 403 errors.

Checklist:

- [ ] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [ ] Have removed `web_hi_res_512.png` when adding a new extension

PR #7920, opened 2025-03-04
Update Mangabat.kt

brej-29:
Mangabat down. Domain changed

Checklist:


brej-29:
kindly change the mangabat old url to the new one

JakeeLuwi:
This won't work with just changing the URL. [Mangabat](https://www.mangabats.com/) uses the same template and layout of new Mangakakalot and Manganato which does not work well with default `Mangabox` class. See #7781 for an example of a working implementation.

Norsze:
It's like they got a new lock, you can't just stick the old key in.

Also actually try to follow templates and look how people do things, if you have questions you're better off asking on Discord than just opening random stuff here.

PR #7935, opened 2025-03-05
Colamanga

dodieboy:
Checklist:


lib-multisrc/colamanga/src/eu/kanade/tachiyomi/multisrc/colamanga/ColaManga.kt
```diff
@@ -164,13 +167,16 @@ abstract class ColaManga(
             statusCompleted -> SManga.COMPLETED
             else -> SManga.UNKNOWN
         }
+        val formatter: SimpleDateFormat = SimpleDateFormat("yyyy-MM-dd")
+        lastUpdatedDate = formatter.parse(document.selectFirst("span.fed-text-muted:contains($lastUpdated) + a")?.text()).getTime()
     }
```
AwkwardPeak7: this would only set a date when user refresh chapter list manually, and this can have race conditions on library update too. Besides, I don't think setting date of each chapter to last upload is a good idea, maybe only set the latest chapter date?

Better to parse and set it `chapterListParse` (or `chapterFromElement`)
also save the `SimpleDateFormat` in class as val to avoid recreating it each time.

AwkwardPeak7 (changes requested):

lib-multisrc/colamanga/src/eu/kanade/tachiyomi/multisrc/colamanga/ColaManga.kt
dodieboy: noted, just updated the code

lib-multisrc/colamanga/src/eu/kanade/tachiyomi/multisrc/colamanga/ColaManga.kt
```diff
@@ -166,7 +171,14 @@ abstract class ColaManga(
         }
     }
 
-    override fun chapterListSelector(): String = "div:not(.fed-hidden) > div.all_data_list > ul.fed-part-rows a"
+    override fun chapterListSelector() = throw UnsupportedOperationException()
+
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val document = response.asJsoup()
+        return document.select("div:not(.fed-hidden) > div.all_data_list > ul.fed-part-rows a").map{ chapterFromElement(it) }.apply {
```
AwkwardPeak7: keep selector in `chapterListSelector`

lib-multisrc/colamanga/src/eu/kanade/tachiyomi/multisrc/colamanga/ColaManga.kt
```diff
@@ -166,7 +171,14 @@ abstract class ColaManga(
...
+            this[0].date_upload = dateFormat.parse(document.selectFirst("span.fed-text-muted:contains($lastUpdated) + a")?.text()).getTime()
```
AwkwardPeak7:
```suggestion
            this[0].date_upload = dateFormat.tryParse(document.selectFirst("span.fed-text-muted:contains($lastUpdated) + a")?.text())
```
import `keiyoushi.utils.tryParse`

AwkwardPeak7 (changes requested):

src/zh/onemanhua/build.gradle
```diff
@@ -3,7 +3,7 @@ ext {
     extClass = '.Onemanhua'
     themePkg = 'colamanga'
     baseUrl = 'https://www.colamanga.com'
-    overrideVersionCode = 15
+    overrideVersionCode = 16
```
AwkwardPeak7: no need to bump since multisrc was bumped

AwkwardPeak7 (changes requested):
bump `baseVersionCode` of the multisrc

lib-multisrc/colamanga/src/eu/kanade/tachiyomi/multisrc/colamanga/ColaManga.kt
```diff
@@ -167,6 +173,13 @@ abstract class ColaManga(
 
     override fun chapterListSelector(): String = "div:not(.fed-hidden) > div.all_data_list > ul.fed-part-rows a"
 
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val document = response.asJsoup()
+        return document.select(chapterListSelector()).map { chapterFromElement(it) }.apply {
+            this[0].date_upload = dateFormat.tryParse(document.selectFirst("span.fed-text-muted:contains($lastUpdated) + a")?.text())
```
cuong-tran: Why only apply to first child? And what if the selector failed and it returns an empty list?

dodieboy: the website only provide the last chapter update date. by apply to only first child, any child below that do not have a date before will inherit the date from the first child base on my testing.
I'm new here, if i'm wrong, pls update me. thks

dodieboy: if selector failed and it returns an empty list, the app will just show "No chapters found"

cuong-tran: > if selector failed and it returns an empty list, the app will just show "No chapters found"

but the part `this[0]` will probably cause a Null pointer or Out of range as such if the list is empty, no?

dodieboy: true, but I not able to check as I'm not sure how to setup the debugging console so I'm not able to see the error log. I have been building apk than run on my android phone

AwkwardPeak7:
```kt
.apply {
    if (isNotEmpty()) {
        this[0].date_upload = ...
    }
}
```
do it like this

dodieboy: Omg, I have been do like this.isnotempty() and keep getting error. Thks

AwkwardPeak7 (approved):

PR #7915, opened 2025-03-04
Rewrite Mangabox (Mangakakalot, Manganato, Mangabat) to Allow Mirrors and CDN Fallbacks

JakeeLuwi:
Closes #7896 
Closes #7873 
Closes #7943

The extension will now try to use every CDNs available to use when retrieving page images. ~~Also did not bumped Kakalot and Nato's version since I only placed their implementation on Mangabox instead since Mangabat apparently became a CC of 2.~~ I might have bumped those extensions incorrectly..

~~Still no update for Mangairo but if it ever comes available during PR, I will include it here.~~ Removed mangairo.

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [x] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [ ] Have removed `web_hi_res_512.png` when adding a new extension

PR #8009, opened 2025-03-10
Mangabat URL update

blackletum:
Mangabat's URL has recently changed from mangabat(dot)com to mangabats(dot)com

This pull request includes changes to the files to fix the URL to the new domain.

### DIFF
--- a/src/en/mangabat/build.gradle
+++ b/src/en/mangabat/build.gradle
@@ -2,7 +2,7 @@ ext {
     extName = 'Mangabat'
     extClass = '.Mangabat'
     themePkg = 'mangabox'
-    baseUrl = 'https://h.mangabat.com'
+    baseUrl = 'https://h.mangabats.com'
     overrideVersionCode = 5
     isNsfw = true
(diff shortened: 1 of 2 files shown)


JakeeLuwi:
This PR is in #7915 already.
Also, it is `https://www.mangabats.com/`, not `https://h.mangabats.com`. The latter one leads to nothing.

blackletum:
> This PR is in #7915 already. Also, it is `https://www.mangabats.com/`, not `https://h.mangabats.com`. The latter one leads to nothing.

Even then, I tried to change the url to `https://www.mangabats.com/` and recompile and it still doesn't work within the mihon app.

then I tried to bring in the changes in the PR you linked, but it just comes up with a 404, even though when you click "open in webview" it will open the site w/o issue...

I will continue this convo on that PR, ty for the heads up!

Norsze:
@blackletum there are more extensive changes than a URL change. Unless you actually know what you're doing you will be of little help.

PR #8049, opened 2025-03-12
Comick: Added Exclude Tags Filter, Fixed Demographic Type, Added "None" / "Others" Filter, Added Content Rating Filter

Cleover:
Checklist:


FourTOne5:
Should be separate PR for each source

cuong-tran:
add Comick name into title

src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
```diff
@@ -78,6 +78,34 @@ abstract class Comick(
...
+            summaryOn = intl["show_alternative_titles_on"]
+            summaryOff = intl["show_alternative_titles_off"]
+            setDefaultValue(SHOW_ALTERNATIVE_TITLES_DEFAULT)
+
+            setOnPreferenceChangeListener { _, newValue ->
+                preferences.edit()
+                    .putBoolean(SHOW_ALTERNATIVE_TITLES_PREF, newValue as Boolean)
+                    .commit()
+            }
+        }.also(screen::addPreference)
```
AwkwardPeak7, line 94: imo should always show alternative titles, so no pref

src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
```diff
@@ -350,6 +390,18 @@ abstract class Comick(
                         }
                     }
 
+                    is ExcludedTagFilter -> {
+                        if (it.state.isNotEmpty()) {
+                            it.state.split(",").forEach {
+                                addQueryParameter(
+                                    "excluded-tags",
+                                    it.trim().lowercase().replace(SPACE_AND_SLASH_REGEX, "-")
+                                        .replace("'-", "-and-039-").replace("'", "-and-039-"),
```
AwkwardPeak7: make a function for this and use it it both places

src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
```diff
@@ -458,7 +512,13 @@ abstract class Comick(
 
                 publishedChapter && noGroupBlock
             }
-            .map { it.toSChapter(mangaUrl) }
+            .map { chapter ->
+                chapter.toSChapter(mangaUrl).apply {
+                    if (preferences.showChapterVotes) {
+                        name = "$name (${chapter.up_count - chapter.down_count}↑)"
```
AwkwardPeak7: not sure if I like this in chapter name...

also what if it is negative? shouldn't the arrow be down?

src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Dto.kt
```diff
@@ -191,6 +193,8 @@ class Chapter(
     private val chap: String = "",
     private val vol: String = "",
     @SerialName("group_name") val groups: List<String> = emptyList(),
+    val up_count: Int = 0,
+    val down_count: Int = 0,
```
AwkwardPeak7: use `@SerialName` and make the kotlin property camelCase

AwkwardPeak7 (changes requested):

src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
Cleover: Is there harm in having the option?

I can see use in enabling it by default, however the alternative titles list takes up a very large part of the description, it's something I've found personally kinda ehh as it makes it harder to see the tags at a glance when clicking on a series as I need to scroll past it all. I can't see it being useful to most people.

IMG_332612574C45-1

src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
Cleover: Honestly I'm unsure how the site does it. The API has down_count and I know the site combines them however I haven't found a way to down vote a series to be able to see if the site uses -1 or a down arrow. I would've assumed -1 however.

It took me awhile to find a chapter with a down vote in general (and even then it was one down vote with thousands of upvotes), I'm not sure if its a legacy thing but I've asked in their Discord for now . 👍

If not in the chapter name where else could it go? I think its useful information as it lets you know which chapter is better to read quality wise, however I've made it disabled by default as I know it may be unwanted for some people)

cuong-tran: will this force app to rename chapter's download each update? will it cause unnecessary storage wear?

src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
AwkwardPeak7: sure ig

src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
AwkwardPeak7: as mentioned by cuong, this will mess with downloads, and honestly I don't see enough of a reason to include this...
The extensions-api doesn't provide any way to show extra data to the users, maybe when (and if) such and option exists, we can reconsider

so please remove this, other changes are fine

Cleover: It's been removed now 👍🏼

Cheers'

AwkwardPeak7 (approved):

PR #8152, opened 2025-03-21
Doujin.io: only use images as pages

LetrixZ:
Checklist:

### DIFF
--- a/src/en/doujinio/build.gradle
+++ b/src/en/doujinio/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Doujin.io - J18'
     extClass = '.Doujinio'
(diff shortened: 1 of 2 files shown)


FourTOne5:
What are the non image content?

LetrixZ:
> What are the non image content?

Currently, this appears at the end of the page list

image

And in the reader shows as a end of chapter screen loaded inside an iframe

FourTOne5:
In my opinion we should filter unnecessary things out instead of only allowing images.

LetrixZ:
But the reader can only display images, right?

AwkwardPeak7 (approved):

PR #8328, opened 2025-04-01
Bato.to: Add option to open links in v3x

Bartuzen:
Checklist:

### DIFF
--- a/src/all/batoto/build.gradle
+++ b/src/all/batoto/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Bato.to'
     extClass = '.BatoToFactory'
-    extVersionCode = 49
+    extVersionCode = 50
     isNsfw = true
 }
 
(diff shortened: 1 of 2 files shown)


AwkwardPeak7 (commented):
doesn't bato already have a link on their v2x site that takes you to v3x? besides I think it should explicitly return v2x urls, until they start defaulting to v3x and the extension is rewritten for v3x

Bartuzen:
> doesn't bato already have a link on their v2x site that takes you to v3x?

Yes, but it's not permanent. You still have to manually navigate every time you open a manga or chapter.

> besides I think it should explicitly return v2x urls, until they start defaulting to v3x and the extension is rewritten for v3x

I think it makes sense to have it as an option, which is false by default, for those who prefer to see the new version in WebView.
Also, I checked Mihon's source code and unless I missed anything, `getMangaUrl` and `getChapterUrl` are only used for WebView and sharing, so it should work without rewriting the extension for v3x.

AwkwardPeak7:
I meant the webview opening a different version of the site than what extension scraps can cause disparities, as primary function of webview is to check if website is working when facing issues in app/extension

Bartuzen:
I hadn’t considered it from that angle. Personally, I use WebView frequently to check data that Mihon doesn’t display, like comments.

AwkwardPeak7:
the same comments are there on v2x page too, so I am rejecting this

PR #8361, opened 2025-04-04
Fix MangaHub (multisrc) Chapter List Selector

### LINKED ISSUE #8357: Mangahub not showing full chapters
Image

### Source information

"Mangahub 1.4.40"

### Source language

"English"

### Steps to reproduce

1. Open a popular entry with more than 100 chapters;
2. Check the number of chapters on actual site;
3. Check number of chapters on the app.
Issue: Number of chapters do not match and there is an unlabeled chapter on the start.

### Expected behavior

"Number of chapters should match"

### Actual behavior

"Number of chapters do not match"

### Mihon/Tachiyomi version

"0.18.0"

### Android version

"Android 13"

### Other details

They somehow made it so the app cant retrieve past the first page I think.

### Acknowledgements

### 

---

JakeeLuwi:
Closes #8357 

Checklist:

### DIFF
--- a/lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
+++ b/lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
@@ -276,7 +276,7 @@ abstract class MangaHub(
 
     private fun chapterFromElement(element: Element, head: Element): SChapter {
         val chapter = SChapter.create()
-        val potentialLinks = element.select("a[href*='$baseUrl/chapter/']:not([rel*=nofollow]):not([rel*=noreferrer])")
+        val potentialLinks = element.select("a[href*='$baseUrl/chapter/']")
         var visibleLink = ""
         potentialLinks.forEach { a ->
             val className = a.className()
--- a/lib-multisrc/mangahub/build.gradle.kts
+++ b/lib-multisrc/mangahub/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 30
+baseVersionCode = 31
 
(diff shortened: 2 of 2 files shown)


lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -276,7 +276,7 @@ abstract class MangaHub(
 
     private fun chapterFromElement(element: Element, head: Element): SChapter {
         val chapter = SChapter.create()
-        val potentialLinks = element.select("a[href*='$baseUrl/chapter/']:not([rel*=nofollow]):not([rel*=noreferrer])")
+        val potentialLinks = element.select("a[href*='$baseUrl/chapter/']")
```
vetleledaal, line 279: This (`:not(...)`) was probably added for a reason. Would it be possible to make this more robust? Check `git blame`, maybe. MangaHub tends to change frequently so the extension breaks.

JakeeLuwi: If I'm correct, there was a case before where there are multiple links (anchor) in the selected element that acts as a "fake" link (probably they use it to hardcode ad links) that the extension mistakenly parses. The only indicator that the link is fake is by the attributes specified in `:not`. Fast forward to today, they seem to have stopped doing that but also recently included those attributes on correct links. Some chapters (mostly chapters 1-100) do not have those attributes but the newer ones do, causing them to be excluded when parsing chapter lists.

I can't see any robust way of implementing that as there is no example case of them messing up the chapter list (or any other feature) heavily. The best thing to do right now is to just remove the `:not` selector to make the extension work as intended.

vetleledaal (approved):

PR #8392, opened 2025-04-07
MangaHub (multisrc) - Rewrite Extension to Use the API Instead of Scraping

### LINKED ISSUE #8390: Mangahub shows weird chapter numbers (i.e 2304920, 3039445, etc)
### Source information

Extension name: MangaHub, Extension version: 1.4.41

### Source language

English 

### Steps to reproduce

Step 1: Just update the library 
Step 2: the new chapter for any entry will include those weird chapters

### Expected behavior

It shouldn't be like this. It should show only actual chapters which should have numbers with tens, hunders or rarely thousands in them. (10, 100, 1000) 

### Actual behavior

But instead it shows chapters with more 4 numbers in them. (203404, 2029648, etc)

### Mihon/Tachiyomi version

version: 1.13.0, app:komikku

### Android version

android 15

### Other details

_No response_

### Acknowledgements

### 

---

JakeeLuwi:
Closes #8390

Improvements and Fixes:
- Now fully uses API (GraphQL), migrated from `ParsedHttpSource` to `HttpSource`
- Improved API queries to adhere kotlin's string interpolation format (`$`) and allow compile-time checks
- Updated filters and now allows you to select multiple genres for search
- Add an toggleable support for named chapter titles
- Dates are now more accurate
- Fixed everything that involves web scraping (it doesn't use one now)
- Loads slightly faster as there are no more overhead from web scraping
- Improved API key refresh logic

Entry migration is not needed but the entry must be refreshed first in order for the details to be updated

Checklist:


MikeZeDev:
Why not using the website API to get chapters ? Ah, perhaps because of ate limit 🤷

JakeeLuwi:
> Why not using the website API to get chapters ? Ah, perhaps because of ate limit 🤷

I believe there's no API for retrieving and loading entry chapters, only search. If you happen to know the endpoints for that, I'll rewrite the extension to use that (If that's more efficient to do than scraping.)

MikeZeDev:
> > Why not using the website API to get chapters ? Ah, perhaps because of ate limit 🤷
> 
> I believe there's no API for retrieving and loading entry chapters, only search. If you happen to know the endpoints for that, I'll rewrite the extension to use that (If that's more efficient to do than scraping.)

There is . "Path" being the website identifier ('m01') and manga.identifier is the manga slug. Unfortunately that sample doesn't return chapter date but it should be here.

```ts
    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
        const query = `{
            manga(x: ${this.path}, slug: "${manga.Identifier}") {
                chapters {
                    id, number, title, slug
                }
            }
        }`;
        const { manga: { chapters } } = await this.FetchGraphQL<APIChapters>(query);
        return chapters.map(chapter => {
            let title = `Ch.${chapter.number}`;
            title += chapter.title ? ` - ${chapter.title}` : '';
            return new Chapter(this, manga, chapter.number.toString(), title.trim());
        });
    }
```

Same for pages 

```ts
    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
        const query = `{
            chapter(x: ${this.path}, slug: "${chapter.Parent.Identifier}", number: ${chapter.Identifier}) {
                pages
            }
        }`;
        const { chapter: { pages } } = await this.FetchGraphQL<APIPages>( query );
        const pagesJSON = JSON.parse(pages) as JSONPages;
        return pagesJSON.i.map(page => new Page(this, chapter, new URL(`${pagesJSON.p}${page}`, this.cdnURL)));
    }
```

the problem with that approach is that it will perhaps nullify chapter identifiers currently used.,Its still more efficient 🤷

JakeeLuwi:
> > > Why not using the website API to get chapters ? Ah, perhaps because of ate limit 🤷
> > 
> > 
> > I believe there's no API for retrieving and loading entry chapters, only search. If you happen to know the endpoints for that, I'll rewrite the extension to use that (If that's more efficient to do than scraping.)
> 
> There is . "Path" being the website identifier ('m01') and manga.identifier is the manga slug. Unfortunately that sample doesn't return chapter date but it should be here.
> 
> ```ts
>     public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
>         const query = `{
>             manga(x: ${this.path}, slug: "${manga.Identifier}") {
>                 chapters {
>                     id, number, title, slug
>                 }
>             }
>         }`;
>         const { manga: { chapters } } = await this.FetchGraphQL<APIChapters>(query);
>         return chapters.map(chapter => {
>             let title = `Ch.${chapter.number}`;
>             title += chapter.title ? ` - ${chapter.title}` : '';
>             return new Chapter(this, manga, chapter.number.toString(), title.trim());
>         });
>     }
> ```
> 
> Same for pages
> 
> ```ts
>     public override async FetchPages(chapter: Chapter): Promise<Page[]> {
>         const query = `{
>             chapter(x: ${this.path}, slug: "${chapter.Parent.Identifier}", number: ${chapter.Identifier}) {
>                 pages
>             }
>         }`;
>         const { chapter: { pages } } = await this.FetchGraphQL<APIPages>( query );
>         const pagesJSON = JSON.parse(pages) as JSONPages;
>         return pagesJSON.i.map(page => new Page(this, chapter, new URL(`${pagesJSON.p}${page}`, this.cdnURL)));
>     }
> ```
> 
> the problem with that approach is that it will perhaps nullify chapter identifiers currently used.,Its still more efficient 🤷

~I'll look into this in a separate PR.~ Will do it in this PR to make the version bump only once.

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -133,35 +148,34 @@ abstract class MangaHub(
...
-        val mangas = doc.select(popularMangaSelector())
-            .map {
-                SMangaDTO(
-                    it.select("h4 a").attr("abs:href"),
-                    it.select("h4 a").text(),
-                    it.select("img").attr("abs:src"),
-                    it.toSignature(),
-                )
-            }
+        val mangaList = json.decodeFromString<ApiSearchResponse>(response.body.string())
```
AwkwardPeak7: use `response.parseaAs` from `keiyoushi.utils`

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -133,35 +148,34 @@ abstract class MangaHub(
...
+
+        val mangas = mangaList.data.search.rows.map {
+            SMangaDTO(
+                "$baseUrl/manga/${it.slug}",
+                it.title,
+                "$baseThumbCdnUrl/${it.image}",
+                it.toSignature(),
+            )
+        }
             .distinctBy { it.signature }
```
AwkwardPeak7, line 247: is this needed?

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -218,158 +215,77 @@ abstract class MangaHub(
...
 
-    override fun searchMangaNextPageSelector() = popularMangaNextPageSelector()
-
     // manga details
-    override fun mangaDetailsParse(document: Document): SManga {
+    override fun mangaDetailsRequest(manga: SManga): Request = postRequestGraphQL(MANGA_DETAILS_QUERY(mangaSource, manga.url.removePrefix("/manga/")))
+
+    override fun mangaDetailsParse(response: Response): SManga {
+        val rawManga = json.decodeFromString<ApiMangaDetailsResponse>(response.body.string())
         val manga = SManga.create()
```
AwkwardPeak7:
```
return SManga.create().apply {
    ...
}

File: lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt [hunk 4]
```diff
@@ -218,158 +215,77 @@ abstract class MangaHub(
...
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val chapterList = json.decodeFromString<ApiMangaDetailsResponse>(response.body.string())
+
+        return chapterList.data.manga.chapters!!.map {
+            val chapter = SChapter.create()
+
+            chapter.name = it.title
+            chapter.url = "/${chapterList.data.manga.slug}/chapter-${it.number}"
+            chapter.chapter_number = it.number
+            chapter.scanlator = "#${if (it.number % 1 == 0f) it.number.toInt() else it.number}" // Title namings are very inconsistent, some of them don't have the numbers specified so we need a way to display it uniformly
```
AwkwardPeak7: add number to chapter title no scanlator

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHubQueries.kt
```diff
@@ -40,3 +89,62 @@ data class ApiChapterPages(
     val p: String,
     val i: List<String>,
 )
+
+// Search, Popular, Latest
+@Serializable
```
AwkwardPeak7: put dtos in separate file and use regular classes instead of data classes

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHubQueries.kt
```diff
@@ -40,3 +89,62 @@ data class ApiChapterPages(
...
+data class ApiSearchResponse(
+    val data: ApiSearchObject,
```
AwkwardPeak7: consider:
```kt
class Data<T>(
    val data: T
)
```
use: `response.parseAs<Data<AnotherClass>>()`

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -133,35 +148,34 @@ abstract class MangaHub(
...
-        val genres = this.select(".genre-label").joinToString { it.text() }
+    private fun ApiMangaSearchItem.toSignature(): String {
+        val author = this.author
+        val chNum = this.latestChapter
+        val genres = this.genres
 
         return author + chNum + genres
     }
 
+    private fun mangaRequest(page: Int, order: String): Request = postRequestGraphQL(SEARCH_QUERY(mangaSource, "", "all", order, (page - 1) * 30))
```
AwkwardPeak7: format a bit differently

AwkwardPeak7 (changes requested):
consider taking a look at MangaPark code, see how it is structured in terms of graphql queries, variables, and requests

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
JakeeLuwi: If you're talking about the signature then yes, I believe it is. The old comment is correct, some entries (common in mangas as they have multiple alternative titles) do have duplicates with the title (sometimes the thumbnail also) being the only difference.

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -69,8 +66,26 @@ abstract class MangaHub(
         .add("Sec-Fetch-Site", "same-origin")
         .add("Upgrade-Insecure-Requests", "1")
 
+    private fun graphQLHeader() = headersBuilder()
```
vetleledaal: Why is this a function, as opposed to a variable? If you want to delay computation you could use `by lazy`, but I don't think that is commonly used.

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -170,34 +184,17 @@ abstract class MangaHub(
...
                 }
             }
-        return MangasPage(mangas, doc.select(popularMangaNextPageSelector()).isNotEmpty())
-    }
 
-    override fun popularMangaSelector() = ".col-sm-6:not(:has(a:contains(Yaoi)))"
-
-    override fun popularMangaFromElement(element: Element): SManga {
-        throw UnsupportedOperationException()
+        return MangasPage(mangas, true)
```
vetleledaal: Consider using a named argument, since a bool otherwise tells you nothing.

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -218,158 +215,77 @@ abstract class MangaHub(
...
         return manga
     }
 
-    // chapters
-    override fun chapterListParse(response: Response): List<SChapter> {
-        val document = response.asJsoup()
-        val head = document.head()
-        return document.select(chapterListSelector()).map { chapterFromElement(it, head) }
-    }
+    override fun getMangaUrl(manga: SManga): String = "${baseUrl}${manga.url}"
```
vetleledaal:
```suggestion
    override fun getMangaUrl(manga: SManga): String = "$baseUrl${manga.url}"
```

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -218,158 +215,77 @@ abstract class MangaHub(
...
-        val newHeaders = headersBuilder()
-            .set("Accept", "application/json")
-            .set("Content-Type", "application/json")
-            .set("Origin", baseUrl)
-            .set("Sec-Fetch-Dest", "empty")
-            .set("Sec-Fetch-Mode", "cors")
-            .set("Sec-Fetch-Site", "cross-site")
-            .removeAll("Upgrade-Insecure-Requests")
-            .build()
+        val chapterUrl = chapter.url.split("/")
```
vetleledaal, line 372: Use `HttpUrl.pathSegments`

vetleledaal (dismissed):

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
JakeeLuwi: I haven't refactored the code yet since the implementation is still incomplete. Search does not work and filters needs to be updated. I only committed to save it remotely.

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
JakeeLuwi: By "named argument", do you mean creating a variable (ex. `val getNextPage: Boolean = true`) and passing it in the return (`return MangasPage(mangas, getNextPage)`)?

vetleledaal: I meant like `return MangasPage(mangas, hasNextPage = true)`, although I'm not 100% sure what the argument is named.

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
JakeeLuwi: `chapter.url` is a string. It won't let me do `toHttpUrl()` either since it expects a scheme in which `chapter.url` do not have.

Line 301:
`chapter.url = "/${chapterList.data.manga.slug}/chapter-${it.number}"`

vetleledaal: Normally you would then use:
```kt
"$baseUrl${chapter.url}".toHttpUrl()...
```

but it might make the code harder to read. I'll leave it up to you if you want to change it or leave as-is.

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -90,6 +105,32 @@ abstract class MangaHub(
...
+    private fun graphQLApiInterceptor(chain: Interceptor.Chain): Response {
+        val request = chain.request()
+
+        // We won't intercept non-graphql requests (like image retrieval)
+        if (!request.hasGraphQLTag()) {
+            return chain.proceed(request)
+        }
+
+        val response = chain.proceed(request)
+        val apiResponse = response.parseAs<ApiResponse<Unit>>() // We don't care about the data, only the possible error associated with it
```
AwkwardPeak7: dont consume the response only to resend the request. make a copy

```kt
response.peekBody(Long.MAX_VALUE).string().parseAs<...>()
```

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -170,218 +212,134 @@ abstract class MangaHub(
...
+
+    override fun mangaDetailsParse(response: Response): SManga {
+        val rawManga = response.parseAs<ApiMangaDetailsResponse>()
+
+        return SManga.create().apply {
+            title = rawManga.data!!.manga.title!!
+            author = rawManga.data.manga.author
+            artist = rawManga.data.manga.artist
+            genre = rawManga.data.manga.genres
+            description = rawManga.data.manga.description
```
AwkwardPeak7:
```kt
            description = buildString {
                // add description and alt titles as necessary
            }
```

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -170,218 +212,134 @@ abstract class MangaHub(
...
+            val chapter = SChapter.create()
 
-    override fun chapterFromElement(element: Element): SChapter {
-        throw UnsupportedOperationException()
+            chapter.name = generateChapterName(it.title.trim().replace("\n", " "), it.number)
+            chapter.url = "/${chapterList.data.manga.slug}/chapter-${it.number}"
+            chapter.chapter_number = it.number
+            chapter.date_upload = dateFormat.tryParse(it.date)
+
+            chapter
```
AwkwardPeak7:
```kt
SChapter.create().apply {
    ...
}
```

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHubDto.kt
```diff
@@ -0,0 +1,84 @@
...
+import kotlinx.serialization.Serializable
+
+typealias ApiChapterPagesResponse = ApiResponse<ApiChapterData>
+typealias ApiSearchResponse = ApiResponse<ApiSearchObject>
+typealias ApiMangaDetailsResponse = ApiResponse<ApiMangaObject>
+
+// Base classes
+@Serializable
+class ApiResponse<T>(
+    val data: T?,
```
AwkwardPeak7: don't make nullable since you are checking for errors in interceptor

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHubDto.kt
```diff
@@ -0,0 +1,84 @@
...
+
+@Serializable
+class ApiErrorMessages(
+    val message: String,
+)
+
+// Chapter metadata (pages)
+@Serializable
+class ApiChapterData(
+    val chapter: ApiChapter?,
```
AwkwardPeak7: here too(?)

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHubDto.kt
```diff
@@ -0,0 +1,84 @@
...
+
+@Serializable
+class ApiChapter(
+    val pages: String,
+)
+
+@Serializable
+class ApiChapterPages(
+    val p: String,
+    val i: List<String>,
```
AwkwardPeak7:
```suggestion
    @SerialName("p") val page: String,
    @SerialName("i") val images: List<String>,
```
meaningful names should be used imo

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHubQueries.kt
```diff
@@ -1,42 +1,63 @@
 package eu.kanade.tachiyomi.multisrc.mangahub
 
-import kotlinx.serialization.Serializable
+class GraphQLTag
 
-private fun buildQuery(queryAction: () -> String) = queryAction().replace("%", "$")
-
-val PAGES_QUERY = buildQuery {
+val SEARCH_QUERY = { mangaSource: String, query: String, genre: String, order: String, offset: Int ->
```
AwkwardPeak7: maybe get page number and calculate offset here

AwkwardPeak7: also should make them regular functions instead (avoid SCREAMING_SNAKE_CASE)

AwkwardPeak7 (changes requested):

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHubDto.kt
JakeeLuwi: But during errors, there's still a property of `data` valued `null`. Wouldn't that thing cause deserialization error?

AwkwardPeak7: Use another dto with only error in the interceptor

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHubDto.kt
vetleledaal: Context (resolved): https://github.com/keiyoushi/extensions-source/pull/8392#discussion_r2037770535
> don't make nullable since you are checking for errors in interceptor

JakeeLuwi:
Sorry with the delay regarding this PR. I was waiting for them to finalize their API token refresh system before doing something.

AwkwardPeak7:
you can make another pr  when the token refresh system changes

JakeeLuwi:
> you can make another pr  when the token refresh system changes

Im in the middle of implementing it anyway so its fine. I also don't want to bump the version twice to force them to update twice. This PR title has "rewrite" on it so I guess its still within the scope.

AwkwardPeak7 (approved):

PR #8457, opened 2025-04-13
ManhwaWeb: Filter out invalid pages

bapeey:
Checklist:


src/es/manhwaweb/src/eu/kanade/tachiyomi/extension/es/manhwaweb/ManhwaWeb.kt
```diff
@@ -155,7 +155,7 @@ class ManhwaWeb : HttpSource() {
 
     override fun pageListParse(response: Response): List<Page> {
         val result = json.decodeFromString<PayloadPageDto>(response.body.string())
-        return result.data.images.filter { it.isNotBlank() }
+        return result.data.images.filter { it.isNotBlank() && it.startsWith("http") }
```
vetleledaal: First condition is always true if the second condition is true.
```suggestion
        return result.data.images.filter { it.startsWith("http") }
```

vetleledaal (approved):

PR #8466, opened 2025-04-14
fix(pornpics): properly configure language settings

marioplus:
- Implement language configuration via SourceFactory

Checklist:

### DIFF
--- a/src/all/pornpics/src/eu/kanade/tachiyomi/extension/all/pornpics/PornPicsFactory.kt
+++ b/src/all/pornpics/src/eu/kanade/tachiyomi/extension/all/pornpics/PornPicsFactory.kt
@@ -5,7 +5,6 @@ import eu.kanade.tachiyomi.source.SourceFactory
 class PornPicsFactory : SourceFactory {
 
     override fun createSources() = listOf(
-        PornPics("all"),
         PornPics("en"),
         PornPics("zh"),
     )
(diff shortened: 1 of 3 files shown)


AwkwardPeak7 (commented):
whats the point of factory if the content will be same across all three? generally factory source is used to filter based on language when it comes to entries during browsing or chapters

marioplus:
> whats the point of factory if the content will be same across all three? generally factory source is used to filter based on language when it comes to entries during browsing or chapters

my goal is to pass the lang parameter to Intl. This is the approach I found. If there are more elegant alternatives, please let me know.

AwkwardPeak7:
~~you didn't pass it to intl in this pr~~ nvm I am blind

src/all/pornpics/src/eu/kanade/tachiyomi/extension/all/pornpics/PornPicsFactory.kt
```diff
@@ -0,0 +1,12 @@
+package eu.kanade.tachiyomi.extension.all.pornpics
+
+import eu.kanade.tachiyomi.source.SourceFactory
+
+class PornPicsFactory : SourceFactory {
+
+    override fun createSources() = listOf(
+        PornPics("all"),
```
AwkwardPeak7: if this is only for intl language, then no need for `all`? since translations are for `en` and `zh`. 
you can assign existing `all` id to `en`

```kt
override val id = when (lang) {
    "en" -> <existing id>
    else -> super.id
}
```
get existing id from https://github.com/keiyoushi/extensions/blob/repo/index.json

marioplus: Do I need to bind ID and Lang?

AwkwardPeak7: so in factory, only return en and zh, and in the source class, set en to old all id

AwkwardPeak7:
```suggestion
```

AwkwardPeak7 (approved):

PR #8532, opened 2025-04-18
Webtoons: Add max quality pref

### LINKED ISSUE #5688: Update LINE Webtoon to download original image quality
### Source name

Webtoons.com

### Source language

English

Right now the downloaded chapters are not the best quality available because the image urls end with ".jpg?type=q90".
I also suggest maybe this to be a setting option for this specific extension because the quality change isn't huge but the image size can easely double.

### Other details

Here's a comparison (https://slow.pics/c/cPfdHzwK the eyes here are the most noticeable)

### Source information
Webtoons.com - All 1.4.41

### Acknowledgements

Stratzcha:
Haven't touched much android code before, went the extra safe route for modifying parameters just in case.

Closes #5688 

This is left disabled as default since it, on average, increases the image size by 2x-3x for what is mostly archival quality reasons. The upside is that practically all JPEG artifacting is gone.

Checklist:


src/all/webtoons/src/eu/kanade/tachiyomi/extension/all/webtoons/WebtoonsSrc.kt
```diff
@@ -45,12 +46,44 @@ open class WebtoonsSrc(
...
+    private fun useMaxQualityPref() = preferences.getBoolean(USE_MAX_QUALITY_KEY, false)
 
     override fun pageListParse(document: Document): List<Page> {
-        var pages = document.select("div#_imageList > img").mapIndexed { i, element -> Page(i, "", element.attr("data-url")) }
+        var pages = document.select("div#_imageList > img").mapIndexed { i, element ->
+            val imageUrl = element.attr("data-url")
+
+            if (useMaxQualityPref()) {
+                // Return original if parsing fails
+                val httpUrl = imageUrl.toHttpUrlOrNull() ?: return@mapIndexed Page(i, "", imageUrl)
```
AwkwardPeak7: parsing failing means url isn't valid http url, and that would not load afterwards too, so just use `toHttpUrl()`

src/all/webtoons/src/eu/kanade/tachiyomi/extension/all/webtoons/WebtoonsSrc.kt
```diff
@@ -45,12 +46,43 @@ open class WebtoonsSrc(
...
+                    val newHttpUrl = httpUrl.newBuilder().apply {
+                        removeAllQueryParameters("type")
+                    }.build()
+
+                    return@mapIndexed Page(i, "", newHttpUrl.toString())
+                }
+            }
+
+            Page(i, "", imageUrl)
+        }
```
AwkwardPeak7:
```suggestion
        val useMaxQuality = useMaxQualityPref()
        var pages = document.select("div#_imageList > img").mapIndexed { i, element ->
            val imageUrl = element.attr("data-url").toHttpUrl()

            if (useMaxQuality && imageUrl.queryParameter("type") == "q90") {
                val newImageUrl= imageUrl.newBuilder().apply {
                    removeAllQueryParameters("type")
                }.build()

                Page(i, "", newImageUrl.toString())
            } else {
                Page(i, "", imageUrl.toString())
            }
        }
```

Stratzcha: Did not want to go and parse the URL if it was just going to get reverted back to a string, but I guess the impact is negligible enough.

AwkwardPeak7 (approved):

Issue #8522, opened 2025-04-18
ReadComicOnline issue in getting image links on chapters with more than 99 pages


ojbd:
### Source information

ReadComicOnline 1.4.32

### Source language

English

### Steps to reproduce

Trying to open any chapter that has more than 99 pages, for example here is one with about 250 pages:

![Image](https://github.com/user-attachments/assets/2b6a92a8-aebb-4a7a-a293-5025a5afaa6b)

In order to make sure this was not a different issue the following steps were taken:

1. Cleared chapter and webview cache as well as reloading the entry and clearing the cookies
2. Tried both logged in and logged out in webview
3. Tried changing the dns over https
4. Tried changing servers + image quality
5. Tried on an older version of Tachiyomi
6. Tried migrating the entry and remigrating back
7. Tried on cellular data and wifi

### Expected behavior

Should open the chapter properly

### Actual behavior

Chapter opens, loads for about 10 seconds befoee throwing the error and exiting

### Mihon/Tachiyomi version

0.18.0

### Android version

Android 13

### Other details

May be related to an issue recently opened and closed about a problem getting image links, might be the fix was partial.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to version **[0.15.3](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I have updated all installed extensions.
- [x] I have tried the [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [x] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [x] I will fill out all of the requested information in this form.

### 

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/keiyoushi/extensions-source/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

AwkwardPeak7:
might be fixed in #8536

PR #8536, opened 2025-04-19
RCO: Add workaround for Tachimanga.

lamaxama:
Tachimanga throws an HttpException when the request is not successful and does not implement shouldInterceptRequest and WebResourceResponse.

https://github.com/tachimanga/Tachidesk-Server/blob/d210b349c9e169508ddf6f49fb9e720b759a397f/server/src/main/kotlin/eu/kanade/tachiyomi/network/OkHttpExtensions.kt#L68

Checklist:

### DIFF
--- a/src/en/readcomiconline/build.gradle
+++ b/src/en/readcomiconline/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'ReadComicOnline'
     extClass = '.Readcomiconline'
-    extVersionCode = 32
+    extVersionCode = 33
 }
 
 apply from: "$rootDir/common.gradle"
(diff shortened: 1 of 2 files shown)


AwkwardPeak7:
increase the timeout to 30 seconds, this should fix #8522 (tho need to test)

lamaxama:
> increase the timeout to 30 seconds, this should fix #8522 (tho need to test)

I think it's not because of the timeout, because latch.await(10, TimeUnit.SECONDS) happens before checking if the img is available.

AwkwardPeak7:
that timeout error appears if latch times out, likely due to website not yet finished loading (to call onPageFinish)

lamaxama:
> that timeout error appears if latch times out, likely due to website not yet finished loading (to call onPageFinish)

We use loadDataWithBaseURL instead of loadUrl, it's supposed to be fast.

lamaxama:
I just tested on simulator and couldn't replicate this issue https://github.com/keiyoushi/extensions-source/issues/8522

AwkwardPeak7:
webview will not call onPageFinish until website finishes loading it's resources and finishes its execution (loading circle stop on a regular browser).
besides increased timeout won't hurt

lamaxama:
> webview will not call onPageFinish until website finishes loading it's resources and finishes its execution (loading circle stop on a regular browser). besides increased timeout won't hurt

Alright, I just don't think it's due to the timeout.

AwkwardPeak7:
hey doesn't hurt to try

AwkwardPeak7 (approved):

PR #8544, opened 2025-04-19
Toonily: various fixes

AwkwardPeak7:
- change `mangaSubString` -> serie
  - prevent 302 redirect
- disable count views
  - 400 bad request
- use LoadMoreRequest
- fetch hd covers if possible

Checklist:

### DIFF
--- a/src/en/toonily/src/eu/kanade/tachiyomi/extension/en/toonily/Toonily.kt
+++ b/src/en/toonily/src/eu/kanade/tachiyomi/extension/en/toonily/Toonily.kt
@@ -3,10 +3,10 @@ package eu.kanade.tachiyomi.extension.en.toonily
 import eu.kanade.tachiyomi.lib.cookieinterceptor.CookieInterceptor
 import eu.kanade.tachiyomi.multisrc.madara.Madara
 import eu.kanade.tachiyomi.network.GET
-import eu.kanade.tachiyomi.source.model.FilterList
-import okhttp3.HttpUrl.Companion.toHttpUrl
-import okhttp3.OkHttpClient
+import eu.kanade.tachiyomi.source.model.SManga
+import okhttp3.Interceptor
 import okhttp3.Request
+import okhttp3.Response
 import java.text.SimpleDateFormat
 import java.util.Locale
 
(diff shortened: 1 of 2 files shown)


src/en/toonily/src/eu/kanade/tachiyomi/extension/en/toonily/Toonily.kt
```diff
@@ -17,63 +17,71 @@ class Toonily : Madara(
...
         return GET("$baseUrl/search/?post_type=wp-manga", headers)
     }
 
-    // The source customized the Madara theme and broke the filter.
-    override val filterNonMangaItems = false
-
-    override val useNewChapterEndpoint: Boolean = true
+    override fun mangaDetailsRequest(manga: SManga): Request {
+        val newManga = SManga.create().apply {
+            url = manga.url.replace("/webtoon/", "/$mangaSubString/")
```
vetleledaal, line 48: I guess this is to avoid migration? You will run into the other issue where browsing will show library items as non-library items. Usually I think the tradeoff here is worth it.

If you want to address it further you would do the opposite- always use the old structure and rewrite it in the interceptor. Downside here is that you are stuck with the old URL format forever.

src/en/toonily/src/eu/kanade/tachiyomi/extension/en/toonily/Toonily.kt
```diff
@@ -17,63 +18,80 @@ class Toonily : Madara(
...
+            }
+        } else {
+            chain.proceed(request)
+        }
+    }
+
     companion object {
         val titleSpecialCharactersRegex = "[^a-z0-9]+".toRegex()
-        val trailingHyphenRegex = "-+$".toRegex()
+        val sdCoverRegex = Regex("""(-\d+x\d+)(\.\w+)${'$'}""")
```
vetleledaal: Why make it more complicated than necessary?
```suggestion
        val sdCoverRegex = Regex("""(-\d+x\d+)(\.\w+)$""")
```

vetleledaal: First capture group is never used. Requires changing other code as well:
```suggestion
        val sdCoverRegex = Regex("""(?:-\d+x\d+)(\.\w+)${'$'}""")
```

vetleledaal: Consider only matching Arabic numerals (super nit-picky):
```suggestion
        val sdCoverRegex = Regex("""(-[0-9]+x[0-9]+)(\.\w+)${'$'}""")
```


Even more nit-picky, only allow natural numbers (probably don't actually do this, adds too much complexity):
```suggestion
        val sdCoverRegex = Regex("""(-(?:0|[1-9][1-9]*)x(?:0|[1-9][1-9]*))(\.\w+)${'$'}""")
```

src/en/toonily/src/eu/kanade/tachiyomi/extension/en/toonily/Toonily.kt
AwkwardPeak7: this is just meant to prevent redirects

vetleledaal (approved):

Issue #8514, opened 2025-04-17
RCO: Some issues fail to load


Emily-123-67:
### Source information

ReadComicOnline 1.4.32

### Source language

English

### Steps to reproduce

In my case i found this bug with a comic called Mind MGMT
1.If you go to issue 13 it doesn't load
2.If you open webview on the issue it says "error 400"
3. This doesnt happen with all the issues, for example issue 29 works just fine

![Image](https://github.com/user-attachments/assets/d7020504-c73b-44e7-a6c2-6bd3f759e5f4)

![Image](https://github.com/user-attachments/assets/69e8f832-9cec-4a09-8a3e-2ce20d69129b)

### Expected behavior

All the issues should open and load correctly

### Actual behavior

Some of them do not load correctly 

### Mihon/Tachiyomi version

0.18.0

### Android version

Android 13

### Other details

This is like the 3rd issue in like 4 days with this extension lmao

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to version **[0.15.3](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I have updated all installed extensions.
- [x] I have tried the [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [x] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [x] I will fill out all of the requested information in this form.

### 

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/keiyoushi/extensions-source/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

AwkwardPeak7:
loads up fine, try clearing chapter cache, webview cache, and change server/quality in extension settings

![Image](https://github.com/user-attachments/assets/77c96598-05a0-4d8d-9d80-cb95017fdf6d)

Emily-123-67:
For some reason, server 2 works but not server 1, which i think it still sounds like a bug.
Server 1 just some pages load but not all of them which i dont actually understand how this even happens

onorino:
No matter what you do it’s broken again.

[...]

PR #8560, opened 2025-04-20
RCO: dedupe URLs, sort by most likely URL

vetleledaal:
This will produce less network requests, and improves performance slightly by using a stricter pre-filter for arrays. On some older chapters I have observed 1 miss before it found the correct set of URLs, which adds ~1s. Does not address #8514.

Removing `break;` to make it more resilient to counter-measures.

This PR is _NOT_ necessary for RCO to work.

Checklist:

### DIFF
--- a/src/en/readcomiconline/assets/script.js
+++ b/src/en/readcomiconline/assets/script.js
@@ -1,56 +1,82 @@
 (() => {
-    const isValidUrl = (str) => {
+    const customProps = (() => {
+        const iframe = document.createElement('iframe');
+        document.body.appendChild(iframe);
+
+        const standardProps = new Set(Object.getOwnPropertyNames(iframe.contentWindow));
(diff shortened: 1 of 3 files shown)


MikeZeDev:
May want to try .... catch the filtering of the arrays. 

Ive got errors on window[k]   while testing in browser. Perhaps they dynamically remove properties or whatever
```js

 customProps.forEach(k => {
     try {
         const obj = window[k];
         if (
             Array.isArray(obj) &&
             obj.length !== 0 &&
             obj.every(el => typeof el === 'string') &&
             obj.every(el => el.length > 8)
         ) {
             multiMaybeUrls.push(obj);
         } else if (
             typeof obj === 'function' &&
             obj.length >= 1
         ) {
             multiMaybeFn.push(obj);
         }
     } catch {}

 });
```

PR #8652, opened 2025-04-26
MyHentaiGallery: minor cleanup, fix URLs

vetleledaal:
Some URLs were just wrong, some were redirecting, some were missing headers.

Prefers `Document.selectFirst` over `Document.select` where applicable.  

Description is now bullet-ified.

If a library item was added with a deeplink it will now use the same URL as if you browse to the item (previously `/gallery/thumbnails/2`, now `/g/2`). Do we want to automatically fix these URLs for existing users? That might cause issues if the user has both items in the library. Could notify the user, but should be actionable in that case (migrate to force duplicate check?). Could just ignore it and hope for the best.

Checklist:


src/en/myhentaigallery/src/eu/kanade/tachiyomi/extension/en/myhentaigallery/MyHentaiGallery.kt
```diff
@@ -107,17 +111,26 @@ class MyHentaiGallery : ParsedHttpSource() {
...
-                description += info.select("div:containsOwn(parodies) a").let { groups ->
-                    if (groups.isNotEmpty()) "Parodies: ${groups.joinToString { it.text() }}" else ""
+                artist = info.select("div:containsOwn(artists) a").joinToString { it.text() }
+                thumbnail_url = document.selectFirst(".comic-listing .comic-inner img")?.absUrl("src")
+                description = buildString {
+                    info.select("div:containsOwn(groups) a")
+                        .takeIf { it.isNotEmpty() }
+                        ?.also { if (isNotEmpty()) append("\n\n") }
+                        ?.also { appendLine("Groups:") }
+                        ?.joinToString("\n") { "• ${it.text()}" }
```
AwkwardPeak7: consider using markdown bullets (-) as main app now support markdown (at least in preview)

AwkwardPeak7 (approved):

PR #8659, opened 2025-04-27
MangaHub: try to refresh api key for all api requests

AwkwardPeak7:
currently, manga details and chapter list fail when no cookies are set, as api key is not fetched/refreshed there.

Checklist:

### DIFF
--- a/lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
+++ b/lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
@@ -1,6 +1,7 @@
 package eu.kanade.tachiyomi.multisrc.mangahub
 
 import android.content.SharedPreferences
+import android.util.Log
 import androidx.preference.PreferenceScreen
 import androidx.preference.SwitchPreferenceCompat
 import eu.kanade.tachiyomi.network.GET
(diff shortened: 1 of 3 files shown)


lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -182,39 +184,72 @@ abstract class MangaHub(
...
+            .onErrorResumeNext { error ->
+                if (errorRegex.containsMatchIn(error.message ?: "")) {
+                    refreshApiKey(manga, chapter)
+                    fetch()
+                } else {
+                    Observable.error(error)
+                }
+            }
+    }
+    private val errorRegex = Regex("""(api)?\s*rate\s*limit\s*(excessed)?|api\s*key\s*(invalid)?""")
```
vetleledaal: As used these parts are useless. Should be equivalent to:
```suggestion
    private val errorRegex = Regex("""\s*rate\s*limit\s*|api\s*key\s*""")
```

lib-multisrc/mangahub/src/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -235,6 +270,9 @@ abstract class MangaHub(
     }
 
     // popular
+    override fun fetchPopularManga(page: Int): Observable<MangasPage> =
+        fetchAndRetry { super.fetchPopularManga(page) }
```
vetleledaal: Would it make more sense to add this to an interceptor?

vetleledaal (approved):

PR #8716, opened 2025-05-02
Contributing guide - use backup links

vetleledaal:
Preferably the contributing guidelines should refer to Mihon instead of Tachiyomi, or be app-agnostic. Until that is done we can at least fix the existing dead links.

Checklist:


FourTOne5:
For now we should probably refer to Mihon instead of Tachiyomi similar to what @AwkwardPeak7 did for website. Being app-agnostic can come later.

PR #8754, opened 2025-05-05
zh/BoyLove: Fix chapter list parsing and enable upload dates display in chapters

morallkat:
This PR updates the code for parsing the chapter list API in the Boylove source, which was changed after the site CMS upgrade a few days ago and returned 400 errors when retrieving the chapters:
```
! Unexpected JSON token at offset 19: Expected start of the object '{', but had '[' instead at path: $.result
JSON input: {"code":1,"result":[],"msg":"","succ":400}
```
A reverse order function has been added as the updated API's JSON now only returns chapters in positive order. The new API also includes upload dates for each chapter, this PR complements the related strings and functions to display them.

Checklist:

### DIFF
--- a/src/zh/boylove/src/eu/kanade/tachiyomi/extension/zh/boylove/BoyLoveDto.kt
+++ b/src/zh/boylove/src/eu/kanade/tachiyomi/extension/zh/boylove/BoyLoveDto.kt
@@ -2,6 +2,7 @@ package eu.kanade.tachiyomi.extension.zh.boylove
 
 import eu.kanade.tachiyomi.source.model.SChapter
 import eu.kanade.tachiyomi.source.model.SManga
+import keiyoushi.utils.tryParse
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.JsonPrimitive
 import kotlinx.serialization.json.long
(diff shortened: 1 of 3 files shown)


src/zh/boylove/src/eu/kanade/tachiyomi/extension/zh/boylove/BoyLoveDto.kt
```diff
@@ -56,10 +56,12 @@ fun String.toImageUrl() =
 class ChapterDto(
     private val id: Int,
     private val title: String,
+    private val create_time: String,
 ) {
     fun toSChapter() = SChapter.create().apply {
         url = "/home/book/capter/id/$id"
         name = title.trim()
+        date_upload = dateFormat.parse(create_time)!!.time
```
AwkwardPeak7:
```suggestion
        date_upload = dateFormat.tryParse(create_time)
```
need to import it too

AwkwardPeak7 (dismissed):

src/zh/boylove/src/eu/kanade/tachiyomi/extension/zh/boylove/BoyLoveDto.kt
morallkat: fixed.

vetleledaal (approved):

PR #8883, opened 2025-05-17
readded homepage support

### LINKED ISSUE #8882: ReadAllComics: Add Support For Homepage Titles
### Source name

ReadAllComics

### Source language

English

Source re-added recently updated titles to the homepage. The extension should be changed so it can parse those titles on the recently updated page and display them. It's not exactly a popular titles page, but it's close.

### Other details

_No response_

### Acknowledgements

### 

---

WyattRoss:
Checklist:

Closes #8882

### DIFF
--- a/src/en/readallcomicscom/src/eu/kanade/tachiyomi/extension/en/readallcomicscom/ReadAllComics.kt
+++ b/src/en/readallcomicscom/src/eu/kanade/tachiyomi/extension/en/readallcomicscom/ReadAllComics.kt
@@ -32,16 +32,28 @@ class ReadAllComics : ParsedHttpSource() {
     override val client = network.cloudflareClient
 
     override fun popularMangaRequest(page: Int): Request {
-        throw Exception("ReadAllComics has no popular titles Page. Please use the search function instead.")
+        val url = baseUrl.toHttpUrl().newBuilder().apply {
+            addPathSegments("page/$page")
+        }.build()
+
+        return GET(url, headers)
     }
 
-    // Never called
     override fun popularMangaFromElement(element: Element): SManga {
-        throw Exception("")
+        val manga = SManga.create().apply {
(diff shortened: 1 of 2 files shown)


src/en/readallcomicscom/src/eu/kanade/tachiyomi/extension/en/readallcomicscom/ReadAllComics.kt
```diff
@@ -32,16 +32,28 @@ class ReadAllComics : ParsedHttpSource() {
...
-    // Never called
     override fun popularMangaFromElement(element: Element): SManga {
-        throw Exception("")
+        val manga = SManga.create().apply {
+            val category = element.classNames()
+                .firstOrNull { it.startsWith("category-") }!!
+                .substringAfter("category-")
+            setUrlWithoutDomain("/category/$category")
+            title = category.replace("-", " ").titleCaseWords()
+            thumbnail_url = element.selectFirst("img")?.attr("src")
```
WyattRoss, line 49: The title won't be exactly correct in all cases (occasionally capitalization will be wrong, or a colon will be missing), and the thumbnail will be of the most recently uploaded chapter (unless either piece of data is cached). This is because of the way the source displays such data on their front page (linking to chapters rather than chapter lists).

It is possible to work around this, but the best solution I found required making a new request for every entry and grabbing information from the chapter list page. This was very slow. I looked into an async solution but couldn't get it working, if such a thing is possible it would probably be optimal though.

AwkwardPeak7: extension was using a similar approach previously too, so it's fine.

and we try to avoid making n calls for n manga on browse to not strain source servers

AwkwardPeak7 (approved):

PR #8923, opened 2025-05-20
Adds the ability to automatically filter chapters in comick based on "score"

tejassharma96:
The comick api includes an `up_count` and a `down_count`, this PR uses those values to automatically filter uploaders based on which release for each chapter has a higher "score" (ie `up_count - down_count`) and only displays the one with the highest score.

I'm not a kotlin dev but it's fairly straightforward logic and I have built and tested this and it works as expected.

Checklist:

- (n/a) Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- (n/a) Referenced all related issues in the PR body (e.g. "Closes #xyz")
- (n/a) Added the `isNsfw = true` flag in `build.gradle` when appropriate
- (n/a) Have explicitly kept the `id` if a source's name or language were changed
- (n/a) Have removed `web_hi_res_512.png` when adding a new extension


src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
```diff
@@ -546,6 +563,20 @@ abstract class Comick(
 
                 publishedChapter && noGroupBlock
             }
+
+        if (preferences.chapterScoreFiltering) {
```
AwkwardPeak7:
```kt
chapterListResponse.chapters
    .filter { 
        \\ ...
    }
    .groupBy { it.chap }
    .mapValues { (_, chapters) -> chapters.maxBy { it.score } }
    .values
    .map { it.toSChapter(mangaUrl) }
```
consider something like this,
add a public `score` value to `Chapter` dto:
```kt
@Serializable
class Chapter(
    \\ ..
) {
    val score get() = upCount - downCount
    
    fun toSChapter() = \\ ...
}
```

src/all/comickfun/assets/i18n/messages_en.properties
```diff
@@ -22,3 +22,6 @@ score_position_top=Top
 score_position_middle=Middle
 score_position_bottom=Bottom
 score_position_none=Hide Score
+chapter_score_filtering_title=Filter scanlators automatically based on chapter score
```
AwkwardPeak7:
```suggestion
chapter_score_filtering_title=Automatically de-duplicate chapters
```

AwkwardPeak7 (changes requested):

src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
tejassharma96: thanks for the suggestion! that certainly is a bit cleaner, my lack of kotlin experience coming to bite me 😅 

since the motivation appears to be going back to returning one fluent chain - how would I handle the check for `preferences.chapterScoreFiltering` and only do the bit after the `filter` in that case?

currently what I have is I'm still declaring `var chapters = chapterListResponse.chapters.filter { }`

and then I have
```
if (preferences.chapterScoreFiltering) {
    chapters = chapters.groupBy { it.chap }
        .mapValues { (_, chapters) -> chapters.maxBy { it.score } }
        .values
        .toList()
}

return chapters
    .map { it.toSChapter(mangaUrl) }
```

is that ok? I have some idea on how I could just make this one chain in swift (I'm an iOS developer by trade) by creating an extension on arrays where the array item is of type Chapter, but not sure how that would look in kotlin. alternatively I could just leave it as I have it above since maybe I'm overcomplicating it

tejassharma96: ok I figured out how to make an extension and used that instead to have a fluent

```
return chapterListResponse.chapters
    .filter {
        // existing code
    }
    .filterOnScore(preferences.chapterScoreFiltering)
    .map { it.toSChapter(mangaUrl) }
```

src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
```diff
@@ -546,9 +563,20 @@ abstract class Comick(
...
             }
+            .filterOnScore(preferences.chapterScoreFiltering)
             .map { it.toSChapter(mangaUrl) }
     }
 
+    private fun List<Chapter>.filterOnScore(shouldFilter: Boolean): Collection<Chapter> {
+        if (shouldFilter) {
+            return groupBy { it.chap }
+                .mapValues { (_, chapters) -> chapters.maxBy { it.score } }
+                .values
```
bapeey:
```suggestion
                .map { (_, chapters) -> chapters.maxBy { it.score } }
```

tejassharma96: updated, thanks!

AwkwardPeak7 (approved):

PR #8962, opened 2025-05-25
Add YYmanhua source

hualiong:
Checklist:


src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,151 @@
...
+    override fun imageUrlParse(response: Response): String {
+        val (_, pix, pvalue) = Regex("var pix=\"(.*?)\".*?var pvalue=\\[\"(.*?)\"")
+            .find(decode(response.body.string()))?.groupValues!!
+        return pix + pvalue
+    }
+
+    override fun imageUrlParse(document: Document) = throw UnsupportedOperationException()
+
+    private fun decode(ciphertext: String): String {
+        val groups = Regex("return p;\\}\\('(.*?)',(\\d+),(\\d+),'(.*?)'")
```
vetleledaal: Extract regex into class/companion. For more performance.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,151 @@
...
+
+    override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
+        return GET(
+            baseUrl +
+                if (query.isNotBlank()) {
+                    "/search?title=$query&page=$page"
+                } else {
+                    "/manga-list-${filters[1]}-${filters[2]}-${filters[3]}-p$page/"
+                },
+        )
```
vetleledaal: Missing headers.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,151 @@
...
+
+    // Search Page
+
+    override fun getFilterList() = buildFilterList()
+
+    override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
+        return GET(
+            baseUrl +
+                if (query.isNotBlank()) {
+                    "/search?title=$query&page=$page"
```
vetleledaal: Need to encode `query`. Use `HttpUrlBuilder`.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,151 @@
...
+
+    // override fun imageRequest(page: Page) = GET(page.url, headers)
+
+    override fun imageUrlParse(response: Response): String {
+        val (_, pix, pvalue) = Regex("var pix=\"(.*?)\".*?var pvalue=\\[\"(.*?)\"")
+            .find(decode(response.body.string()))?.groupValues!!
+        return pix + pvalue
+    }
+
+    override fun imageUrlParse(document: Document) = throw UnsupportedOperationException()
```
vetleledaal, line 134: Is this a well-known algorithm, or something bespoke to just this site?

hualiong: This decode method is specific to this website, because it has encrypted JavaScript code.

src/zh/yymanhua/build.gradle
```diff
@@ -0,0 +1,8 @@
+ext {
+    extName = 'YYmanhua'
+    extClass = '.YYmanhua'
+    extVersionCode = 2
```
vetleledaal:
```suggestion
    extVersionCode = 1
```

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,160 @@
...
+
+    override val baseUrl = "https://www.yymanhua.com"
+    override val lang = "zh"
+    override val name = "YY漫画"
+    override val supportsLatest = true
+
+    companion object {
+        val DESC_REGEX = Regex("\\[\\+展开]|\\[-折叠]")
+        val CHAPTER_REGEX = Regex("第(\\d+(?:\\.\\d+)?)[话話]")
+        val NUM_REGIX = Regex("\\d+")
```
vetleledaal:
```suggestion
        val NUM_REGEX = Regex("\\d+")
```

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,160 @@
...
+    override fun chapterFromElement(element: Element) = SChapter.create().apply {
+        url = element.attr("href")
+        name = element.text()
+        chapter_number = CHAPTER_REGEX.find(name)?.groups?.get(1)?.value?.toFloat() ?: -0F
+    }
+
+    // Manga View Page
+
+    override fun pageListParse(document: Document): List<Page> {
+        val cid = NUM_REGIX.find(document.location())?.groups?.get(0)?.value
```
vetleledaal:
```suggestion
        val cid = NUM_REGEX.find(document.location())?.groups?.get(0)?.value
```

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,160 @@
...
+        val CHAPTER_REGEX = Regex("第(\\d+(?:\\.\\d+)?)[话話]")
+        val NUM_REGIX = Regex("\\d+")
+        val IMG_REGEX = Regex("var pix=\"(.*?)\".*?var pvalue=\\[\"(.*?)\"")
+        val DECODE_REGEX1 = Regex("return p;\\}\\('(.*?)',(\\d+),(\\d+),'(.*?)'")
+        val DECODE_REGEX2 = Regex("\\b\\w+\\b")
+    }
+
+    override fun headersBuilder() = super.headersBuilder()
+        .set("User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36 Edg/136.0.0.0")
+        .add("Referer", baseUrl)
```
vetleledaal:
```suggestion
        .add("Referer", "$baseUrl/")
```

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,160 @@
...
+        .build()
+
+    // Popular Page
+
+    override fun popularMangaRequest(page: Int) = GET("$baseUrl/manga-list-p$page", headers)
+
+    override fun popularMangaSelector() = ".mh-item"
+
+    override fun popularMangaFromElement(element: Element) = SManga.create().apply {
+        title = element.select(".mh-item-detali .title a").text()
```
vetleledaal: You only expect one title element, 
```suggestion
        title = element.selectFirst(".mh-item-detali .title a").text()
```

Also review all other uses of `.select`, and if they can be replaced with `.selectFirst`.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,160 @@
...
+        element.select("a:nth-child(1)").let {
+            thumbnail_url = it.select("img").attr("src")
+            url = it.attr("href")
```
vetleledaal:
```kt
setUrlWithoutDomain(it.absUrl("href"))
```

(might need non-null assert)

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,160 @@
...
+    // Popular Page
+
+    override fun popularMangaRequest(page: Int) = GET("$baseUrl/manga-list-p$page", headers)
+
+    override fun popularMangaSelector() = ".mh-item"
+
+    override fun popularMangaFromElement(element: Element) = SManga.create().apply {
+        title = element.select(".mh-item-detali .title a").text()
+        element.select("a:nth-child(1)").let {
+            thumbnail_url = it.select("img").attr("src")
```
vetleledaal: Use abs URL here.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
vetleledaal: This still needs to be resolved.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,160 @@
...
+    override fun searchMangaNextPageSelector() = popularMangaNextPageSelector()
+
+    // Manga Detail Page
+
+    override fun mangaDetailsParse(document: Document) = SManga.create().apply {
+        description = document.select(".detail-info-content").text()
+            .replace(DESC_REGEX, "").trim()
+        val els = document.select(".detail-info-tip > span")
+        els[0].select("a").let {
+            author = it[0].text()
```
vetleledaal, line 92: Do you know if the 0th element will always exist? Need to guard against errors here.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,160 @@
...
+            artist = it.getOrNull(1)?.text()
+        }
+        status = when (els[1].select("span")[1].text()) {
+            "连载中", "連載中" -> SManga.ONGOING
+            "已完结", "已完結" -> SManga.COMPLETED
+            else -> SManga.UNKNOWN
+        }
+        genre = els[2].select(".item").joinToString { it.text() }
```
vetleledaal, line 100: No size check here too.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,160 @@
...
+        initialized = true
+    }
+
+    // Manga Detail Page / Chapters Page (Separate)
+
+    override fun chapterListRequest(manga: SManga) = GET(baseUrl + manga.url, headers)
```
vetleledaal: Useless override. This is the default.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,160 @@
...
+
+    override fun chapterListSelector() = "#chapterlistload a"
+
+    override fun chapterFromElement(element: Element) = SChapter.create().apply {
+        url = element.attr("href")
+        name = element.text()
+        chapter_number = CHAPTER_REGEX.find(name)?.groups?.get(1)?.value?.toFloat() ?: -0F
```
vetleledaal: `-0F` seems wrong. I think only `-1F` and `-2F` had special meaning. Please double check.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    // Manga Detail Page / Chapters Page (Separate)
+
+    override fun chapterListRequest(manga: SManga) = GET(baseUrl + manga.url, headers)
+
+    override fun chapterListSelector() = "#chapterlistload a"
+
+    override fun chapterFromElement(element: Element) = SChapter.create().apply {
+        url = element.attr("href")
```
vetleledaal: Use `setUrlWithoutDomain()` and `.absUrl()`.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
hualiong: This is a fixed style for these three elements on the page, even if there is no content, the tags will always exist, it is secure.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
hualiong: Same as above

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
hualiong: The href attribute is originally a relative path without domain. And this is a fixed format in this site
There's no need to use these two methods.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
hualiong: The href attribute is originally a relative path without domain. And this is a fixed format in this site
I think there's no need to use these two methods.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
vetleledaal: It's not required, but it makes it more robust. The site happens to use a relative path starting with a slash, but if they change it it will also mess up the extension, even if everything else stays the same. For example, if the site gives you a href of `../abc/def.html` you need to know the current document path to calculate the actual path.

By using `absUrl` you get Jsoup to calculate the proper path. By using `setUrlWithoutdomain()` you don't have to care about how to strip the scheme/domain.

hualiong: Alright, I'll change it later.

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    override fun headersBuilder() = super.headersBuilder()
+        .set("User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36 Edg/136.0.0.0")
+        .add("Referer", "$baseUrl/")
+        .add("Cookie", "yymanhua_lang=2")
+
+    override val client = network.cloudflareClient.newBuilder()
+        .connectTimeout(15, TimeUnit.SECONDS)
+        .readTimeout(15, TimeUnit.SECONDS)
```
AwkwardPeak7: app defaults to 30 seconds, so imo not needed?

src/zh/yymanhua/src/eu/kanade/tachiyomi/extension/zh/yymanhua/YYmanhua.kt
```diff
@@ -0,0 +1,157 @@
...
+        val DESC_REGEX = Regex("\\[\\+展开]|\\[-折叠]")
+        val CHAPTER_REGEX = Regex("第(\\d+(?:\\.\\d+)?)[话話]")
+        val NUM_REGEX = Regex("\\d+")
+        val IMG_REGEX = Regex("var pix=\"(.*?)\".*?var pvalue=\\[\"(.*?)\"")
+        val DECODE_REGEX1 = Regex("return p;\\}\\('(.*?)',(\\d+),(\\d+),'(.*?)'")
+        val DECODE_REGEX2 = Regex("\\b\\w+\\b")
+    }
+
+    override fun headersBuilder() = super.headersBuilder()
+        .set("User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36 Edg/136.0.0.0")
```
AwkwardPeak7, line 32: does site not work with mobile or other specific useragents? because hard coding useragent is generally not a good idea

hualiong: This site has different page structures on different devices. If using mobile user-agent, the extension may not be able to find the element. I don't want to write a redundant set of code for the mobile devices, it's unnecessary.

vetleledaal (approved):

PR #9000, opened 2025-05-28
VortexScans: actually get latest chapters

lbrooney:
Their homepage uses a different endpoint that can get posts by most recent update. JSON returned is the same as `query` endpoint. I believe `posts` accepts search terms, but not filters. 

On their website they get 48 per page when using this endpoint, however the parse function uses 18 so I used that value instead. `perPage` in the Iken multisrc is private and I wasn't sure whether I should make it public, so instead I just gave Vortex its own private value.

Edit: It appears both `Nyx Scans` and `Magus Manga` would benefit from this change as well. For the the other implementations of Iken, the `query` endpoint already seems to sort them by latest and while they do have a posts endpoint, if requested to with the query `tag=latestUpdate` causes them to only return the same page of results regardless of the page number (Hive, Aurora, MangaPro). It seems the pattern is that sources with their API on a subdomain support the query param.

Checklist:

### DIFF
--- a/src/en/arvenscans/build.gradle
+++ b/src/en/arvenscans/build.gradle
@@ -2,7 +2,7 @@ ext {
     extName = 'Vortex Scans'
     extClass = '.VortexScans'
     themePkg = 'iken'
-    overrideVersionCode = 37
+    overrideVersionCode = 38
     isNsfw = false
 }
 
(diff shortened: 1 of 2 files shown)


AwkwardPeak7:
closing in favor of #9002

PR #9003, opened 2025-05-29
SMBC: Handle 500 "error" on archive page

raxod502:
Checklist:

See the added comment which explains this change. I haven't been able to update SMBC since late February; with the modified extension from this PR branch, it again works fine.

### DIFF
--- a/src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
+++ b/src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
@@ -55,7 +55,6 @@ class Hiveworks : ParsedHttpSource() {
             if (response.code == 500) {
                 val newResponse = response.newBuilder()
                     .code(200)
-                    .body(response.body)
                     .build()
                 newResponse
             } else {
@@ -201,7 +200,6 @@ class Hiveworks : ParsedHttpSource() {
     // Included to call custom error codes
     override fun fetchChapterList(manga: SManga): Observable<List<SChapter>> {
         return if (manga.status != SManga.LICENSED) {
-            val uri = Uri.parse(manga.url).buildUpon()
             client.newCall(chapterListRequest(manga))
                 .asObservableSuccess()
                 .map { response ->
--- a/src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
+++ b/src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
@@ -39,6 +39,29 @@ class Hiveworks : ParsedHttpSource() {
         .readTimeout(1, TimeUnit.MINUTES)
         .retryOnConnectionFailure(true)
         .followRedirects(true)
+        .addNetworkInterceptor { chain ->
+            val request = chain.request()
+            if (!request.url.toString().contains("smbc-comics")) {
+                return@addNetworkInterceptor chain.proceed(request)
+            }
+
+            val response = chain.proceed(request)
+            // As of March 2025, SMBC chapter list page returns status code 500 even
+            // though it still has correct data. Do not throw an error in this case.
+            //
+            // I reported this error to SMBC on 2025-05-28 and it was not fixed by
+            // 2025-06-11, but even if it is fixed eventually, the same problem might
+            // occur again in the future.
+            if (response.code == 500) {
+                val newResponse = response.newBuilder()
+                    .code(200)
+                    .body(response.body)
+                    .build()
+                newResponse
+            } else {
+                response
+            }
+        }
         .build()
 
     // Popular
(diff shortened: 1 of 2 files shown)


src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
```diff
@@ -178,8 +178,27 @@ class Hiveworks : ParsedHttpSource() {
     // Included to call custom error codes
     override fun fetchChapterList(manga: SManga): Observable<List<SChapter>> {
         return if (manga.status != SManga.LICENSED) {
+            val uri = Uri.parse(manga.url).buildUpon()
             client.newCall(chapterListRequest(manga))
-                .asObservableSuccess()
+                .asObservable()
+                .doOnNext { response ->
+                    if (!response.isSuccessful) {
```
vetleledaal: Should implement this check in an interceptor. See:
https://github.com/keiyoushi/extensions-source/blob/d7d84b3419b6c6e0fca2420957319323eecc41c4/src/en/xoxocomics/src/eu/kanade/tachiyomi/extension/en/xoxocomics/XoxoComics.kt#L27-L42

Sites tend to error on multiple pages, not just one.

src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
```diff
@@ -178,8 +178,27 @@ class Hiveworks : ParsedHttpSource() {
...
+                            // status code 500 even though it still has correct data.
+                            // Do not throw an error in this case.
+                            //
+                            // I reported this error to SMBC on 2025-05-28, but even if
+                            // it is fixed, the same problem might occur again in the future.
+                            return@doOnNext
+                        }
+                        // Otherwise, an unsuccessful response status indicates that
+                        // there is an error. Apparently, HttpException isn't able to
+                        // be imported from extensions-lib, so we have to copy and paste.
```
vetleledaal: I don't think this context (comments) is needed.

src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
```diff
@@ -178,8 +178,27 @@ class Hiveworks : ParsedHttpSource() {
...
-                .asObservableSuccess()
+                .asObservable()
+                .doOnNext { response ->
+                    if (!response.isSuccessful) {
+                        if ("smbc-comics" in uri.toString() && response.code == 500) {
+                            // As of March 2025, SMBC chapter list page returns
+                            // status code 500 even though it still has correct data.
+                            // Do not throw an error in this case.
+                            //
+                            // I reported this error to SMBC on 2025-05-28, but even if
```
vetleledaal: Should wait a week and see if they fix it.

src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
raxod502: Updated. Thank you for this pointer. I did not find this pattern, although I would be happy to start writing an extension code style guide / common patterns list for the documentation, if that would be helpful.

src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
raxod502: Removed.

raxod502: If you meant to remove both comments, and not just the second one, let me know and I can do that.

src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
raxod502: Waited and have not heard back nor has the issue been fixed, so I have updated the implementation here with your comments and have re-tested.

src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
```diff
@@ -39,6 +39,29 @@ class Hiveworks : ParsedHttpSource() {
...
+            // As of March 2025, SMBC chapter list page returns status code 500 even
+            // though it still has correct data. Do not throw an error in this case.
+            //
+            // I reported this error to SMBC on 2025-05-28 and it was not fixed by
+            // 2025-06-11, but even if it is fixed eventually, the same problem might
+            // occur again in the future.
+            if (response.code == 500) {
+                val newResponse = response.newBuilder()
+                    .code(200)
+                    .body(response.body)
```
AwkwardPeak7:
```suggestion
```
I could be wrong but just changing the response code shouldn't change the body, so this is not needed

src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
```diff
@@ -178,6 +201,7 @@ class Hiveworks : ParsedHttpSource() {
     // Included to call custom error codes
     override fun fetchChapterList(manga: SManga): Observable<List<SChapter>> {
         return if (manga.status != SManga.LICENSED) {
+            val uri = Uri.parse(manga.url).buildUpon()
```
AwkwardPeak7: unused?
tho upon looking again, this method override isn't changing anything from the implementation in super class, so should remove it.

src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
raxod502: I tested and you're right. Removed.

src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
raxod502: Removed the accidental leftover variable. However, I think the method override is still needed, it adds the conditional that reports an error when the manga is licensed (some Hiveworks series listed in the index are available on web while others are not).

FourTOne5 (approved):

PR #9106, opened 2025-06-05
Add MayoTune

kyrie25:
Checklist:


src/en/mayotune/src/eu/kanade/tachiyomi/extension/en/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,16 @@
+package eu.kanade.tachiyomi.extension.en.mayotune
+
+import eu.kanade.tachiyomi.multisrc.mayotune.MayoTune
+import eu.kanade.tachiyomi.source.model.SManga
+
+class MayoTune : MayoTune("MayoTune", "https://mayotune.xyz/", "en") {
```
vetleledaal: `baseUrl` should not include trailing `/`.

src/en/mayotune/src/eu/kanade/tachiyomi/extension/en/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,16 @@
...
+    override val sourceList = listOf(
+        SManga.create().apply {
+            title = "Mayonaka Heart Tune"
+            url = baseUrl
+            thumbnail_url = "${baseUrl}img/cover.jpg"
+            author = "Masakuni Igarashi"
+            artist = "Masakuni Igarashi"
```
vetleledaal: This is implied/hidden if the author is set to the same IIRC
```suggestion
```

lib-multisrc/mayotune/src/eu/kanade/tachiyomi/multisrc/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    override fun pageListRequest(chapter: SChapter): Request {
+        return GET(chapter.url, headers)
+    }
+
+    // Details
+
+    override fun mangaDetailsParse(document: Document): SManga = SManga.create().apply {
+        val statusTxt = document.select("div.text-center").find {
```
vetleledaal: Don't need to save 1 character here.
```suggestion
        val statusText = document.select("div.text-center").find {
```

lib-multisrc/mayotune/src/eu/kanade/tachiyomi/multisrc/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,162 @@
...
+            it.text().contains("Status")
+        }?.text()?.substringBefore("Status")?.trim()
```
vetleledaal:
```suggestion
        val statusTxt = document.select("div.text-center:contains(Status)")?.text()?.substringBefore("Status")?.trim()
```

lib-multisrc/mayotune/src/eu/kanade/tachiyomi/multisrc/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,162 @@
...
+        status = when (statusTxt) {
+            "Ongoing" -> SManga.ONGOING
+            "Completed" -> SManga.COMPLETED
+            "Cancelled" -> SManga.CANCELLED
+            "Hiatus" -> SManga.ON_HIATUS
+            "Finished" -> SManga.PUBLISHING_FINISHED
+            else -> SManga.UNKNOWN
+        }
+        thumbnail_url = if (document.select(".object-contain").attr("src").isNotEmpty()) {
+            baseUrl + document.select(".object-contain").attr("src")
```
vetleledaal: Use absolute URLs for this
```suggestion
        thumbnail_url = if (document.select(".object-contain").absUrl("src").isNotEmpty()) {
            baseUrl + document.select(".object-contain").absUrl("src")
```

lib-multisrc/mayotune/src/eu/kanade/tachiyomi/multisrc/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,162 @@
...
+    override fun mangaDetailsParse(document: Document): SManga = SManga.create().apply {
+        val statusTxt = document.select("div.text-center").find {
+            it.text().contains("Status")
+        }?.text()?.substringBefore("Status")?.trim()
+
+        url = sourceList.first().url
+        title = sourceList.first().title
+        artist = sourceList.first().artist
+        author = sourceList.first().author
+        description = document.select(".text-lg").text()
```
vetleledaal: Prefer using `.selectFirst` where you only expect to match one element.

lib-multisrc/mayotune/src/eu/kanade/tachiyomi/multisrc/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,162 @@
...
+    override fun chapterFromElement(element: Element): SChapter =
+        throw UnsupportedOperationException()
+
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val jsonObj = json.decodeFromString<JsonArray>(response.body.string())
+        return jsonObj.sortedByDescending {
+            it.jsonObject["number"]?.jsonPrimitive?.float
+        }.map { json ->
+            SChapter.create().apply {
+                url = "${baseUrl}chapter/${json.jsonObject["id"]?.jsonPrimitive?.contentOrNull}"
```
vetleledaal: Consider using a DTO instead.

lib-multisrc/mayotune/src/eu/kanade/tachiyomi/multisrc/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,162 @@
...
+                name =
+                    if (!json.jsonObject["title"]?.jsonPrimitive?.contentOrNull.isNullOrEmpty()) {
+                        "Chapter ${json.jsonObject["number"]?.jsonPrimitive?.contentOrNull}: " + json.jsonObject["title"]?.jsonPrimitive?.contentOrNull
+                    } else {
+                        "Chapter ${json.jsonObject["number"]?.jsonPrimitive?.contentOrNull}"
+                    }
+                chapter_number = json.jsonObject["number"]?.jsonPrimitive?.float ?: 0f
+                date_upload = json.jsonObject["date"]?.jsonPrimitive?.contentOrNull?.let {
+                    var sdf = SimpleDateFormat("yyyy-MM-dd", Locale.US)
```
vetleledaal: Extract `SimpleDateFormat` to class / companion object.

lib-multisrc/mayotune/src/eu/kanade/tachiyomi/multisrc/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,162 @@
...
+                    sdf.timeZone = TimeZone.getDefault()
```
vetleledaal: This timezone relies on user-specific settings and should be avoided. If you don't know the correct timezone you should just remove it.

lib-multisrc/mayotune/src/eu/kanade/tachiyomi/multisrc/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    // Pages
+
+    override fun pageListParse(document: Document): List<Page> =
+        document.select("div.w-full > img").mapIndexed { index, img ->
+            Page(
+                index,
+                "",
+                baseUrl + img.attr("src"),
```
vetleledaal:
```suggestion
                imageUrl = img.absUrl("src"),
```

lib-multisrc/mayotune/src/eu/kanade/tachiyomi/multisrc/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,139 @@
...
+import uy.kohesive.injekt.injectLazy
+import java.text.SimpleDateFormat
+import java.util.Locale
+import kotlin.getValue
+
+abstract class MayoTune(
+    override val name: String,
+    override val baseUrl: String,
+    override val lang: String,
+) : ParsedHttpSource() {
```
AwkwardPeak7: you are basically not using most of `ParsedHttpSource` methods, consider directly implementing `HttpSource`

lib-multisrc/mayotune/src/eu/kanade/tachiyomi/multisrc/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,139 @@
...
+    override fun mangaDetailsParse(document: Document): SManga = SManga.create().apply {
+        val statusText =
+            document.selectFirst("div.text-center:contains(Status)")?.text()
+                ?.substringBefore("Status")
+                ?.trim()
+
+        url = sourceList.first().url
+        title = sourceList.first().title
+        artist = sourceList.first().artist
+        author = sourceList.first().author
```
AwkwardPeak7: why always first?

lib-multisrc/mayotune/src/eu/kanade/tachiyomi/multisrc/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,139 @@
...
+            ?.ifEmpty { sourceList.first().thumbnail_url }
+    }
+    // Chapters
+
+    override fun chapterListSelector() = throw UnsupportedOperationException()
+    override fun chapterFromElement(element: Element): SChapter =
+        throw UnsupportedOperationException()
+
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val chapters = json.decodeFromString<List<ChapterDto>>(response.body.string())
```
AwkwardPeak7:
```suggestion
        val chapters = response.parseAs<List<ChapterDto>>()
```
need to import, also remove the json instance declared near the top

lib-multisrc/mayotune/src/eu/kanade/tachiyomi/multisrc/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,139 @@
...
+        return chapters.sortedByDescending { it.number }.map { chapter ->
+            SChapter.create().apply {
+                url = chapter.getChapterURL(baseUrl)
```
AwkwardPeak7: save relative urls, or just the id

lib-multisrc/mayotune/src/eu/kanade/tachiyomi/multisrc/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,139 @@
...
+                name = chapter.getChapterTitle()
+                chapter_number = chapter.number
+                date_upload = chapter.date.let { sdf.parse(it)?.time } ?: 0L
```
AwkwardPeak7:
```suggestion
                date_upload = sdf.tryParse(chapter.date)
```
need to import

src/en/mayotune/build.gradle
```diff
@@ -0,0 +1,10 @@
+ext {
+    extName = 'MayoTune'
+    extClass = '.MayoTune'
+    themePkg = 'mayotune'
+    baseUrl = 'https://mayotune.xyz/'
```
AwkwardPeak7: no trailing slash

src/en/mayotune/src/eu/kanade/tachiyomi/extension/en/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,15 @@
+package eu.kanade.tachiyomi.extension.en.mayotune
+
+import eu.kanade.tachiyomi.multisrc.mayotune.MayoTune
+import eu.kanade.tachiyomi.source.model.SManga
+
+class MayoTune : MayoTune("MayoTune", "https://mayotune.xyz", "en") {
```
AwkwardPeak7: why make a multisrc if there is only one source in it?

AwkwardPeak7 (changes requested):

src/en/mayotune/src/eu/kanade/tachiyomi/extension/en/mayotune/MayoTune.kt
kyrie25: My bad, there wasn't any documents that I could find for development and it is my first time making an extension so I followed the semantics based on what I could find in the source code.
Could you point me to the relevant documents so I can make the appropriate changes?

AwkwardPeak7: See the contributing guide. You can ask any questions in our discord

But basically, remove the multisrc and implement this class as httpsource directly. See another existing source for reference.

kyrie25: I've updated the changes correspondingly, it should meet your criterias now 👌

src/en/mayotune/src/eu/kanade/tachiyomi/extension/en/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,129 @@
...
+
+class MayoTune() : HttpSource() {
+    override val name: String = "MayoTune"
+    override val baseUrl: String = "https://mayotune.xyz"
+    override val lang: String = "en"
+    override val versionId: Int = 1
+
+    private val source = SManga.create().apply {
+        title = "Mayonaka Heart Tune"
+        url = baseUrl
```
vetleledaal: This should also be relative, as with chapter URLs. IIRC it makes it easier to migrate to a different domain for existing users.
```suggestion
        url = "/"
```

Needs other changes.

vetleledaal (dismissed):

src/en/mayotune/src/eu/kanade/tachiyomi/extension/en/mayotune/MayoTune.kt
kyrie25: I tried to figure this one out before. 
Without override `mangaDetailsRequest`, relative URLs gets resolved into `baseUrl`, so the address becomes `<baseUrl><baseUrl>`. Atm I have no clue how to do it without overriding.

kyrie25: Nvm just changing the URL was enough, should be good now

src/en/mayotune/src/eu/kanade/tachiyomi/extension/en/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,128 @@
...
+        throw UnsupportedOperationException()
+
+    // Get Override
+    override fun mangaDetailsRequest(manga: SManga): Request {
+        return GET(manga.url, headers)
+    }
+
+    override fun chapterListRequest(manga: SManga): Request {
+        return GET(manga.url + "/api/chapters", headers)
+    }
```
vetleledaal: Don't you also need to change this so it handles the relative URL?
```suggestion
    override fun chapterListRequest(manga: SManga): Request {
        return GET("$baseUrl${manga.url}api/chapters", headers)
    }
```

kyrie25: Apparently relative URLs in the overwritten method gets concatenated with `baseUrl`, and not a replacement of it. 
Removing the manga details override breaks the extension for the same reason as it tries to resolve into absolute URLs and concats them anyway(?)

I've tested it myself before committing and it's fine.

kyrie25: > Removing the manga details override breaks the extension for the same reason as it tries to resolve into absolute URLs and concats them anyway(?)

This is what I was talking about:

vetleledaal: Right, but that might be a leftover from the entry being in the database already. From a fresh state I would imagine it acts differently.

kyrie25: The cache may be there at launch, but the error toast will show up as soon as you refresh the entry if you decide to remove the override methods. Endpoints are part of the extension code, not the cache database, is it not?

kyrie25: I tested the changes on a fresh AVD and indeed it did not concatentate `baseURL` by default. I don't get *why* that behaviour is cached and changes were ignored but you were right, my apologies.

src/en/mayotune/src/eu/kanade/tachiyomi/extension/en/mayotune/MayoTune.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    override fun searchMangaRequest(page: Int, query: String, filters: FilterList) =
+        throw UnsupportedOperationException()
+
+    override fun searchMangaParse(response: Response): MangasPage =
+        throw UnsupportedOperationException()
+
+    // Get Override
+    override fun chapterListRequest(manga: SManga): Request {
```
AwkwardPeak7, line 72: need to override `getChapterUrl` to provide webview url for chapter (webview button in reader)

kyrie25: That part works fine without needing to override though, since it still behaves normally with relative URLs
image

AwkwardPeak7: no the `getChapterUrl` method of `HttpSource`, it should return website url instead of api url, which you return in `chapterListRequest`

AwkwardPeak7: if you open the reader in app, there is a webview button, this method is for that

kyrie25: > if you open the reader in app, there is a webview button, this method is for that

That is how I got there, and it is behaving as expected on a fresh installation/AVD. Is it still required?

https://github.com/user-attachments/assets/eecf2de4-0e21-4ca3-87be-67b56eddf411

AwkwardPeak7: ah sorry, I thought that method uses the `chapterListRequest`, it actually uses `pageListRequest`...  mb

AwkwardPeak7 (approved):

PR #9175, opened 2025-06-10
NhatTruyen update domain & fix missing chapter & fix search

dejavui:
Checklist:

### DIFF
--- a/src/vi/nhattruyen/src/eu/kanade/tachiyomi/extension/vi/nhattruyen/NhatTruyen.kt
+++ b/src/vi/nhattruyen/src/eu/kanade/tachiyomi/extension/vi/nhattruyen/NhatTruyen.kt
@@ -9,15 +9,13 @@ import eu.kanade.tachiyomi.source.model.MangasPage
 import eu.kanade.tachiyomi.source.model.SChapter
 import eu.kanade.tachiyomi.source.model.SManga
 import eu.kanade.tachiyomi.util.asJsoup
+import keiyoushi.utils.parseAs
 import keiyoushi.utils.tryParse
-import kotlinx.serialization.decodeFromString
-import kotlinx.serialization.json.Json
 import okhttp3.HttpUrl.Companion.toHttpUrl
 import okhttp3.Request
 import okhttp3.Response
 import org.jsoup.nodes.Document
 import rx.Observable
-import uy.kohesive.injekt.injectLazy
 import java.text.SimpleDateFormat
 import java.util.Locale
 
@@ -29,8 +27,6 @@ class NhatTruyen : WPComics(
     gmtOffset = null,
 ) {
     override val searchPath = "tim-truyen"
-
-    private val json: Json by injectLazy()
 
     /**
      * NetTruyen/NhatTruyen redirect back to catalog page if searching query is not found.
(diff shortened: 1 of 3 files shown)


src/vi/nhattruyen/src/eu/kanade/tachiyomi/extension/vi/nhattruyen/NhatTruyen.kt
```diff
@@ -177,4 +187,32 @@ class NhatTruyen : WPComics(
...
+                .asObservableSuccess()
+                .map { response -> chapterListParse(response) }
+        }
+        return super.fetchChapterList(manga)
+    }
+
+    private fun checkChapterLists(document: Document) = document.selectFirst("a.view-more.hidden")!!.text()
+
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val json = json.decodeFromString<ChapterDTO>(response.body.string())
```
AwkwardPeak7:
```suggestion
        val json = response.parseAs<ChapterDTO>()
```
need to import, and remove the json by injectLazy above too

src/vi/nhattruyen/src/eu/kanade/tachiyomi/extension/vi/nhattruyen/NhatTruyen.kt
```diff
@@ -177,4 +187,32 @@ class NhatTruyen : WPComics(
...
+    override fun fetchChapterList(manga: SManga): Observable<List<SChapter>> {
+        val url = client.newCall(GET(baseUrl + manga.url, headers)).execute().asJsoup()
+        if (checkChapterLists(url).isNotEmpty()) {
+            val slug = manga.url.removePrefix("/truyen-tranh/")
+            return client.newCall(GET("$baseUrl/Comic/Services/ComicService.asmx/ChapterList?slug=$slug", headers))
+                .asObservableSuccess()
+                .map { response -> chapterListParse(response) }
+        }
+        return super.fetchChapterList(manga)
+    }
```
AwkwardPeak7: just fetch from the ajax call all the time, instead of checking for view-more. and you return super method but you changed chapterlistparse for json parsing, so will it work?

AwkwardPeak7 (changes requested):

src/vi/nhattruyen/src/eu/kanade/tachiyomi/extension/vi/nhattruyen/NhatTruyen.kt
dejavui: > just fetch from the ajax call all the time, instead of checking for view-more. and you return super method but you changed chapterlistparse for json parsing, so will it work?

It works. it’s better for the web to make the call only when there’s manga with many chapters.

AwkwardPeak7: you have changed `chapterListParse` so now there is no logic for html parsing, which you are relying on in `super.fetchChapterList`

dejavui: > you have changed `chapterListParse` so now there is no logic for html parsing, which you are relying on in `super.fetchChapterList`

I changed it to chapterlistRequest. Now it's complete, ready to merge 👍

AwkwardPeak7 (approved):

PR #9208, opened 2025-06-13
MangaDex: Add setting to include unavailable chapters in chapter list

FlaminSarge:
The purpose of this is to assist with migration to another source (if a chapter is hidden from chapter list its read/unread state does not get migrated even if it exists in the DB somewhere, it seems) and allow users with downloaded chapters to be able to see/read/delete them.

I believe the chapter url is used as the 'source of truth' in order to rename downloaded chapters if the title or scanlators change, so including UNAVAILABLE in the scanlator text should pose no issue (unless the user already pulled the chapter list with the unavailable chapters hidden, which may or may not remove them from the DB, and if they were removed from the DB the downloads may have already gotten orphaned).

When a user tries to view an unavailable chapter they get the standard "404 not found" toast.

<details>
<summary>Screenshots</summary>

image
image

</details>

Checklist:


FlaminSarge:
The boolean is nullable as I'm not sure whether MD sends it for every chapter (they should, but...) and it's better to be safe about it.

This setting could potentially be expanded to also include External chapters in the listing (isInvalid chapters) but I don't think that's in scope for the purposes of this PR.

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
```diff
@@ -546,6 +547,7 @@ abstract class MangaDex(final override val lang: String, private val dexLang: St
 
         return chapterListResults
             .filterNot { it.attributes!!.isInvalid }
+            .filter { it.attributes!!.isUnavailable != true || preferences.includeUnavailable }
```
FlaminSarge: This line is _probably_ not necessary but is defensive, I guess.

FourTOne5: remove it

FlaminSarge: _If_ it were to be kept, does preferences.includeUnavailable need to be moved outside the filter so it doesn't fetch prefs repeatedly? Though as you said probs just better to remove.

AwkwardPeak7 (approved):

PR #9270, opened 2025-06-15
BacaKomik: update domain, skip popular/latest redirect

vetleledaal:
Was looking into the redirects on search. For example `?genre[]=action` redirects to `?genre[0]=action`. Minor issue, not addressed in this PR. Otherwise on any other page than 1 it also does not like empty query values. These can be normalized locally with:
```kt
if (page > 1) {
    // Removes `=` for empty query parameters. Skips a redirect. 
    val httpUrl = url.build()
    for (queryParameterName in httpUrl.queryParameterNames) {
        if (httpUrl.queryParameter(queryParameterName) == "") {
            url.setQueryParameter(queryParameterName, null)
        }
    }
}
```

Not changing in this PR since that doesn't fully solve the problem.

Checklist:

### DIFF
--- a/src/id/bacakomik/src/eu/kanade/tachiyomi/extension/id/bacakomik/BacaKomik.kt
+++ b/src/id/bacakomik/src/eu/kanade/tachiyomi/extension/id/bacakomik/BacaKomik.kt
@@ -20,7 +20,7 @@ import java.util.Locale
 
 class BacaKomik : ParsedHttpSource() {
     override val name = "BacaKomik"
-    override val baseUrl = "https://bacakomik.one"
+    override val baseUrl = "https://bacakomik.my"
     override val lang = "id"
     override val supportsLatest = true
     private val dateFormat: SimpleDateFormat = SimpleDateFormat("MMM d, yyyy", Locale.US)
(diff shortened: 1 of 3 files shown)


src/id/bacakomik/src/eu/kanade/tachiyomi/extension/id/bacakomik/BacaKomik.kt
```diff
@@ -66,7 +68,6 @@ class BacaKomik : ParsedHttpSource() {
         val builtUrl = if (page == 1) "$baseUrl/daftar-komik/" else "$baseUrl/daftar-komik/page/$page/?order="
         val url = builtUrl.toHttpUrl().newBuilder()
         url.addQueryParameter("title", query)
-        url.addQueryParameter("page", page.toString())
```
AwkwardPeak7, line 69: was this intentionally removed from search request?

vetleledaal: Yes, the page is already in the path and I noticed this query argument was useless. 

Didn't mention it in the commit message as it was a relatively minor change. Probably looks out of place with the other changes in this function being reverted.

AwkwardPeak7 (approved):

PR #9457, opened 2025-06-27
Feat/pixiv deeplink

### LINKED ISSUE #9452: Pixiv: open links in app
### Source name

Pixiv

### Source language

all

Some extensions register an Intent-Filter for generic `VIEW` intents pointing to the source's relevant website. This allows opening any shared manga (chapter/page) in the browser, and jumping straight to Mihon to bookmark it. This feature should also be added to the Pixiv source, especially because the search sucks a lot and finding the same illustration that you *literally* have open already is sometimes near impossible.

### Other details

I'm willing to contribute this feature if I find the time, but I can't guarantee stellar quality as I am still learning Android development. The implementation from the `all/mangadex` extension could be used as a blueprint.

After slightly deeper digging, it seems the primary way to open any manga in Mihon using a link (and the one MangaDex uses) is to craft a special search query which produces just the single result. Pixiv doesn't currently have this capability, and adding it would change the behavior of its search; by extending it with the ability to search for specific IDs, which is great, but still a potentially breaking(?) change.

### Acknowledgements

### 

---

ilonachan:
Closes #9452.

The propsed implementation adds three new dummy search handlers:

- `aid:12345` for direct deeplinks to a specific illustration
- `sid:12345` for direct deeplinks to a specific series (both of these will be treated as "manga", and an illustration within the series points to the same manga as the series itself)
- `user:12345` to filter for only the illustrations by a specific user. This differs from the corresponding filter in that
  1. it is accessible to a deeplink
  2. it uses the user ID instead of the name
  3. it therefore works with users whose names contain spaces, unlike the filter variant

I chose these patterns so they hopefully won't be accidentally encountered by a user. I decided to make the last one more speaking, because it is potentially useful for regular searching (though its functionality could be expanded by allowing it to appear beside regular search terms). All of these decisions can obviously be debated, and would ideally be approved by someone who is more primarily responsible for this plugin.

It seems that the entire search mechanism of this plugin could benefit from a general overhaul (for example, to prevent the constant `IllegalArgumentException`s when "violating the [nonexistent] contract" by retrying a search). But without changing more than necessary, this is the best I can do so far.

Checklist:

### DIFF
--- a/src/all/pixiv/AndroidManifest.xml
+++ b/src/all/pixiv/AndroidManifest.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application>
+        <activity
+            android:name=".all.pixiv.PixivUrlActivity"
+            android:excludeFromRecents="true"
+            android:exported="true"
+            android:theme="@android:style/Theme.NoDisplay">
+            <intent-filter android:autoVerify="true">
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data android:host="pixiv.net" />
+                <data android:host="www.pixiv.net" />
+                <data android:scheme="http" />
+                <data android:scheme="https" />
+
+                <data android:pathPattern="/en/artworks/..*" />
+                <data android:pathPattern="/artworks/..*" />
+                <data android:pathPattern="/en/users/..*" />
+                <data android:pathPattern="/users/..*" />
+                <data android:pathPattern="/user/..*/series/..*" />
+            </intent-filter>
+        </activity>
+    </application>
+</manifest>
--- a/src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
+++ b/src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
@@ -1,8 +1,5 @@
 package eu.kanade.tachiyomi.extension.all.pixiv
 
-import eu.kanade.tachiyomi.extension.all.pixiv.PixivConstants.ID_ILLUST_PREFIX
-import eu.kanade.tachiyomi.extension.all.pixiv.PixivConstants.ID_SERIES_PREFIX
-import eu.kanade.tachiyomi.extension.all.pixiv.PixivConstants.ID_USER_PREFIX
 import eu.kanade.tachiyomi.source.model.FilterList
 import eu.kanade.tachiyomi.source.model.MangasPage
 import eu.kanade.tachiyomi.source.model.Page
@@ -12,6 +9,7 @@ import eu.kanade.tachiyomi.source.online.HttpSource
 import eu.kanade.tachiyomi.util.asJsoup
 import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.decodeFromJsonElement
 import okhttp3.Headers
 import okhttp3.HttpUrl
 import okhttp3.HttpUrl.Companion.toHttpUrl
@@ -41,14 +39,26 @@ class Pixiv(override val lang: String) : HttpSource() {
             client.newCall(request.url(url.build()).build()).execute()
     }
 
+    class PixivApiException(message: String? = null) : Exception(message, null)
+
     private inner class ApiCall(href: String?) : HttpCall(href) {
         init {
             url.addEncodedQueryParameter("lang", lang)
             request.addHeader("Accept", "application/json")
         }
 
-        inline fun <reified T> executeApi(): T =
-            json.decodeFromString<PixivApiResponse<T>>(execute().body.string()).body!!
+        /**
+         * Sends the previously constructed API call to the Pixiv API.
+         * If the server reports an error, A [PixivApiException] will be
+         * returned as a [Result.failure].
+         */
+        inline fun <reified T> executeApi(): Result<T> {
+            val resp = json.decodeFromString<PixivApiResponse>(execute().body.string())
+            if (resp.error) {
+                return Result.failure(PixivApiException(resp.message))
+            }
+            return Result.success(json.decodeFromJsonElement<T>(resp.body!!))
+        }
     }
 
     private var popularMangaNextPage = 1
@@ -62,13 +72,13 @@ class Pixiv(override val lang: String) : HttpSource() {
                 for (p in countUp(start = 1)) {
                     call.url.setEncodedQueryParameter("page", p.toString())
 
-                    val entries = call.executeApi<PixivRankings>().ranking!!
+                    val entries = call.executeApi<PixivRankings>().getOrThrow().ranking!!
                     if (entries.isEmpty()) break
 
                     val call = ApiCall("/touch/ajax/illust/details/many")
                     entries.forEach { call.url.addEncodedQueryParameter("illust_ids[]", it.illustId!!) }
 
-                    call.executeApi<PixivIllustsDetails>().illust_details!!.forEach { yield(it) }
+                    call.executeApi<PixivIllustsDetails>().getOrThrow().illust_details!!.forEach { yield(it) }
                 }
             }
                 .toSManga()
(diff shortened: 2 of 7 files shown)


ilonachan:
An alternative approach, which I've only just thought about: the deeplink service could (after basic validation) just pass the URL straight into the query. In fact, pasting a URL into Mihon search is a feature that I feel should be expected to work, even though it normally doesn't. Since a URL is about as different as can get from any search term, this would completely avoid the accidental triggering. (the main reason why I didn't implement it this way is that I was copying Mangadex, which used search handlers, and so I guess I sorta tunnel-visioned)

I believe this new idea should *definitely* be implemented, as it is a straight improvement (at least for the direct deeplinks to single manga chapters/series). The introduction of more sophisticated search features should perhaps be the subject of a different issue/PR. Thoughts?

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -85,6 +88,25 @@ class Pixiv(override val lang: String) : HttpSource() {
     private lateinit var searchIterator: Iterator<SManga>
 
     override fun fetchSearchManga(page: Int, query: String, filters: FilterList): Observable<MangasPage> {
+        // Deeplink selection of specific IDs: simply fetch the single object and return
+        if (query.startsWith(ID_ILLUST_PREFIX) &&
+            query.removePrefix(ID_ILLUST_PREFIX).matches("\\d+".toRegex())
+        ) {
```
AwkwardPeak7:
```suggestion
        if (query.startsWith(ID_ILLUST_PREFIX)) {
```
no need to check, since anything else will result in 404 or similar from pixiv

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -85,6 +88,25 @@ class Pixiv(override val lang: String) : HttpSource() {
...
+            val illustId = query.removePrefix(ID_ILLUST_PREFIX)
+            val manga = getIllustCached(illustId).toSManga()
+            return Observable.just(MangasPage(listOf(manga), hasNextPage = false))
+        }
+        if (query.startsWith(ID_SERIES_PREFIX) &&
+            query.removePrefix(ID_SERIES_PREFIX).matches("\\d+".toRegex())
+        ) {
```
AwkwardPeak7: same here

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -95,19 +117,35 @@ class Pixiv(override val lang: String) : HttpSource() {
...
-                    dateBefore = filters.dateBefore.ifBlank { null },
-                    dateAfter = filters.dateAfter.ifBlank { null },
-                )
-
-                predicates = buildList {
-                    filters.makeTagsPredicate()?.let(::add)
-                    filters.makeUsersPredicate()?.let(::add)
+                // TODO: it would be useful to allow multiple user: tags in the query
+                if (query.startsWith(ID_USER_PREFIX) &&
+                    query.removePrefix(ID_USER_PREFIX).matches("\\d+".toRegex())
```
AwkwardPeak7: same here

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/PixivConstants.kt
```diff
@@ -0,0 +1,7 @@
+package eu.kanade.tachiyomi.extension.all.pixiv
+
+object PixivConstants {
```
AwkwardPeak7: no need for separate object, global constants should be fine

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/PixivUrlActivity.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        var couldBuildIntent = false
+
+        do {
+            var pathSegments = intent?.data?.pathSegments ?: break
+
+            if ("en".equals(pathSegments[0])) {
```
AwkwardPeak7: en should not be hardcoded? since pixiv supports other languages

src/all/pixiv/AndroidManifest.xml
```diff
@@ -0,0 +1,29 @@
...
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data android:host="pixiv.net" />
+                <data android:host="www.pixiv.net" />
+                <data android:scheme="http" />
+                <data android:scheme="https" />
+
+                <data android:pathPattern="/en/artworks/..*" />
```
AwkwardPeak7, line 21: en shouldn't be strict requirement? since pixiv supports other languages

AwkwardPeak7 (changes requested):

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
ilonachan: The reason I added that check is so regular queries that just happen to look like `sid:dharta or sth` aren't captured by this mechanism. Idea being that the user either knows about this special handling, or must be *extremely* unlucky to trigger it. But if you don't think it's important to go that far, yeah we can skip the check.

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/PixivUrlActivity.kt
ilonachan: I would've thought so too, but every language code I've tried just gives a 404 (even "jp"). So for now I assumed that "en" is special for some weird reason.

But on the off chance that there are other codes that would work (or you or someone else knows them already), what should I do instead? Make a whitelist of those codes, or just allow trying any known language code? (this might balloon the `intent-filter` list tho.) Or another wildcard? (that might complicate matching logic a bit.)

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/PixivConstants.kt
ilonachan: Oh yeah right this is Kotlin :rofl: My own first instinct is that having constants just... be there, floating around in the package, feels kinda weird. For super generally useful ones maybe, but the instinct to put a static holder class around them is strong. But I imagine that's not the kotlin-esque way to do it, so sure I can get rid of the wrapper.

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
ilonachan: I've decided (for now) to move completely to the URL-based system, meaning this isn't really relevant anymore. I've left the code for parsing and handling the special search queries in, including this check, because especially the way they're now conceptually set up there needs to be *some* validation. But it's currently not doing anything anyway, and the exact way it's gonna be used/useful/changed/deleted in the future is completely up in the air for now.

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/PixivUrlActivity.kt
ilonachan: For now I've gone with a `KNOWN_LOCALES` constant, a list with (currently) only one entry. I don't understand the entirety of Pixiv's URL system, so until someone with more concrete knowledge can clarify things this is probably the best solution for now.

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
ilonachan: (see [above](https://github.com/keiyoushi/extensions-source/pull/9457#discussion_r2173485971))

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
ilonachan: (see [above](https://github.com/keiyoushi/extensions-source/pull/9457#discussion_r2173485971))

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/PixivLinkUtil.kt
```diff
@@ -0,0 +1,101 @@
...
+        if (!id.matches("\\d+".toRegex())) return null
+        return fromSearchQueryId(id)
+    }
+
+    fun fromSearchQueryId(id: String): T?
+}
+
+sealed class PixivTarget {
+    companion object {
+        val BASE_URI = Uri.parse("https://www.pixiv.net")
```
AwkwardPeak7: consider using okhttp's `HttpUrl` instead of `Uri`

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/PixivLinkUtil.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    data class Series(val seriesId: String, val authorUserId: String? = null) : PixivTarget() {
+        companion object : PixivTargetCompanion<Series> {
+            override val SEARCH_PREFIX = "sid:"
+            override fun fromSearchQueryId(id: String) = Series(id)
+        }
+
+        override fun toUri(): Uri {
+            if (authorUserId == null) {
+                TODO("Not yet implemented")
```
AwkwardPeak7: throw regular exception, cuz iirc this can crash the app

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/PixivLinkUtil.kt
ilonachan: I mostly used `Uri` because of Android compatibility, the intent gives us a `Uri` after all. But maybe `HttpUrl` should be the default, and `from(uri: Uri)` could just be an overload. I assume the only difference is that okhttp actually does validation?

AwkwardPeak7 (approved):

PR #9651, opened 2025-07-14
TempleScan(es): Auto update domain

bapeey:
Checklist:


src/es/templescanesp/src/eu/kanade/tachiyomi/extension/es/templescanesp/TempleScanEsp.kt
```diff
@@ -31,6 +33,20 @@ class TempleScanEsp :
         else -> preferences.prefBaseUrl
     }
 
+    private val fetchedDomainUrl: String by lazy {
+        if (!preferences.fetchDomainPref()) return@lazy preferences.prefBaseUrl
+        try {
+            val initClient = network.cloudflareClient
+            val headers = super.headersBuilder().build()
+            val host = initClient.newCall(GET(preferences.prefBaseUrl, headers)).execute().request.url.host
```
AwkwardPeak7: set `followRedirect` to false for the client, and get new url from `Location` header (3xx response)

AwkwardPeak7 (dismissed):

src/es/templescanesp/src/eu/kanade/tachiyomi/extension/es/templescanesp/TempleScanEsp.kt
stevenyomi, line 1: Remove the `isCi` variable and its related logic. The CI workflow can properly handle preferences.

This pattern is also misused in the repo. I wrote it to handle dynamic pulling URL from network on CI build, not for skipping preferences.

src/es/templescanesp/src/eu/kanade/tachiyomi/extension/es/templescanesp/TempleScanEsp.kt
```diff
@@ -31,6 +34,24 @@ class TempleScanEsp :
...
 
+    private val fetchedDomainUrl: String by lazy {
+        if (!preferences.fetchDomainPref()) return@lazy preferences.prefBaseUrl
+        try {
+            val initClient = network.cloudflareClient.newBuilder()
+                .followRedirects(false)
+                .build()
+            val headers = super.headersBuilder().build()
+            val response = initClient.newCall(GET(preferences.prefBaseUrl, headers)).execute()
+            if (!response.isRedirect) return@lazy preferences.prefBaseUrl
```
stevenyomi, line 40: Since the site seems to redirect requests to old domain to the new domain, you can, instead, add an interceptor to the client to inspect redirected requests and get the new domain transparently.

```kotlin
override var client = buildClient()

private fun buildClient() = super.client.newBuilder()
    .rateLimitHost(baseUrl.toHttpUrl(), 3, 1)
    .addInterceptor { chain ->
        val request = chain.request()
        if (
            response.priorResponse != null &&
            request.url.toString().startsWith(preferences.prefBaseUrl) &&
            request.url.pathSegments == response.request.url.pathSegments
        ) {
            preferences.prefBaseUrl = "https://${response.request.url.host}"
            client = buildClient()
        }
        response
    }
    .build()
```

stevenyomi (commented):
Searching for `private val fetchedDomainUrl` yields several results in the repo; it's a weird pattern which we might want to check later.

src/es/templescanesp/src/eu/kanade/tachiyomi/extension/es/templescanesp/TempleScanEsp.kt
bapeey: I prefer to check only the site's homepage at startup, due to cases where the site redirects to others for reading

stevenyomi: What do you mean by "redirects to others for reading"?

bapeey: To another domain, like [this](https://visorikigai.skteq.xyz/series/el-sobreviviente-mas-fuerte-del-mundo/), maybe its very niche

stevenyomi: You mean redirecting to a third-party website. The code I provided actually checks if the `pathSegments` match after redirection. However, you can say there's still possibility that the third-party URL may have the same path. A mitigation might be checking if the host contains "templescan" but that's not guaranteed to be correct either.

The tradeoff is, your code only run once and introduces delay on the first network request. The interceptor approach can capture redirections at any time - even if the user kept the app running in background and never restarts it.

Feel free to resolve the conversation if you think it's better to be safe and correct.

stevenyomi (approved):

PR #9656, opened 2025-07-14
baozimhorg: fix illegal State Exception & more

### LINKED ISSUE #9543: GoDa (zh.baozimhorg) - illegal State Exception
### Source information

GoDa (zh.baozimhorg) - 1.4.33

### Source language

Chinese

### Steps to reproduce

1. search any manga
2. it will show "illegalStateException: Check failed."
3. you can always press refresh to reproduce this issue as well

### Expected behavior

able to refresh without any error message

### Actual behavior

This happened when try to go any manga chapter page.

### Mihon/Tachiyomi version

Stable 0.18.0 (2025/3/10 06:40)

### Android version

Andriod 15

### Other details

Image

### Acknowledgements

### 

---

dodieboy:
fix #9543

Checklist:


dodieboy:
btw do i need to update extVersionCode if i updated baseVersionCode?

stevenyomi:
SInce you updated baseVersionCode, each override automatically has incremented versionCode. You should not update overrideVersionCode individually.

lib-multisrc/goda/src/eu/kanade/tachiyomi/multisrc/goda/GoDa.kt
```diff
@@ -80,11 +80,13 @@ open class GoDa(
...
     private fun Element.getMangaId() = selectFirst("#mangachapters")!!.attr("data-mid")
 
+    open val elementCheck: Int = 6
+
     override fun mangaDetailsParse(response: Response) = SManga.create().apply {
         val document = response.asJsoup().selectFirst("main")!!
         val titleElement = document.selectFirst("h1")!!
         val elements = titleElement.parent()!!.parent()!!.children()
-        check(elements.size == 6)
+        check(elements.size == elementCheck)
```
stevenyomi:
```suggestion
        check(elements[4].tagName() == "p")
```

The element count varies across mirrors, as the linked issue's comments discovered. Let's change it to a more stable check. Remove the `elementCheck` field.

stevenyomi (changes requested):

dodieboy:
If goda api have author and all genre, we will not need that element checker
image

lib-multisrc/goda/src/eu/kanade/tachiyomi/multisrc/goda/GoDa.kt
dodieboy: btw, I move the check to below status since status do not use elements, prevent status from breaking when check is false

stevenyomi: `check()` will throw an exception. Should the page structure change, moving it below will not make the status data return successfully.

dodieboy: O, good to know, I still very new to android coding. Decided to do it because when it throw "illegal State Exception" error, the status is showing unkown when it should not affect it.

lib-multisrc/goda/src/eu/kanade/tachiyomi/multisrc/goda/GoDa.kt
```diff
@@ -81,17 +81,20 @@ open class GoDa(
...
-        val elements = titleElement.parent()!!.parent()!!.children()
-        check(elements.size == 6)
+        val elements = document.children()
 
         title = titleElement.ownText()
         status = when (titleElement.child(0).text()) {
             "連載中", "Ongoing" -> SManga.ONGOING
             "完結" -> SManga.COMPLETED
+            "停止更新" -> SManga.ON_HIATUS
+            "休刊" -> SManga.CANCELLED
```
stevenyomi: 休刊 is "on hiatus".

停止更新 can be mapped to "cancelled".

stevenyomi (changes requested):

lib-multisrc/goda/src/eu/kanade/tachiyomi/multisrc/goda/GoDa.kt
dodieboy: but from the manga that have see have those status, 停止更新 is for manga that is waiting for new seasion and 休刊 is manga that got cancel

dodieboy: here is a example: it stop updating/transalting due to personal reason and is mark as 休刊
https://m.baozimh.one/manga/qingchunqishilu-js919509

and this is on 停止更新 as the season ended
https://m.baozimh.one/manga/gelougongzhu-nr32104

stevenyomi: The former one is indeed "on hiatus/休刊" (artist taking a break for personal reasons).

The latter one is where the scanlator team discontinued translating new chapters. You can search the internet to see there was chapter 71 way back in 2023. Mapping it to "cancelled" is the best choice, although it originally refers to the artist/publisher cancelling the series.

dodieboy: my bad, it is so confusing as to me 休刊 mean suspended but base on the below manga 停止更新 mean cancel, so you are right, will make the change now
https://m.baozimh.one/manga/shenwuzhiling-suoyiwenhua

stevenyomi (approved):

PR #9694, opened 2025-07-17
Refactor ColaManga to use WebView for page loading

### LINKED ISSUE #9286: Colamanga: `Could not find key for keytype xxxxx` error when loading chapter.
### Source information

Colamanga 1.4.24

### Source language

Chinese

### Steps to reproduce

1. Open a comic from the source.
2. Click a chapter and then you'll received an error called `Could not find key for keytype xxxxx`.

### Expected behavior

The chapter should successfully loaded.

### Actual behavior

It doesn't load and an error has occurred.

### Mihon/Tachiyomi version

0.18.0-7281

### Android version

Android 15

### Other details

**Error logs**: [mihon_crash_logs.txt](https://github.com/user-attachments/files/20762563/mihon_crash_logs.txt)

### Acknowledgements

### 

---

Howard20181:
Replaces the previous deobfuscation and key mapping logic with a WebView-based approach for loading and decrypting manga pages. Adds a new webview-script.js asset for image extraction and communication with the app. Integrates data image interceptors and improves error handling for timeouts and script loading failures.

Closes #9286

Checklist:


Howard20181:
Note: debug for JavaScript part
https://gist.github.com/Howard20181/7cb220417da9451f135b9b999845d2ec

Howard20181:
It is now expected that system WebView versions 90 and below do not work. The fallback canvas drawing logic will be implemented.

Howard20181:
All work done. Works fine for me.

lib-multisrc/colamanga/build.gradle.kts
stevenyomi, line 1: Move the files into `src/zh/onemanhua` and make it back to a single source. The only other override, MangaDig, has been removed for a long time.

Refrain from merging any files to prevent git from not recognizing the file move.

lib-multisrc/colamanga/build.gradle.kts
```diff
@@ -2,8 +2,11 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 9
+baseVersionCode = 10
 
 dependencies {
     api(project(":lib:synchrony"))
```
stevenyomi: Remove unused dependency.

lib-multisrc/colamanga/src/eu/kanade/tachiyomi/multisrc/colamanga/ColaManga.kt
```diff
@@ -47,25 +63,31 @@ abstract class ColaManga(
 
     override val supportsLatest = true
 
-    private val json: Json by injectLazy()
-
-    private val intl = ColaMangaIntl(lang)
-
+    private var pages: MutableMap<String, ArrayList<Page>> = mutableMapOf()
```
stevenyomi: Use a more specific name like `pageCache` or `pagesMap`.

lib-multisrc/colamanga/src/eu/kanade/tachiyomi/multisrc/colamanga/ColaManga.kt
```diff
@@ -47,25 +63,31 @@ abstract class ColaManga(
...
-            TimeUnit.MILLISECONDS,
-        )
-        .addInterceptor(ColaMangaImageInterceptor())
-        .build()
-
-    override fun headersBuilder() = super.headersBuilder()
-        .add("Origin", baseUrl)
-        .add("Referer", "$baseUrl/")
+    var webView: MutableMap<String, WebView> = mutableMapOf()
+    val interfaceName = randomString()
```
stevenyomi: Use a more specific name like `webViewCache` or `webViewMap`.

I think these two fields can be private.

lib-multisrc/colamanga/src/eu/kanade/tachiyomi/multisrc/colamanga/ColaManga.kt
```diff
@@ -47,25 +63,31 @@ abstract class ColaManga(
...
+            .toLong(),
+        TimeUnit.MILLISECONDS,
+    ).setRandomUserAgent(
+        preferences.getPrefUAType(),
+        preferences.getPrefCustomUA(),
+        filterInclude = listOf("Chrome"),
+    ).addInterceptor(DataImageInterceptor()).build()
+
+    override fun headersBuilder() =
+        super.headersBuilder().add("Origin", baseUrl).add("Referer", "$baseUrl/")
```
stevenyomi: Refrain from introducing alternative formatting unless the original code broke any of the linting rules. This specific PR is a big code change and formatting changes are very noisy.

Please revert any unnecessary formatting changes, including the changes below.

lib-multisrc/colamanga/src/eu/kanade/tachiyomi/multisrc/colamanga/ColaManga.kt
```diff
@@ -187,78 +217,167 @@ abstract class ColaManga(
...
-            <script>
-                !function () {
-                    __cr.init();
-                    __cad.setCookieValue();
-
-                    const pageCountKey = __cad.getCookieValue()[1] + mh_info.pageid.toString();
-                    const pageCount = parseInt($.cookie(pageCountKey) || "0");
-                    const images = [...Array(pageCount).keys()].map((i) => __cr.getPicUrl(i + 1));
+    val handler = Handler(Looper.getMainLooper())
+    fun postDelayed(
```
stevenyomi: These can be private.

lib-multisrc/colamanga/src/eu/kanade/tachiyomi/multisrc/colamanga/ColaManga.kt
```diff
@@ -187,78 +217,167 @@ abstract class ColaManga(
...
+            handler.postDelayed(r, token, delayMillis)
+        } else {
+            return handler.postAtTime(r, token, SystemClock.uptimeMillis() + delayMillis)
+        }
+    }
 
-                    __cr.isfromMangaRead = 1;
+    override fun pageListParse(document: Document) = throw UnsupportedOperationException()
+    val baseUrlTopPrivateDomain = baseUrl.toHttpUrl().topPrivateDomain()
+    val emptyResourceResponse = WebResourceResponse(null, null, 204, "No Content", null, null)
```
stevenyomi: These can be private.

lib-multisrc/colamanga/src/eu/kanade/tachiyomi/multisrc/colamanga/ColaManga.kt
```diff
@@ -187,78 +217,167 @@ abstract class ColaManga(
...
+    @SuppressLint("SetJavaScriptEnabled")
+    private fun pageListParse(chapterUrl: String): List<Page> {
+        val url = baseUrl + chapterUrl
+        val latch = CountDownLatch(1)
+        val jsInterface = JsInterface(latch, chapterUrl, pages)
+        handler.post {
+            webView[chapterUrl]?.let {
+                it.destroy()
+                webView.remove(chapterUrl)
+            }
```
stevenyomi:
```suggestion
            webView.remove(chapterUrl)?.let { it.destroy() }
```

Same below.

lib-multisrc/colamanga/build.gradle.kts
```diff
@@ -2,8 +2,11 @@ plugins {
...
+    implementation(project(":lib:dataimage"))
+    implementation(project(":lib:randomua"))
+    implementation(project(":lib:i18n"))
```
AwkwardPeak7: remove i18n dependency as well when moving to `src/zh/onemanhua` as only Chinese strings are relevant, so they should be inlined

lib-multisrc/colamanga/src/eu/kanade/tachiyomi/multisrc/colamanga/ColaManga.kt
```diff
@@ -47,25 +63,31 @@ abstract class ColaManga(
...
+        baseUrl.toHttpUrl(),
+        preferences.getString(RATE_LIMIT_PREF_KEY, RATE_LIMIT_PREF_DEFAULT)!!.toInt(),
+        preferences.getString(RATE_LIMIT_PERIOD_PREF_KEY, RATE_LIMIT_PERIOD_PREF_DEFAULT)!!
+            .toLong(),
+        TimeUnit.MILLISECONDS,
+    ).setRandomUserAgent(
+        preferences.getPrefUAType(),
+        preferences.getPrefCustomUA(),
+        filterInclude = listOf("Chrome"),
+    ).addInterceptor(DataImageInterceptor()).build()
```
AwkwardPeak7: any reason for adding randomua?

randomua is known to cause issues, when app and webview use different useragents

Howard20181:
Will take a look tomorrow.

lib-multisrc/colamanga/src/eu/kanade/tachiyomi/multisrc/colamanga/ColaManga.kt
```diff
@@ -47,25 +63,31 @@ abstract class ColaManga(
...
-            preferences.getString(RATE_LIMIT_PERIOD_PREF_KEY, RATE_LIMIT_PERIOD_PREF_DEFAULT)!!.toLong(),
-            TimeUnit.MILLISECONDS,
-        )
-        .addInterceptor(ColaMangaImageInterceptor())
-        .build()
-
-    override fun headersBuilder() = super.headersBuilder()
-        .add("Origin", baseUrl)
-        .add("Referer", "$baseUrl/")
+    var webView: MutableMap<String, WebView> = mutableMapOf()
```
AwkwardPeak7: this will be taxing for lower end devices, when users start to open up chapter after chapter without reading completely, so there should be a upper limit to the number of cached webviews. you can just throw exceptions in `fetchImageUrl` if webview is no loner available.

use something like this to keep only 5 webviews at a time:
```kt
private val webViewCache = object : LinkedHashMap<String, WebView>() {
        override fun removeEldestEntry(eldest: MutableMap.MutableEntry<String, WebView>): Boolean {
            if (size <= 5) return false
            handler.post {
                eldest.value.destroy()
            }
            return true
        }
    }
```

Howard20181:
All request changes resolved.

src/zh/onemanhua/assets/webview-script.js
```diff
@@ -0,0 +1,94 @@
...
+            } else {
+                images[targetPageIndex].scrollIntoView();
+                currentPageIndex = targetPageIndex;
+                resolve();
+            }
+        });
+    });
+}
+function reloadPic(pageIndex) {
+    __cr.reloadPic(document.querySelectorAll("#mangalist"), pageIndex + 1)
```
Howard20181: Maybe `__cr.reloadPic(document.querySelector("#mangalist"), pageIndex + 1)`

Howard20181:
Reload image is buggy (T＿T)
I'm trying to fix it.

Howard20181:
~~It should work now.~~ But I'm curious about how long this pile of magic code can work.

Randomly broken when loading pages count greater than 20.

Howard20181:
It's a bad idea to read images using webview. If someone has a way to decrypt images, please close this PR
Tachidesk will no longer support.

Howard20181:
The current issue: Web pages load images sequentially, and images only begin loading when the user scrolls close to that page. Pages that have not yet begun loading do not exist on the page, so it is not possible to directly scroll to the target page. 
There is an extreme case on the reader (TachiX) side: If a chapter has 40 pages and the first 15 pages have already loaded, when the user opens it on the reader, the reader begins requesting page 16 or later. At this point, the reader must wait for the webview to load the first three pages, scroll to page 2, wait for pages 4 ~ 6 to load, scroll to page 5, and wait for pages 7 ~ 8 or fewer to load. This process must be repeated in a loop to reach the target page. Additionally, the scrolling speed must be sufficiently slow to avoid triggering the website's speed limit.

Howard20181:
I've tried my best to make it work. If it fails to load, just click Refresh and wait.

Howard20181:
Lazy loading has now been correctly triggered

Howard20181:
Current status: the first page randomly loads failed. Maybe start scrolling too early.

Howard20181:
All work done. It works on my side. But need more tests.

Howard20181:
Now the pages that have been read will be cleared, and CG will not be triggered continuously because of the large amount of memory usage.

Howard20181:
It's hard to sync the status between the webview and the app. Before all bugs are fixed, mark it draft.

Howard20181:
The loading failure caused by state synchronization issues has been fixed. However, there is no guarantee that the WebView will be destroyed after the chapter has been fully loaded.
In this case, the webview will not be destroyed: quickly flip back through several chapters and then flip forward through several chapters. Only the first few pages of each chapter are read. Then, read the last 5 pages of the previous chapter you stopped at. In this case, the reader only reads the first few pages and then switches to another page, while preloading also begins. It cannot be guaranteed that all chapters of the webview created each time will be filled into the pagesMap.
The number of WebViews is limited to a maximum of five, and they will be forcibly recycled after 30 minutes. This is not a major issue.

Fabali:
> The loading failure caused by state synchronization issues has been fixed. However, there is no guarantee that the WebView will be destroyed after the chapter has been fully loaded. In this case, the webview will not be destroyed: quickly flip back through several chapters and then flip forward through several chapters. Only the first few pages of each chapter are read. Then, read the last 5 pages of the previous chapter you stopped at. In this case, the reader only reads the first few pages and then switches to another page, while preloading also begins. It cannot be guaranteed that all chapters of the webview created each time will be filled into the pagesMap. The number of WebViews is limited to a maximum of five, and they will be forcibly recycled after 30 minutes. This is not a major issue.

Where should I download the plugin?

Norsze:
@Fabali Refrain from coming onto Github pull requests with support questions, this is the wrong forum for users with no programming experience. You wait for the pull request to be completed and merged, i.e. you do nothing and wait for an update button in your app. Commenting unnecessarily pings the people contributing to fixing these things in their spare time, they are not supposed to also deal with users who got lost on Github.

Your questions should be directed on our Discord from here on.

AwkwardPeak7 (approved):

MikeZeDev:
Now that Cola chapters are supposedly accessible only by using their app, would this PR still be accurate? Just asking.

xiyewuqiu:
Right now, you can only read cola through the app. I tried reverse engineering it before but couldn't figure it out - seems kind of difficult

PR #9767, opened 2025-07-23
Batoto: Fix bug baseUrl = "Auto"

dejavui:
Checklist:

**Image preview**
image

### DIFF
--- a/src/all/batoto/build.gradle
+++ b/src/all/batoto/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Bato.to'
     extClass = '.BatoToFactory'
-    extVersionCode = 59
+    extVersionCode = 51
     isNsfw = true
 }
 
--- a/src/all/batoto/build.gradle
+++ b/src/all/batoto/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Bato.to'
     extClass = '.BatoToFactory'
-    extVersionCode = 50
+    extVersionCode = 59
     isNsfw = true
 }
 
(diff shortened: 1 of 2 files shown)


src/all/batoto/build.gradle
```diff
@@ -1,7 +1,7 @@
 ext {
     extName = 'Bato.to'
     extClass = '.BatoToFactory'
-    extVersionCode = 50
+    extVersionCode = 59
```
vetleledaal:
```suggestion
    extVersionCode = 51
```

dejavui: sorry for my spelling mistake 😭

src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
```diff
@@ -97,7 +97,7 @@ open class BatoTo(
         get() {
             val current = field
             if (current.isNotEmpty()) {
-                return current
+                return if (current == "Auto") getMirrorPref() else current
```
stevenyomi: Put `.coerceAtLeast(0)` after the `val seed = runCatching {}.getOrElse {}` below in the `getMirrorPref()` to fix the issue. Remove this `if` expression.

vetleledaal: Won't that prefer one mirror over the others 50% of the time? Could [abs](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.math/abs.html) work?

stevenyomi: Will work, but I doubt if there are a bunch of different negative values returned.

stevenyomi: Also:

> `abs(Int.MIN_VALUE)` is `Int.MIN_VALUE` due to an overflow

stevenyomi (approved):

PR #9798, opened 2025-07-25
Manhastro: Fix auth cookie duplicated

choppeh:
This change fixes the request error when the expired cookie is sent with the new auth cookie. 

Checklist:

### DIFF
--- a/src/pt/manhastro/src/eu/kanade/tachiyomi/extension/pt/manhastro/Manhastro.kt
+++ b/src/pt/manhastro/src/eu/kanade/tachiyomi/extension/pt/manhastro/Manhastro.kt
@@ -58,10 +58,8 @@ class Manhastro :
     private var showWarning: Boolean = true
 
     private val authCookie: Cookie by lazy {
-        handler.post {
-            cookieManager.removeAllCookies { }
-            cookieManager.flush()
-        }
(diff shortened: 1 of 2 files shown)


src/pt/manhastro/src/eu/kanade/tachiyomi/extension/pt/manhastro/Manhastro.kt
```diff
@@ -52,9 +53,16 @@ class Manhastro :
 
     private val application: Application by lazy { Injekt.get<Application>() }
 
+    private val cookieManager by lazy { CookieManager.getInstance() }
+
     private var showWarning: Boolean = true
 
     private val authCookie: Cookie by lazy {
+        handler.post {
+            cookieManager.removeAllCookies { }
```
stevenyomi: I think you can do `cookieManager.removeAllCookies(null)` and not use the handler.

> This method is asynchronous. If a `ValueCallback` is provided, `onReceiveValue()` will be called on the current thread's `Looper` once the operation is complete. The value provided to the callback indicates whether any cookies were removed. You can pass `null` as the callback if you don't need to know when the operation completes or whether any cookies were removed, and in this case it is safe to call the method from a thread without a Looper.

stevenyomi (approved):

PR #9804, opened 2025-07-26
LeitorDeManga: Bypass jschallenge

### LINKED ISSUE #9764: Leitor de Mangá Error 403 and 500
### Source information

Leitor de Mangá 1.4.42

### Source language

Portuguese

### Steps to reproduce

1- I open the app
2- I go to the browser
3- I click on Leitor de manga
4- I click on any manga in the list
or
1- I open the app
2- I go to the browser
3- I click on migrate
4- Select any manga you want to migrate.
5- Click on the 3 dots and search manually.

### Expected behavior

be able to normally access the mangas and their chapters.

### Actual behavior

I always go into the manga from this source, it shows these 403 or 500 errors, and I can't update it or read it.

### Mihon/Tachiyomi version

TachiyomeSY 1.12.0

### Android version

android 14

### Other details

Sometimes I can't even access it through webview.

Image

Image

Image

### Acknowledgements

### 

---

choppeh:
Closes #9764

Checklist:

### DIFF
--- a/src/en/yakshascans/src/eu/kanade/tachiyomi/extension/en/yakshascans/YakshaScans.kt
+++ b/src/en/yakshascans/src/eu/kanade/tachiyomi/extension/en/yakshascans/YakshaScans.kt
@@ -44,13 +44,18 @@ class YakshaScans : Madara(
         return chain.proceed(chain.request())
     }
 
-    private fun fetchToken(chain: Interceptor.Chain, attempt: Int = 0): String {
+    private tailrec fun fetchToken(chain: Interceptor.Chain, attempt: Int = 0): String {
         if (attempt > MAX_ATTEMPT) {
             throw IOException("Failed to fetch challenge token!")
         }
-        return chain.proceed(GET("$baseUrl/hcdn-cgi/jschallenge", headers)).let {
-            TOKEN_REGEX.find(it.body.string())?.groups?.get(1)?.value
-                ?.takeUnless { it == "nil" } ?: fetchToken(chain, attempt.plus(1))
+
+        val response = chain.proceed(GET("$baseUrl/hcdn-cgi/jschallenge", headers))
+        val token = TOKEN_REGEX.find(response.body.string())?.groups?.get(1)?.value
+
+        return if (token != null && token != "nil") {
+            token
+        } else {
+            fetchToken(chain, attempt + 1)
         }
     }
 
--- a/src/en/yakshascans/src/eu/kanade/tachiyomi/extension/en/yakshascans/YakshaScans.kt
+++ b/src/en/yakshascans/src/eu/kanade/tachiyomi/extension/en/yakshascans/YakshaScans.kt
@@ -36,11 +36,13 @@ class YakshaScans : Madara(
             .add("challenge", token)
             .build()
 
-        chain.proceed(POST("$baseUrl/hcdn-cgi/jschallenge-validate", headers, body)).use {
-            if (it.isSuccessful.not()) {
-                throw IOException("Failed to bypass js challenge!")
+        chain.proceed(POST("$baseUrl/hcdn-cgi/jschallenge-validate", headers, body))
+            .apply(Response::close)
+            .run {
+                if (!isSuccessful) {
+                    throw IOException("Failed to bypass js challenge!")
+                }
             }
-        }
         return chain.proceed(chain.request())
     }
 
--- a/src/en/yakshascans/src/eu/kanade/tachiyomi/extension/en/yakshascans/YakshaScans.kt
+++ b/src/en/yakshascans/src/eu/kanade/tachiyomi/extension/en/yakshascans/YakshaScans.kt
@@ -22,33 +22,36 @@ class YakshaScans : Madara(
         .addInterceptor(::jsChallengeInterceptor)
         .build()
 
+    // Linked to src/pt/leitordemanga
     private fun jsChallengeInterceptor(chain: Interceptor.Chain): Response {
-        val request = chain.request()
-        val origRes = chain.proceed(request)
-        if (origRes.code != 403) return origRes
-        origRes.close()
+        val response = chain.proceed(chain.request())
+        if (response.code != 403) {
+            return response
+        }
+        response.close()
 
-        // Same delay as the source
         Thread.sleep(3000L)
         val token = fetchToken(chain).sha256()
+        val body = FormBody.Builder()
+            .add("challenge", token)
+            .build()
 
-        val body = FormBody.Builder().add("challenge", token).build()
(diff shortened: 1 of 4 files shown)


src/pt/leitordemanga/src/eu/kanade/tachiyomi/extension/pt/leitordemanga/LeitorDeManga.kt
```diff
@@ -17,5 +24,46 @@ class LeitorDeManga : Madara(
 
     override val client: OkHttpClient = super.client.newBuilder()
         .rateLimit(1, 2, TimeUnit.SECONDS)
+        .addInterceptor(::jsChallengeInterceptor)
         .build()
+
+    private fun jsChallengeInterceptor(chain: Interceptor.Chain): Response {
```
vetleledaal, line 31: Should probably update the JS challenge interceptor in YakshaScans, so there is less unique code to maintain. Possibly also add a comment linking them.

vetleledaal (dismissed):

src/pt/leitordemanga/src/eu/kanade/tachiyomi/extension/pt/leitordemanga/LeitorDeManga.kt
```diff
@@ -17,5 +24,50 @@ class LeitorDeManga : Madara(
...
+        val token = fetchToken(chain).sha256()
+        val body = FormBody.Builder()
+            .add("challenge", token)
+            .build()
+
+        chain.proceed(POST("$baseUrl/hcdn-cgi/jschallenge-validate", headers, body)).use {
+            if (it.isSuccessful.not()) {
+                throw IOException("Failed to bypass js challenge!")
+            }
+        }
```
stevenyomi: Just close it immediately instead of using `use()`. Also use `!` instead of `.not()`.

src/pt/leitordemanga/src/eu/kanade/tachiyomi/extension/pt/leitordemanga/LeitorDeManga.kt
```diff
@@ -17,5 +24,50 @@ class LeitorDeManga : Madara(
...
+
+    private fun fetchToken(chain: Interceptor.Chain, attempt: Int = 0): String {
+        if (attempt > MAX_ATTEMPT) {
+            throw IOException("Failed to fetch challenge token!")
+        }
+        return chain.proceed(GET("$baseUrl/hcdn-cgi/jschallenge", headers)).let {
+            TOKEN_REGEX.find(it.body.string())?.groups?.get(1)?.value
+                ?.takeUnless { it == "nil" } ?: fetchToken(chain, attempt.plus(1))
+        }
+    }
```
stevenyomi: Rewrite like this to make it `tailrec`.

```suggestion
    private tailrec fun fetchToken(chain: Interceptor.Chain, attempt: Int = 0): String {
        if (attempt > MAX_ATTEMPT) {
            throw IOException("Failed to fetch challenge token!")
        }
        val response = chain.proceed(GET("$baseUrl/hcdn-cgi/jschallenge", headers)).body.string()
        val token = TOKEN_REGEX.find(response)?.run { groups[1]!!.value }?.takeUnless { it == "nil" }
        return token ?: fetchToken(chain, attempt + 1)
    }
```

src/en/yakshascans/src/eu/kanade/tachiyomi/extension/en/yakshascans/YakshaScans.kt
stevenyomi, line 1: Bump this extension too. There are subtle behavior changes like the Regex.

src/pt/leitordemanga/src/eu/kanade/tachiyomi/extension/pt/leitordemanga/LeitorDeManga.kt
choppeh: > Just close it immediately instead of using `use()`. Also use `!` instead of `.not()`.

Why? I don't understand the preference for using a variable to get the `Response` and thus close the connection and this code.

And why is it preferable not to use the `.not()` function?

vetleledaal: I'm guessing `.not()` is primarily used in more functional parts of Kotlin, such as `.filter(Boolean::not)`.

You _can_ also do something similar in Python, but nobody uses it directly like this.
```py
>>> from operator import not_
>>> not_(True)
False
>>> not_(False)
True
```

 Using `.not()` in Kotlin might be from a convention I'm unaware of.

stevenyomi: > Why? I don't understand the preference for using a variable to get the `Response` and thus close the connection and this code.

I didn't say you need a variable to store it. You can do:

```kotlin
chain.proceed().apply { close() }.run {
    if (!isSuccessful) ...
}
// or even
chain.proceed().run {
    close()
    if (!isSuccessful) ...
}
```

> And why is it preferable not to use the `.not()` function?

There was a discussion on this topic among core contributors several years ago. Basically everyone who tried using `not()` switched back to `!` eventually. Although `not()` seems to be more readable and explicit, the counterarguments are:
- It is an operator overloading function. Using `not()` is like using `1.plus(1)` which is redundant.
- Negating the expression at the end is unnatural and doesn't read well (in your example "it is successful, not" vs "not it is successful"). Ideally it should look like `not condition`, similar in Python and C++, but sadly Kotlin doesn't have this syntax.
- It's not a widespread convention. There's no Kotlin style convention that encourages writing `not()`. This syntax introduces an unwanted significant style divergence.
- Every other language use it that kind of way (`!condition`) and most people read it fine. Following this convention makes it feel natural even for people familiar with other languages.

choppeh: > ```kotlin
> chain.proceed().apply { close() }.run {
>     if (!isSuccessful) ...
> }
> // or even
> chain.proceed().run {
>     close()
>     if (!isSuccessful) ...
> }
> ```

Ok, I'll change it, but with regard to the ```use{}``` function, what's wrong with using it to close the connection? Performance?

stevenyomi: Yeah, you're introducing a useless `try` block and some closing logic.

Technically you're not using the response body at all, so it doesn't seem to be a proper use of `use()`.

stevenyomi (approved):

PR #9810, opened 2025-07-26
Bato.to: expose all mirrors for CI

vetleledaal:
Not checked in GitHub Actions.

Checklist:

### DIFF
--- a/src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
+++ b/src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
@@ -53,7 +53,11 @@ open class BatoTo(
     private val preferences by getPreferencesLazy { migrateMirrorPref() }
 
     override val name: String = "Bato.to"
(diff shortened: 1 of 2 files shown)


src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
```diff
@@ -53,7 +53,11 @@ open class BatoTo(
     private val preferences by getPreferencesLazy { migrateMirrorPref() }
 
     override val name: String = "Bato.to"
-    override val baseUrl: String get() = mirror
+    override val baseUrl: String get() = when (System.getenv("CI")) {
+        "true" -> (MIRROR_PREF_ENTRY_VALUES.drop(1) + DEPRECATED_MIRRORS).joinToString("#, ")
```
stevenyomi: Put it inside `getMirrorPref()` as an early return.

I think you can also just `override var baseUrl` and absorb the `mirror` field.

stevenyomi (approved):

PR #9853, opened 2025-07-30
Create SlashLib

Mondenegro:
Add new ru site


Smol-Ame:
Don't open a blank PR. Plus, it supposedly already exists as "YaoiLib"

Next time learn to check first, and open a proper request instead, otherwise find somewhere else that will work on demand for free for you.

PR #9955, opened 2025-08-03
Fix weebcentral filter search showing duplicates

thatDudo:
Checklist:

When using filter search with weebcentral there was an issue of there being duplicate entries being shown. Every 32 entries there were 8 duplicates.

image

This appears to be a bug with their api, if you use limit=24 you still get 32 entries.
You can reproduce this by opening dev tools in https://weebcentral.com/search/data?limit=24&offset=32&sort=Latest+Updates&order=Descending&official=Any&anime=Any&adult=Any&included_tag=Action&display_mode=Full+Display and typing `document.querySelectorAll("body > article")`.

An easy fix is to set `FETCH_LIMIT` to 32.

### DIFF
--- a/src/en/weebcentral/src/eu/kanade/tachiyomi/extension/en/weebcentral/WeebCentral.kt
+++ b/src/en/weebcentral/src/eu/kanade/tachiyomi/extension/en/weebcentral/WeebCentral.kt
@@ -246,7 +246,7 @@ class WeebCentral : ParsedHttpSource() {
     )
 
     companion object {
-        const val FETCH_LIMIT = 24
+        const val FETCH_LIMIT = 32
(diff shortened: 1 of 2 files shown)


src/en/weebcentral/src/eu/kanade/tachiyomi/extension/en/weebcentral/WeebCentral.kt
```diff
@@ -246,7 +246,7 @@ class WeebCentral : ParsedHttpSource() {
     )
 
     companion object {
-        const val FETCH_LIMIT = 24
+        const val FETCH_LIMIT = 32
```
stevenyomi, line 251: Add a comment above this line to explain the reason.

stevenyomi (approved):

PR #10033, opened 2025-08-07
Comick: increase rate limit and separate

### LINKED ISSUE #9057: Comick: Increase rate-limit to avoid/reduce HTTP 429 error
### Source name

Comick

### Source language

English

Recently, Comick experienced a wave of DDoS attacks, and on 5/31, they performed maintenance to update their Cloudflare anti-DDoS protection.

Since that update, many people started encountering frequent 429 “Too Many Requests” errors when updating their followed series. The issue appears to follow a pattern: a batch of requests will go through successfully, but then the extension begins triggering 429s for several series at a time, only a few updates succeed per second before more errors occur.

According to Comick’s lead developer on Discord:

> “We have zero control over the anti-DoS system, that's entirely managed by Cloudflare.”

Another developer has also commented that the extension sends too many requests, and it would need to be fixed on the extension's end. (Referring to Mihon at large)

> if you are using Tachi/Mihon then you have to ask their developers
> Yeah, because their app is not optimized
> The app sent more than 120 req/1 minute

Given it is unlikely to change on their side, I'd like to request an increase to the rate limit to help reduce the number of 429 errors.

### Other details

_No response_

### Acknowledgements

### 

---

Riztard:
Closes #9057

Checklist:

### DIFF
--- a/src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
+++ b/src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
@@ -1,6 +1,7 @@
 package eu.kanade.tachiyomi.extension.all.comickfun
 
 import android.content.SharedPreferences
+import android.util.Log
 import androidx.preference.EditTextPreference
 import androidx.preference.ListPreference
 import androidx.preference.PreferenceScreen
@@ -9,7 +10,6 @@ import eu.kanade.tachiyomi.lib.i18n.Intl
 import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.network.asObservableSuccess
 import eu.kanade.tachiyomi.network.interceptor.rateLimit
-import eu.kanade.tachiyomi.network.interceptor.rateLimitHost
 import eu.kanade.tachiyomi.source.ConfigurableSource
 import eu.kanade.tachiyomi.source.model.FilterList
 import eu.kanade.tachiyomi.source.model.MangasPage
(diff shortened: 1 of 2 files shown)


src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/Comick.kt
```diff
@@ -248,7 +249,8 @@ abstract class Comick(
 
     override val client = network.cloudflareClient.newBuilder()
         .addNetworkInterceptor(::errorInterceptor)
-        .rateLimit(3, 1, TimeUnit.SECONDS)
+        .rateLimit(12, 8, TimeUnit.SECONDS) // == 1.5req/1sec == 3req/2sec == 90req/60sec
```
AwkwardPeak7, line 256: 1) You are adding double rate limits on the api, one general and one specific. 
2) imo time should be 1 for any ratelimit. 12/8 would fire 12 requests immediately then wait 8 seconds instead of spreading it out.

Riztard: > You are adding double rate limits on the api

ah true

2. yea thats my intention. my intention was more targetted for the thumbnail when browsing so it wont be too slow but still under reasonable limit

Riztard: but i also understand it would need to wait for 8 sec. and i think that is short enough for "normal" use

AwkwardPeak7: But if you scroll down quickly, no thumbnail will load as it will start to wait 8+8+8...seconds 

You could remove the ratelimit on the thumbnails. Append some fragment to the thumbnail URLs and check for the fragment in an interceptor, if found redirect the request through another non ratelimit client. 
See Dynasty extension, I did this there

stevenyomi (approved):

stevenyomi (approved):

PR #10020, opened 2025-08-06
Enable R8

stevenyomi:
Related to #8928

Only tested with MangaDex and another simple extension yet

TODO: run extension inspector on PR build (and need release variant, currently debug)

PR #10039, opened 2025-08-07
Update utility functions in `core` module

stevenyomi:
Split from #10020.

- [ ] Investigate R8's inlining heuristics on `SDF.tryParse()`
- [ ] Probably add `Call.executeSuccess()`
- [ ] Add KDoc on usage
- [ ] Add to OWNERS

### DIFF
--- a/core/src/main/kotlin/keiyoushi/utils/Date.kt
+++ b/core/src/main/kotlin/keiyoushi/utils/Date.kt
@@ -1,14 +1,10 @@
 package keiyoushi.utils
 
-import java.text.ParseException
+import java.text.ParsePosition
 import java.text.SimpleDateFormat
 
 fun SimpleDateFormat.tryParse(date: String?): Long {
     date ?: return 0L
 
-    return try {
-        parse(date)!!.time
-    } catch (_: ParseException) {
-        0L
-    }
+    return parse(date, ParsePosition(0))?.time ?: 0L
 }
--- a/core/src/main/kotlin/keiyoushi/utils/Date.kt
+++ b/core/src/main/kotlin/keiyoushi/utils/Date.kt
@@ -3,12 +3,11 @@ package keiyoushi.utils
 import java.text.ParseException
 import java.text.SimpleDateFormat
 
-@Suppress("NOTHING_TO_INLINE")
-inline fun SimpleDateFormat.tryParse(date: String?): Long {
+fun SimpleDateFormat.tryParse(date: String?): Long {
     date ?: return 0L
 
     return try {
-        parse(date)?.time ?: 0L
+        parse(date)!!.time
     } catch (_: ParseException) {
         0L
     }
--- a/core/src/main/kotlin/keiyoushi/utils/Json.kt
+++ b/core/src/main/kotlin/keiyoushi/utils/Json.kt
@@ -1,13 +1,16 @@
 package keiyoushi.utils
 
+import kotlinx.serialization.DeserializationStrategy
 import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.decodeFromStream
+import kotlinx.serialization.serializer
 import okhttp3.Response
-import uy.kohesive.injekt.injectLazy
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
 
-val jsonInstance: Json by injectLazy()
+val jsonInstance: Json = Injekt.get()
 
 /**
  * Parses JSON string into an object of type [T].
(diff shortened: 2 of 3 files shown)


core/src/main/kotlin/keiyoushi/utils/Json.kt
```diff
@@ -1,13 +1,16 @@
...
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.decodeFromStream
+import kotlinx.serialization.serializer
 import okhttp3.Response
-import uy.kohesive.injekt.injectLazy
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
 
-val jsonInstance: Json by injectLazy()
+val jsonInstance: Json = Injekt.get()
```
stevenyomi, line 13: 1. This is actually a static field in the `JsonKt` class, so it's only initialized when any of the non-inlined members is accessed (which is itself).
2. `Injekt.get()` is actually [faster](https://discord.com/channels/1193460528052453448/1195321719427174460/1392155833789317233) than initializing a `lazy` placeholder. I think `injectLazy()` is primarily used for registered class factories that take a lot of time to construct an instance, so it shouldn't be misused for already constructed instances like `Json`. I'm planning to add a linting rule later to remove this kind of misuse.

core/src/main/kotlin/keiyoushi/utils/Json.kt
```diff
@@ -19,10 +22,20 @@ inline fun <reified T> String.parseAs(json: Json = jsonInstance): T =
  * Parses the response body into an object of type [T].
  */
 inline fun <reified T> Response.parseAs(json: Json = jsonInstance): T =
-    use { json.decodeFromStream(body.byteStream()) }
+    json.decodeFromResponse(serializer(), this)
```
stevenyomi, line 25: 1. Extracting a non-inlined function to reduce code size with no performance impact.
2. This is the same as https://github.com/mihonapp/mihon/commit/0d8f1c8560c5d65712d1656856ab33b6de0adb09 (PR at https://github.com/tachiyomiorg/tachiyomi/pull/9047 or [archive](https://git.mihon.tech/tachiyomi/tachiyomi/pulls/9047)). Basically this `serializer()` call will be evaluated at compile-time if we update to [serialization 1.5.0 and Kotlin plugin 1.8.0-ish](https://redirect.github.com/Kotlin/kotlinx.serialization/issues/1348#issuecomment-1472065974).

core/src/main/kotlin/keiyoushi/utils/Json.kt
```diff
@@ -19,10 +22,20 @@ inline fun <reified T> String.parseAs(json: Json = jsonInstance): T =
...
     json.encodeToString(this)
+
+@PublishedApi
+internal fun <T> Json.decodeFromResponse(
+    deserializer: DeserializationStrategy<T>,
+    response: Response,
+): T {
+    return response.body.byteStream().use {
+        decodeFromStream(deserializer, it)
+    }
```
stevenyomi, line 40: And we can use `decodeFromBufferedSource()` like https://github.com/mihonapp/mihon/commit/2d19729869d1c88a46c0d1a4582d29a80d7ab9cb if the dependency on `kotlinx-serialization-json-okio` is ensured.

core/src/main/kotlin/keiyoushi/utils/Preferences.kt
```diff
@@ -25,4 +25,8 @@ inline fun HttpSource.getPreferencesLazy(
  */
 @Suppress("NOTHING_TO_INLINE")
 inline fun getPreferences(sourceId: Long): SharedPreferences =
-    Injekt.get<Application>().getSharedPreferences("source_$sourceId", 0x0000)
+    getApplication().getSharedPreferences("source_$sourceId", 0x0000)
+
+// This ensures the type token isn't duplicated at each call site.
+@PublishedApi
+internal fun getApplication(): Application = Injekt.get()
```
stevenyomi: This was discovered in #10020. Even though this line was in an `inline` function, it still drops a type token here in this file, which is hard to optimize out due to how R8 rules are written. So I guess it's best to not duplicate it in the extension code.

core/src/main/kotlin/keiyoushi/utils/Date.kt
```diff
@@ -3,12 +3,11 @@ package keiyoushi.utils
 import java.text.ParseException
 import java.text.SimpleDateFormat
 
-@Suppress("NOTHING_TO_INLINE")
-inline fun SimpleDateFormat.tryParse(date: String?): Long {
+fun SimpleDateFormat.tryParse(date: String?): Long {
```
stevenyomi, line 6: Inlining normal functions doesn't provide performance boost. R8 will inline it if there's only 1 call site.

core/src/main/kotlin/keiyoushi/utils/Date.kt
```diff
@@ -3,12 +3,11 @@ package keiyoushi.utils
...
     date ?: return 0L
 
     return try {
-        parse(date)?.time ?: 0L
+        parse(date)!!.time
```
stevenyomi: Originally mentioned at https://github.com/keiyoushi/extensions-source/pull/1808#discussion_r1521311791:

> It won't be `null` at all, see [`Date parse(String) throws ParseException`](https://docs.oracle.com/javase/8/docs/api/java/text/DateFormat.html#parse-java.lang.String-). IDE wrongly prompts it as nullable because it wraps [`Date? parse(String, ParsePosition)`](https://docs.oracle.com/javase/8/docs/api/java/text/DateFormat.html#parse-java.lang.String-java.text.ParsePosition-) and throws on a check of `ParsePosition` so the IDE can't tell.

stevenyomi: So I just switched to `parse(String, ParsePosition)` to let the code look safer -- saves a `try` block too.

core/src/main/kotlin/keiyoushi/utils/Json.kt
cuong-tran: will dropping the `use` make the `response` not properly closed and cause memory leaks?

AwkwardPeak7:
This pull request has been closed due to a force push to main. Please rebase and open a new PR. Thank you for your understanding.

PR #10059, opened 2025-08-08
AnimeSama: add filtering by genre

Yakoo63:
Using almost the same code as FuzzyDoodle, I added the only filter available on animesama: genre

Checklist:


src/fr/animesama/src/eu/kanade/tachiyomi/extension/fr/animesama/AnimeSama.kt
```diff
@@ -33,13 +39,56 @@ class AnimeSama : ParsedHttpSource() {
...
+        }
+        if (filters.size < 1) {
+            filters.add(0, Filter.Header("Press 'reset' to load more filters"))
+        }
+
+        return FilterList(filters)
+    }
+
+    private val scope = CoroutineScope(Dispatchers.IO)
+    private fun launchIO(block: suspend () -> Unit) = scope.launch { block() }
```
stevenyomi:
```suggestion
    private fun launchIO(block: suspend () -> Unit) = GlobalScope.launch { block() }
```

stevenyomi (approved):

PR #10083, opened 2025-08-10
Fix "restart tachiyomi" message when making changes to komga [ettings

chelming:
Removing "tachiyomi" from the "apply settings" message


stevenyomi:
There are many other references to Tachiyomi (like in the extension APK's name) and that's not a serious problem. This can be resolved later. Open an issue if you want to track the problem.

PR #9552, opened 2025-07-05
Add bilimanga source

hualiong:
Closes #9497

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [ ] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [x] Have removed `web_hi_res_512.png` when adding a new extension

PR #10146, opened 2025-08-15
Bilimanga & XGMN: fix manga url error when opening webview and the issue with page turning

hualiong:
Checklist:


src/zh/bilimanga/src/eu/kanade/tachiyomi/extension/zh/bilimanga/BiliManga.kt
```diff
@@ -27,7 +27,7 @@ class BiliManga : HttpSource(), ConfigurableSource {
 
     override val lang = "zh"
 
-    override val name = "Bilimanga.net"
+    override val name = "嗶哩漫畫"
```
stevenyomi: Sources should not be renamed without good reason. I mentioned this name being misleading in #9552.

hualiong: Oh，I forget it

stevenyomi (approved):

Issue #10110, opened 2025-08-12
Changes in actions workflow don't trigger rebuild


stevenyomi:
Workflow changes in `build_{pull_request,push}.yml` like #10109 should trigger a full rebuild.

Reference: https://github.com/keiyoushi/extensions-source/commit/d1d3b8a51a3adf13d76a49243aefcfad4e8125b9

We should just remove the path filter in PR build check in favor of the Python  script due to https://github.com/orgs/community/discussions/44490 (official workaround: https://stackoverflow.com/a/70284299).

Should also move `always_build.json` into `scripts` (and rename to `ci`) to simplify path checking.

PR #10190, opened 2025-08-19
Update CI workflows and add security scanning

stevenyomi:
- Add zizmor scanning and fixed reported issues
- Does NOT target #10110; that will be done when refactoring CI code
- The zizmor workflow has path filters and won't be enforced in PRs in the future like the PR build check; the scanning feature will add comments on violation, should be easy to spot
- Only used `permissions` at top level because our workflows are simple; didn't go the pedantic way (use `{}` at top level and add permissions in specific jobs)
- Uses GitHub Actions token for issue moderator; membership check through a special member token

### DIFF
--- a/.github/workflows/build_pull_request.yml
+++ b/.github/workflows/build_pull_request.yml
@@ -8,6 +8,8 @@ on:
       - '!.github/**'
       - '.github/workflows/build_pull_request.yml'
 
+permissions: {}
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
   cancel-in-progress: true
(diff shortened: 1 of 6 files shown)


.github/workflows/build_pull_request.yml
```diff
@@ -68,5 +73,7 @@ jobs:
           cache-read-only: true
 
       - name: Build extensions (${{ matrix.chunk.number }})
+        env:
+          MODULES: ${{ join(matrix.chunk.modules, ' ') }}
         run: |
-          ./gradlew $(echo '${{ toJson(matrix.chunk.modules) }}' | jq -r 'join(" ")')
+          ./gradlew $MODULES
```
vetleledaal, line 81: Shellcheck would usually complain about this, but I guess it's intentional here - spaces separate arguments.

vetleledaal (dismissed):
Consider using [`timeout-minutes`](https://github.com/suzuki-shunsuke/ghalint/blob/main/docs/policies/012.md#why).

PR #10067, opened 2025-08-09
Add Xgmn Source

hualiong:
Checklist:

- [ ] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [ ] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [ ] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [ ] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [x] Have removed `web_hi_res_512.png` when adding a new extension

PR #10235, opened 2025-08-22
fix(YellowNote): adapt to web page structure changes

marioplus:
Checklist:


src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNote.kt
```diff
@@ -50,86 +50,131 @@ class YellowNote(
...
+            .firstNotNullOfOrNull { dateRegex.find(it)?.value }
+            ?.let { dateFormat.tryParse(it) }
+    }
+
     override fun pageListParse(document: Document): List<Page> {
-        return document.select("div.article.mask .photos img.cr_only")
-            .mapIndexed { i, imgEl -> Page(i, imageUrl = imgEl!!.absUrl("src")) }
+        return document.select(simpleMangaSelector())
+            .map { imageElement -> parseUrlFormStyle(imageElement.selectFirst("div.img"))!! }
+            .mapIndexed { i, url -> Page(i, imageUrl = url) }
```
stevenyomi: Fuse the two operations into one `mapIndexed()`.

src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNote.kt
```diff
@@ -50,86 +50,131 @@ class YellowNote(
...
+                date_upload = uploadAt
             }
         }
     }
 
+    private fun parseUploadDateFromVersionInfo(doc: Document): Long? {
+        val navsElement = doc.selectFirst("div.tab-navs")
+        val versionInfoElementIndex = navsElement
+            ?.select("div.tab-nav")
+            ?.indexOfFirst { it.attr("data-tab") == "tab_68a8963ddf8a5_5" }
```
stevenyomi: Doesn't look like a stable value.

src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNote.kt
```diff
@@ -50,86 +50,131 @@ class YellowNote(
...
-    private val styleUrlRegex = """url\(['"]?([^'"]+)['"]?\)""".toRegex()
+    // yyyy.MM.dd
+    private val dateRegex = """\d{4}\.\d{2}\.\d{2}""".toRegex()
+
+    private val styleUrlRegex = """background-image\s*:\s*url\('([^']+)'\)""".toRegex()
+
+    // 100P + 2V
+    private val mediaCountRegex = """\d+P( \+ \d+V)?""".toRegex()
+
+    private val selectParts = listOf("photo", "amateur")
```
stevenyomi: Inline this.

src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNote.kt
marioplus: Speculation based on the structure of 'data-tab'

src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNote.kt
```diff
@@ -50,86 +50,134 @@ class YellowNote(
...
+    private val imageSelector = "div.list.photo-items > div.item.photo-image, div.list.amateur-items > div.item.amateur-image"
     override fun pageListParse(document: Document): List<Page> {
-        return document.select("div.article.mask .photos img.cr_only")
-            .mapIndexed { i, imgEl -> Page(i, imageUrl = imgEl!!.absUrl("src")) }
+        return document.select(imageSelector)
+            .mapIndexed { i, imageElement ->
+                val url = parseUrlFormStyle(imageElement.selectFirst("div.img"))!!
+                Page(i, imageUrl = url)
+            }
+            .toList()
```
stevenyomi:
```suggestion
```

src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNote.kt
```diff
@@ -50,86 +50,134 @@ class YellowNote(
...
+            .asSequence()
+            // info
+            .filter { it.id().endsWith("_5") }
+            // info items
+            .map { it.select("div.info-card > div.text") }
+            .flatten()
+            .filterNotNull()
+            .map { it.text() }
+            .firstNotNullOfOrNull { dateRegex.find(it)?.value }
+            ?.let { dateFormat.tryParse(it) }
```
stevenyomi: This is too complicated. Please rewrite it into a simple for-loop.

stevenyomi:
```kotlin
for (info in doc.select("div.info-card div.text")) {
    val date = dateRegex.find(info.text()) ?: continue
    return dateFormat.parse(date.value)!!.time
}
return 0L
```

You sure your selectors are correct? I have `div.tab-content` instead of `div.tab-contents` and the DOM says `div.info-card > div.item > div.text` instead of your two level thing.

src/all/yellownote/build.gradle
```diff
@@ -1,7 +1,7 @@
 ext {
     extName = 'YellowNote'
     extClass = '.YellowNoteSourceFactory'
```
stevenyomi: This source should not be a multiple-language factory. It should be one source with `"all"` language tag (shown in the app as "Multi"), and provide language selection as an option. The ideal approach is to require user to select a language in preferences before any request. In the setting description remind user that if they changed language, favorited items might not refresh their titles unless an experimental feature is toggled in the app.

stevenyomi: Multiple-language source factory is used for sources that have manga **translations** in multiple languages, not UI language. See also https://github.com/keiyoushi/extensions-source/pull/10067#discussion_r2264707318.

marioplus: I have two questions:
1. How to determine what language a user is using
2. How to prompt users to select a language

Is there any documentation or example code on this?

src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNote.kt
marioplus: image

```kotlin
doc.select("""div.tab-contents > div.tab-content[id^="tab_"][id$="_5"] div.text""")
  .asSequence()
  .map { it.text() }
  .map { dateRegex.find(it)?.value }
  .filterNotNull()
  .map { dateFormat.tryParse(it) }
  .find { it != 0L }
```

src/all/yellownote/build.gradle
marioplus: When prompting users to select a language, the prompt itself should be displayed in a language they understand. Wouldn’t that eliminate the need for a language selection?

stevenyomi: You should implement `ConfigurableSource`, then in the method, add:

```kotlin
ListPreference(screen.context).run {
    key = LANG_PREF
    title = "Language" // or intl["..."]
    summary = "If you changed language, ..."
    entries = arrayOf("None", "简体中文", "English")
    entryValues = arrayOf("", "zh", "en")
    setDefaultValue("")
    screen.addPreference(this)
}
```

In the class, add:

```kotlin
private val preferences = getPreferences()

private val langPrefix: String get() = preferences.getString(LANG_PREF, "")!!.ifEmpty { 
    throw Exception("Select language in extension settings") // or intl["..."]
    // or automatically match their system language
    val result = "en"
    preferences.edit().putString(LANG_PREF, result).apply()
    result
}

private const val LANG_PREF = "LANG"
```

To check user's system language, refer to this:

https://github.com/keiyoushi/extensions-source/blob/1933abd91d1fb6d1be4ae52839240b97040f782d/src/ko/newtoki/src/eu/kanade/tachiyomi/extension/ko/newtoki/Strings.kt#L8-L12

stevenyomi: You'll need to adjust the language in `Intl` accordingly. Either match system or use `en` until user selects their language and restarts their app.

src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNote.kt
stevenyomi: Please use my suggested `for-loop` version.

Also, the `_5` doesn't look reliable since the website might change the tabs' order.

src/all/yellownote/build.gradle
marioplus: I checked #10067 and
https://github.com/keiyoushi/extensions-source/blob/1933abd91d1fb6d1be4ae52839240b97040f782d/src/ko/newtoki/src/eu/kanade/tachiyomi/extension/ko/newtoki/NewToki.kt#L38

But I still don’t know how to implement internationalization for the settings interface. ManaToki seems to hardcode the language as "ko".

I tried:

```kontlin
override val lang by lazy { preferences.language() }
```
But this causes the extension to disappear from the list.

stevenyomi: Do 

```kotlin
override val lang get() = "all"
```

src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNote.kt
```diff
@@ -44,92 +44,133 @@ class YellowNote(
...
-            val imgEl = element.selectFirst("img")!!
-            val titleAppend = element.selectFirst("div.tag > div")?.text()?.let { "($it)" }.orEmpty()
-            title = "${imgEl.attr("alt")}$titleAppend"
+        val formatMediaCount = element.select("div.tags > div")
+            .map { it.text() }
+            .firstOrNull { mediaCountRegex.matches(it) }
+            ?.let { "($it)" }
+            .orEmpty()
+        title = "${mangaEl.attr("title")}$formatMediaCount"
+        Log.d("mario12", "title: ${mangaEl.attr("title")}, mediaCount: $formatMediaCount")
```
stevenyomi: Remove this log.

src/all/yellownote/assets/i18n/messages_en.properties
```diff
@@ -78,3 +78,8 @@ config.domain.dialog.message=Default domain:
 config.domain.toast.changed-success=Domain changed successfully. It will take effect after restarting the app.
 config.domain.toast.changed-failed=Invalid domain!
 
+config.language.title=Language
+config.language.summary=Set extension language. Effective after restart.
+
+error.unset-language=Please select a language in extension settings first.
```
stevenyomi: This is not used.

src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/LanguageUtils.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    val supportedLocaleTags = languageToSubdomainMap.keys.toTypedArray()
+
+    fun getDefaultLanguage(): String {
+        val defaultLocale =
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                LocaleList.getDefault().getFirstMatch(supportedLocaleTags) ?: baseLocale
+            } else {
+                Locale.getDefault() ?: baseLocale
+            }
```
stevenyomi, line 35: No need for `?:`.

src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/LanguageUtils.kt
```diff
@@ -0,0 +1,68 @@
...
+                }
+            }
+            // Secondary: Determine by country code
+            else -> {
+                when (defaultLocale.country) {
+                    "TW", "HK", "MO" -> "zh-Hant"
+                    else -> "zh-Hans"
+                }
+            }
+        }
```
stevenyomi, line 43: This function only returns either `zh-Hans` or `zh-Hant`.

For the `Build.VERSION.SDK_INT >= Build.VERSION_CODES.N` branch you should return the result directly.

For the second branch you can do some matching and return `en` as the default value.

src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNote.kt
```diff
@@ -25,12 +26,10 @@ import org.jsoup.nodes.Element
 import java.text.SimpleDateFormat
 import java.util.Locale
 
-class YellowNote(
-    override val lang: String,
-    private val subdomain: String? = null,
-) : SimpleParsedHttpSource(), ConfigurableSource {
+class YellowNote : SimpleParsedHttpSource(), ConfigurableSource {
```
stevenyomi, line 29: Add an `override val id get() = ...` and pick a source ID from:

https://github.com/keiyoushi/extensions/blob/c8675034e52a35db700edc58eebd317909636087/index.json#L8117-L8151

The value type is `Long`.

src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNotePreferences.kt
```diff
@@ -13,17 +15,22 @@ object YellowNotePreferences {
...
-        val value = getString(key, defValue)!!
+    private fun SharedPreferences.getStringOrDefault(key: String, defValue: String): String {
+        val value = getString(key, defValue) ?: defValue
         return value.ifBlank { defValue }
     }
 
+    private fun SharedPreferences.getStringOrDefault(key: String, defValueSupplier: () -> String): String {
+        val value = getString(key, "") ?: defValueSupplier.invoke()
+        return value.ifBlank { defValueSupplier.invoke() }
+    }
```
stevenyomi: Remove these helpers. `getString(key, NON_NULL_VALUE)` is guaranteed to not return a non-null value and you can chain `!!.ifBlank { }` if needed.

src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNote.kt
```diff
@@ -44,92 +45,138 @@ class YellowNote(
...
-                date_upload = dateUpload
+                date_upload = uploadAt
             }
         }
     }
 
+    private fun parseUploadDateFromVersionInfo(doc: Document): Long? {
+        for (info in doc.select("div.tab-content > div.info-card div.text")) {
+            val date = dateRegex.find(info.text()) ?: continue
+            return dateFormat.parse(date.value)!!.time
```
AwkwardPeak7: use `tryParse` here too

AwkwardPeak7 (approved):

PR #10252, opened 2025-08-23
zh/boylove: add manga region filter

morallkat:
This PR adds a dedicated filter by manga source region for BoyLove, since the site split the region tags into a dedicated bar earlier this week. This meant that users could no longer query manga by region from the genre list as before.

Since the API supports multi-tag queries (by adding tags together using "+"), this PR has made the necessary changes to enable users to query two tags (region and genre) simultaneously.

Checklist:

### DIFF
--- a/src/zh/boylove/src/eu/kanade/tachiyomi/extension/zh/boylove/BoyLoveFilters.kt
+++ b/src/zh/boylove/src/eu/kanade/tachiyomi/extension/zh/boylove/BoyLoveFilters.kt
@@ -6,7 +6,7 @@ import eu.kanade.tachiyomi.source.model.FilterList
 /*
  * 1-0-2-1-1-0-1-0
  * [0] cate, useless
- * [1] tag(genre), 0=all
+ * [1] tag(genre and region), 0=all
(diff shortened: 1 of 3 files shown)


src/zh/boylove/src/eu/kanade/tachiyomi/extension/zh/boylove/BoyLoveFilters.kt
```diff
@@ -17,20 +17,30 @@ import eu.kanade.tachiyomi.source.model.FilterList
...
+            is RegionFilter -> region = REGION_KEYS[filter.state]
             is GenreFilter -> if (filter.state > 0) genre = filter.values[filter.state]
             is SortFilter -> sort = SORT_KEYS[filter.state]
             is VipFilter -> vip = VIP_KEYS[filter.state]
             else -> {}
         }
     }
-    return "1-$genre-$status-$sort-$page-$type-1-$vip"
+
+    var tags = when {
```
stevenyomi:
```suggestion
    val tags = when {
```

Not sure why that's a typo.

stevenyomi (approved):

PR #10277, opened 2025-08-25
MiMiHentai: Update search support exclude tag and remove 'yaoi' tag from Latest/Popular page

dejavui:
Many mimihentai users suggested removing the 'yaoi' tag from the Latest/Popular page 
Checklist:


src/vi/mimihentai/src/eu/kanade/tachiyomi/extension/vi/mimihentai/MiMiHentai.kt
```diff
@@ -44,6 +44,7 @@ class MiMiHentai : HttpSource() {
             addPathSegments("tatcatruyen")
             addQueryParameter("page", (page - 1).toString())
             addQueryParameter("sort", "updated_at")
+            addQueryParameter("ex", "196")
```
cuong-tran, line 47: Don't think you could make assumption for all users. If you want a permanent exclusion, better put it into a preference setting.

stevenyomi (approved):

PR #10353, opened 2025-09-01
fix(Madara): preserve the capitalization of tags and genres

zweimach:
Fixed the capitalization of tags and genres by manually lowercasing strings and comparing with `any` + predicate.  

Checklist:

### DIFF
--- a/lib-multisrc/madara/src/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
+++ b/lib-multisrc/madara/src/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
@@ -702,12 +702,11 @@ abstract class Madara(
                 }
             }
             val genres = select(mangaDetailsSelectorGenre)
-                .map { element -> element.text().lowercase(Locale.ROOT) }
-                .toMutableSet()
+                .mapTo(ArrayList()) { element -> element.text() }
 
             if (mangaDetailsSelectorTag.isNotEmpty()) {
                 select(mangaDetailsSelectorTag).forEach { element ->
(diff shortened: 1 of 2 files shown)


lib-multisrc/madara/src/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
```diff
@@ -702,42 +702,32 @@ abstract class Madara(
...
-                    if (it.isLowerCase()) {
-                        it.titlecase(
-                            Locale.ROOT,
-                        )
-                    } else {
-                        it.toString()
-                    }
-                }
-            }
+            manga.genre = genres.toList().joinToString()
```
stevenyomi:
```suggestion
            manga.genre = genres.distinctBy(String::lowercase).joinToString()
```

You'll also need to remove the `toMutableSet()` call and the `any` comparisons.

lib-multisrc/madara/src/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
```diff
@@ -702,42 +702,32 @@ abstract class Madara(
                 }
             }
             val genres = select(mangaDetailsSelectorGenre)
-                .map { element -> element.text().lowercase(Locale.ROOT) }
+                .map { element -> element.text() }
```
stevenyomi:
```suggestion
                .mapTo(ArrayList()) { element -> element.text() }
```

stevenyomi (approved):

PR #10389, opened 2025-09-04
Fix PoseidonScans missing premium chapters and volume detection

Nyantad:
Fixed missing chapters with expired premium status and added automatic volume detection to display "Volume X" instead of "Chapitre X" for volume entries.

Checklist:

### DIFF
--- a/src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
+++ b/src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
@@ -388,33 +388,22 @@ class PoseidonScans : HttpSource() {
         }
 
         return mangaDto.chapters
-            ?.filter { ch ->
-                // If chapter is not premium, include it
-                if (ch.isPremium != true) return@filter true
-
-                // If chapter is premium, check if premium period has expired
-                ch.premiumUntil?.let { premiumUntilString ->
-                    val cleanedDateString = if (premiumUntilString.startsWith("\"\$D")) {
-                        premiumUntilString.removePrefix("\"\$D").removeSuffix("\"")
-                    } else if (premiumUntilString.startsWith("\$D")) {
-                        premiumUntilString.removePrefix("\$D")
-                    } else if (premiumUntilString.startsWith("\"") && premiumUntilString.endsWith("\"") && premiumUntilString.length > 2) {
-                        premiumUntilString.substring(1, premiumUntilString.length - 1)
-                    } else {
-                        premiumUntilString
-                    }
-
-                    val premiumUntilDate = isoDateFormatter.tryParse(cleanedDateString)
-                    if (premiumUntilDate > 0) {
-                        // Include if premium period has expired
-                        return@filter System.currentTimeMillis() > premiumUntilDate
-                    }
-                }
-
-                // If we can't parse the premium until date, exclude the chapter for safety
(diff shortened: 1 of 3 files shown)


src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
```diff
@@ -388,11 +388,46 @@ class PoseidonScans : HttpSource() {
...
+                        premiumUntilString.removePrefix("\"\$D").removeSuffix("\"")
+                    } else if (premiumUntilString.startsWith("\$D")) {
+                        premiumUntilString.removePrefix("\$D")
+                    } else if (premiumUntilString.startsWith("\"") && premiumUntilString.endsWith("\"") && premiumUntilString.length > 2) {
+                        premiumUntilString.substring(1, premiumUntilString.length - 1)
+                    } else {
+                        premiumUntilString
+                    }
+
+                    val premiumUntilDate = isoDateFormatter.tryParse(cleanedDateString)
```
stevenyomi: Why not use `parseIsoDate()`?

src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
```diff
@@ -388,11 +388,46 @@ class PoseidonScans : HttpSource() {
         }
 
         return mangaDto.chapters
-            ?.filter { it.isPremium != true }
+            ?.filter { ch ->
```
stevenyomi: Fuse with the `mapNotNull` below.

stevenyomi (commented):
It's kinda weird to force hiding the premium chapters. You could just show them and prepend an emoji to show it is a premium chapter. And you can just use `premiumUntilDate` as `date_upload` so the available date is easy to see.

src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
Nyantad: My bad i forgot

stevenyomi (approved):

Nyantad:
> It's kinda weird to force hiding the premium chapters. You could just show them and prepend an emoji to show it is a premium chapter. And you can just use `premiumUntilDate` as `date_upload` so the available date is easy to see.

Yeah I get the idea but I was thinking "why show the chapter if it's not readable" if I had implemented a login system for the premium stuff then sure
But I see the point of showing future availability with premiumUntilDate as date_upload. It would give better visibility on the release schedule

stevenyomi:
> Yeah I get the idea but I was thinking "why show the chapter if it's not readable" if I had implemented a login system for the premium stuff then sure

The extension uses the same cookies as WebView, so if you log in through WebView, the extension can get the same results for many websites (unless they use LocalStorage to store the token).

Nyantad:
> > Yeah I get the idea but I was thinking "why show the chapter if it's not readable" if I had implemented a login system for the premium stuff then sure
> 
> The extension uses the same cookies as WebView, so if you log in through WebView, the extension can get the same results for many websites (unless they use LocalStorage to store the token).

Ok my bad, didn't know that and just saw there was a webview (on suwayomi) but yeah makes way more sense now

Issue #9963, opened 2025-08-03
Webtoons.com Display webtoon chapter number in the total Chs number as before


Qwerty0uiop:
### Source name

Webtoons.Com

### Source language

English

### Describe your suggested feature

The previous Versions of the webtoons. Com extension was really specific and accurately displayed the chapter number & the specific Ch number in the total Chs numbers as before.

As example, the really nice previous versions of the extension Displayed the chapter as : 

Episode 345 Ch. 353

We want this back again

The current one only display it just like that:
Episode 345.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/mihonapp/mihon/issues/new/choose).
- [x] I have updated the app to version **[0.15.3](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

### 

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/keiyoushi/extensions-source/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

AwkwardPeak7:
you can change it to display the chapter number instead of chapter title: in manga page > filters (three lines top right) > display > chapter number

the previous version of extension used to just display the chapter index along with number in title, now extension tries to parse the actual chapter number in title

previously:
```
"episode 12.1 ch. 13" -> 13
"episode 12 ch. 12" -> 12
```

now:
```
"episode 12.1" -> 12.1
"episode 12" -> 12
```

Qwerty0uiop:
> you can change it to display the chapter number instead of chapter title: in manga page > filters (three lines top right) > display > chapter number
> 
> the previous version of extension used to just display the chapter index along with number in title, now extension tries to parse the actual chapter number in title
> 
> previously:
> 
> ```
> "episode 12.1 ch. 13" -> 13
> "episode 12 ch. 12" -> 12
> ```
> 
> now:
> 
> ```
> "episode 12.1" -> 12.1
> "episode 12" -> 12
> ```

You are talking now about the mihon app itself, with this method, too. It doesn't make what i meant.

The method will show only The numbers without the title name.

[...]

PR #10451, opened 2025-09-08
Webtoons: Simplify chapter numbering to fix multilingual issues

Nyantad:
Replaced complex regex-based episode parsing with simple sequential numbering to resolve inconsistencies across different languages and improve maintainability.

Checklist:


AwkwardPeak7:
Can you provide any unresolved issues with the current implementation? I know it is a complex implementation, but I tried to address most issues raised for it.

The main issue with the sequential index is that it doesn't take into account extra chapters, which means they end up with an incorrect chapter number and any later chapter labelled 53 may have an index of 59.

Nyantad:
The main issue this pull request aims to solve is the one from #9963. The current regex-based implementation is not compatible with other languages.

For example, in French, chapter titles are formatted like [Saison 3] Ep. 235, which the current regex fails to parse. Supporting every language would require an overly complex and unmaintainable regex that would likely break with any future changes from Webtoons.

You are correct about the main drawback of my approach: the sequential index won't always match the displayed episode number, especially when there are extra chapters. This is a trade-off I made intentionally.

I believe that having a simple, reliable, and language-agnostic numbering system is a better long-term solution than a complex one that only works for a subset of languages and is hard to maintain. And also it break consistency for the old chapter downloaded.

Qwerty0uiop:
As @Nyantad mentioned, if you keep previously some old downloaded chapters, and you keeping track of the chapter number in whole Chs Numbers.

it break the consistency for the old chapter downloaded.

We prefer the original & official index, the same as the original Source.

You can check it, too by the Site / webview.

As example,

- The best version of the webtoons extension was displaying the chapter as :

[Season 3] Ep. 233 Ch. 650

The same way The original Source display it.
The Index and the chapter in the whole Manhwa here is: Ch 650

After you update the extension after my issue, it didn't fix the problem yet

The chapter display now is:

[Season 3] Ep. 233 (ch. 648)

That is only what i observed, i don't know what else get missed up.

I know you already implemented the other two display options, the one with title, and the other  with numbers (including ch 1, 2, 3, - 3,1 3,2 3,3 - 4)

But we prefer the original one, be identical to the original Source.

You can implement this as default, and keep the other options for who might like them as display options. If you can make it so.

But we like to keep the consistency for all the new chapters with the old downloaded chapters, too.

Thanks, to all of you for you efforts & all your hard work.

src/all/webtoons/src/eu/kanade/tachiyomi/extension/all/webtoons/Webtoons.kt
```diff
@@ -279,84 +278,25 @@ open class Webtoons(
...
                     append(Parser.unescapeEntities(episode.episodeTitle, false))
-                    append(" (ch. ", numberFormatter.format(episode.chapterNumber), ")")
+                    append(" (ch. ", globalChapterNumber.toInt(), ")")
                     if (episode.hasBgm) {
                         append(" ♫")
                     }
                 }
                 date_upload = episode.exposureDateMillis
-                chapter_number = episode.chapterNumber
+                chapter_number = globalChapterNumber
```
FourTOne5: I don't care much about the display name but `chapter_number` should not be the index

Nyantad: Yeah, ideally chapter_number should be the official number, not just the index.

However, the core issue is that reliably extracting this "official" number is difficult due to inconsistencies in how Webtoons labels its chapters, especially across different seasons, special episodes, and languages.

For example, even in @AwkwardPeak7's current implementation, a chapter like [Season 3] Ep. 235 (Season 3 Finale) is parsed as chapter 650, because the logic only sums the main episode counts of the three seasons.

But the fundamental problem here is that the current implementation does not work for all languages, and creating one that does would be difficult to maintain. That's why I proposed changing it to use the sequential index for the total count of all chapters, including bonuses, mini-episodes, etc.

Nyantad:
The core problem remains: the current parser only works for English. We can fix it for French with a simple, one-word change to the original regex:

```diff
- """(?:(s(eason)?|part|vol(ume)?)\s*\.?\s*(\d+).*?)?(.*?(mini|bonus|special).*?)?(e(p(isode)?)?|ch(apter)?)\s*\.?\s*(\d+(\.\d+)?)"""
+ """(?:(s(eason)?|saison|part|vol(ume)?)\s*\.?\s*(\d+).*?)?(.*?(mini|bonus|special).*?)?(e(p(isode)?)?|ch(apter)?)\s*\.?\s*(\d+(\.\d+)?)"""
```
This doesn't fix it for other languages, but would solve the problem for French users.

Nyantad:
Okay, I've updated the PR with a compromise.

I've kept the original parser and added the fix for French. My sequential numbering logic is now included as a new user-selectable option.

This gives us accurate parsing by default for english and french, with a simple fallback for those who prefer or need it.

src/all/webtoons/src/eu/kanade/tachiyomi/extension/all/webtoons/Webtoons.kt
```diff
@@ -431,6 +432,13 @@ open class Webtoons(
             summary = "Enable to load images in maximum quality."
             setDefaultValue(false)
         }.also(screen::addPreference)
+
+        SwitchPreferenceCompat(screen.context).apply {
+            key = USE_SEQUENTIAL_NUMBERING_KEY
+            title = "Use sequential chapter numbering"
+            summary = "Enable to use sequential numbering instead of official episode numbers. Useful for multilingual support."
```
AwkwardPeak7:
```suggestion
            summary = "Enable to use sequential numbering instead of official episode numbers."
```

AwkwardPeak7 (approved):

PR #10538, opened 2025-09-14
Filter Comick.kt

08shiro80:
add Genre Filter to Ignored Tags to make Permanent Blacklist, and apply on Search

Checklist:


Smol-Ame:
You need to bump the version so it registers as an update, did you even test your changes?

08shiro80:
yeah i did test it and it is working with the second little patch (worked before too)
all in tag filter from the extention settings are blacklisted if they are in it, Text search /Groups /Category/ Theme/ Format are Filtered only Demographic and  Origination are not Filtered

PR #10575, opened 2025-09-16
Fix StarboundScans revert to old loading by using src attribute 

Nyantad:
Update pageListParse to use src instead of data-src attribute
Yeah the site just use data-src for 4 days and just revert.

Checklist:

### DIFF
--- a/src/fr/starboundscans/src/eu/kanade/tachiyomi/extension/fr/starboundscans/StarboundScans.kt
+++ b/src/fr/starboundscans/src/eu/kanade/tachiyomi/extension/fr/starboundscans/StarboundScans.kt
@@ -8,6 +8,7 @@ import keiyoushi.utils.tryParse
 import okhttp3.Headers
 import okhttp3.Response
 import org.jsoup.nodes.Document
+import org.jsoup.nodes.Element
 import java.text.SimpleDateFormat
 import java.util.Locale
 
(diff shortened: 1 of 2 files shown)


lamaxama:
`element.attr("abs:src").ifEmpty { element.attr("abs:data-src") }` might be better.

src/fr/starboundscans/src/eu/kanade/tachiyomi/extension/fr/starboundscans/StarboundScans.kt
```diff
@@ -65,7 +65,7 @@ class StarboundScans : Madara(
 
     override fun pageListParse(document: Document): List<Page> {
         return document.select("img.wp-manga-chapter-img").mapIndexed { index, element ->
-            val imageUrl = element.attr("abs:data-src")
+            val imageUrl = element.attr("abs:src")
```
AwkwardPeak7:
```kt
    private fun Element.imgAttr(): String {
        return when {
            hasAttr("data-src") -> attr("abs:data-src")
            else -> attr("abs:src")
        }
    }
```
you can use this function to check for both

Nyantad:
Yeah my bad, thanks for the suggestion!

AwkwardPeak7 (approved):

PR #10740, opened 2025-09-26
[poseidonscans] - Fixes broken search due to site changes.

Nyantad:
The site moved from client-side to server-side search, which broke the extension.
  
Checklist:

### DIFF
--- a/src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
+++ b/src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
@@ -19,6 +19,7 @@ import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import okhttp3.Request
 import okhttp3.Response
 import org.jsoup.nodes.Document
+import java.net.URLDecoder
 import java.text.SimpleDateFormat
 import java.util.Locale
 import java.util.TimeZone
@@ -506,7 +507,7 @@ class PoseidonScans : HttpSource() {
                     ?.attr("srcset")
                     ?.substringBefore(" ")
                     ?.let {
-                        java.net.URLDecoder.decode(it, "UTF-8")
+                        URLDecoder.decode(it, "UTF-8")
                             .substringAfter("url=")
                             .substringBefore("&")
                     }
--- a/src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
+++ b/src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
@@ -12,7 +12,6 @@ import eu.kanade.tachiyomi.util.asJsoup
 import keiyoushi.utils.parseAs
 import keiyoushi.utils.tryParse
 import kotlinx.serialization.json.JsonObject
-import kotlinx.serialization.json.jsonArray
 import kotlinx.serialization.json.jsonObject
 import kotlinx.serialization.json.jsonPrimitive
 import okhttp3.HttpUrl.Companion.toHttpUrl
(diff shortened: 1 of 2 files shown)


src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
```diff
@@ -481,63 +480,51 @@ class PoseidonScans : HttpSource() {
...
-                val slug = mangaObject["slug"]?.jsonPrimitive?.content ?: return@mapNotNull null
-                val cover = mangaObject["coverImage"]?.jsonPrimitive?.content
+                val url = element.attr("href").takeIf { it.isNotBlank() } ?: return@mapNotNull null
+                val title = element.selectFirst("h2")?.text()?.takeIf { it.isNotBlank() } ?: return@mapNotNull null
+
+                val thumbnailUrlPath = element.selectFirst("img[alt]")
+                    ?.attr("srcset")
+                    ?.substringBefore(" ")
+                    ?.let {
+                        java.net.URLDecoder.decode(it, "UTF-8")
```
AwkwardPeak7: import

src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
```diff
@@ -481,63 +480,51 @@ class PoseidonScans : HttpSource() {
...
-
-        val allMangas = mangaListJsonArray.mapNotNull { mangaElement ->
+        val mangas = document.select("div.grid a.block.group").mapNotNull { element ->
             try {
-                val mangaObject = mangaElement.jsonObject
-                val title = mangaObject["title"]?.jsonPrimitive?.content ?: return@mapNotNull null
-                val slug = mangaObject["slug"]?.jsonPrimitive?.content ?: return@mapNotNull null
-                val cover = mangaObject["coverImage"]?.jsonPrimitive?.content
+                val url = element.attr("href").takeIf { it.isNotBlank() } ?: return@mapNotNull null
+                val title = element.selectFirst("h2")?.text()?.takeIf { it.isNotBlank() } ?: return@mapNotNull null
```
AwkwardPeak7, line 504: are these empty/null checks necessary?

src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
```diff
@@ -481,63 +480,51 @@ class PoseidonScans : HttpSource() {
...
+                    }
 
                 SManga.create().apply {
+                    this.setUrlWithoutDomain(url)
                     this.title = title
-                    setUrlWithoutDomain("/serie/$slug")
-                    this.thumbnail_url = cover?.takeIf { it.isNotBlank() }?.toApiCoverUrl()
+                    this.thumbnail_url = thumbnailUrlPath?.takeIf { it.isNotBlank() }?.toApiCoverUrl()
                 }
             } catch (e: Exception) {
```
AwkwardPeak7, line 520: I don't think we need to catch such exceptions

src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
Nyantad: i kept them as a safeguard this way it just skips a broken item instead of crashing

src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
Nyantad: same here just in case the site sends something unexpected, it makes the parser more resilient

AwkwardPeak7: if extension encounters something unexpected, then it is a bug in the extension. not catching such exceptions allows us/users to easily discover such bugs and report/fix them.

AwkwardPeak7 (approved):

src/fr/poseidonscans/src/eu/kanade/tachiyomi/extension/fr/poseidonscans/PoseidonScans.kt
Nyantad: my bad yeah good idea

PR #10846, opened 2025-10-03
Add BookWalker Global

TheUnlocked:
Checklist:

Adds an extension for [BookWalker Global](https://global.bookwalker.jp/). This is a pretty complicated extension as it needs to interface with a proprietary canvas-based renderer, but I've been testing it for the past few months and aside from some occasional issues with a page timing out and needing to be retried, it works quite reliably.


KazaBooo:
Is it not possible to fetch the real images/ebooks instead of from the canvas?

And would it also work with the japanese bw website or are there too many differences?

TheUnlocked:
It is unfortunately not possible (or at least not feasible) to fetch the images directly without a WebView. Images are fetched using dynamic URLs and are encoded using a complex and proprietary algorithm.

As for the Japanese site, I honestly just haven't tested it. My guess is that the image-reading logic would probably work on the Japanese site, but I'm not confident that the rest of it would since the two sites are at least superficially different.

MikeZeDev:
Bookwalker is using Publus. Publus sources are most of the time heavily obfuscated and deobfuscation is tricky because of character encoding i guess. If you are able to run a script from the Webview you could run ```pagescript``` from [here ](https://github.com/manga-download/haruneko/blob/publus/web/src/engine/websites/BookWalker.ts) and get page with pieces infos.


Its more or less how publus does things, the problem is the important classes & function names always changes.

TheUnlocked:
I have been burned by relying on mangled names in web scraping before, so I wanted to either figure out how the algorithm works (which is not easy) or just go around it and pull data from the canvas and interact with the page as if I was a user (the approach I took here). There's probably some performance cost to that, but there's a lot of development gain with not needing to figure out new names whenever they change.

src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
```diff
@@ -0,0 +1,872 @@
...
+            polymorphic(HoldBookEntityDto::class) {
+                subclass(SingleDto::class)
+                subclass(SeriesDto::class)
+            }
+        }
+    }
+
+    val app by injectLazy<Application>()
+
+    private val preferences by lazy {
```
AwkwardPeak7: please use `getPreferences` from `keiyoushi.utils` for consistency

src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
```diff
@@ -0,0 +1,872 @@
...
+                    match.groups[0]!!.value.replaceFirstChar(Char::titlecase),
+                    match.groups[1]!!.value.toFloat(),
+                )
+            }
+        }
+        // Cannot parse chapter number
+        return null
+    }
+
+    private inline fun <reified T> Response.parseAs(): T = use {
```
AwkwardPeak7: use `parseAs` from `keiyoushi.utils`. you can pass your own json instance as parameter to it.

src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
```diff
@@ -0,0 +1,872 @@
...
+    }
+
+    override fun mangaDetailsParse(document: Document): SManga =
+        throw UnsupportedOperationException()
+
+    private fun fetchSeriesMangaDetails(manga: SManga): Observable<SManga> {
+        return rxSingle {
+            val seriesPage = client.newCall(
+                GET("$baseUrl${manga.url}?order=release&np=1", callHeaders),
+            ).awaitSuccess()
```
AwkwardPeak7, line 416: You don't need to worry about running network calls in nested rx.Single and Coroutines; the fetch* methods are called on the IO thread on the app, so you can simply use blocking `.execute()` on a call and use the response.

src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
```diff
@@ -0,0 +1,872 @@
...
+
+    private fun SManga.updateDetailsFromSeriesPage(document: Document) = run {
+        // Take the thumbnail from the first chapter that is not on pre-order.
+        // Pre-order chapters often just have a gray rectangle with "NOW PRINTING" as their
+        // thumbnail, which doesn't look very pretty for the catalog.
+        thumbnail_url = document
+            .select(".o-tile:not(:has(.a-ribbon-pre-order)) .a-tile-thumb-img > img")
+            .attr("data-srcset")
+            .getHighestQualitySrcset()
+        title = document.selectFirst(".title-main-inner")?.ownText()?.cleanTitle().orEmpty()
```
AwkwardPeak7: title is mandatory, use non null assertion (`!!`) instead of `orEmpty()`

src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
```diff
@@ -0,0 +1,872 @@
...
+
+            val firstPage = validateLogin(getDocumentForPage(1))
+            val publishers = getAvailableFilterNames(firstPage, "side-publisher")
+                .joinToString()
+
+            val pageCount = firstPage.selectFirst(".pager-area li:has(+ .next) > a")
+                ?.ownText()?.toIntOrNull()
+                ?: 1
+
+            val laterPages = (2..pageCount).map { n -> async { getDocumentForPage(n) } }
```
AwkwardPeak7, line 533: see my earlier comment on nested rx.Single and Coroutines. you can use `runBlocking` here if you want async execution

src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
```diff
@@ -0,0 +1,872 @@
...
+        // This URL doesn't actually contain the image. It will be intercepted, and the actual image
+        // will be extracted from a webview of the URL being sent here.
+        return GET(
+            page.imageUrl!!.toHttpUrl().newBuilder()
+                .removeAllQueryParameters(CACHE_BUSTING_QUERY_PARAMETER)
+                .build()
+                .toString(),
+            callHeaders.newBuilder()
+                .set(HEADER_IS_REQUEST_FROM_EXTENSION, "true")
+                .set(HEADER_PAGE_INDEX, page.index.toString())
```
AwkwardPeak7: note that the app reindexes all pages, so this index might not be what you set in page list. Better to put the index somewhere in the url (like fragment) if it is required for the request.

src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
```diff
@@ -0,0 +1,872 @@
...
+                    throw Exception("Logged out, check website in WebView")
+                }
+            }
+
+            Log.d("bookwalker", "Successfully redirected to $redirectUrl")
+            redirectUrl
+        }
+    }
+
+    private fun openUrlInWebview(url: String) {
```
AwkwardPeak7: I would say don't do this. it is better to throw errors messages to tell users to login

src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalkerChapterReader.kt
```diff
@@ -0,0 +1,413 @@
...
+
+                @SuppressLint("SetJavaScriptEnabled")
+                settings.javaScriptEnabled = true
+                settings.domStorageEnabled = true
+                // Mobile vs desktop doesn't matter much, but the mobile layout has a longer page
+                // slider which allows for more accuracy when trying to jump to a particular page.
+                settings.userAgentString = USER_AGENT_MOBILE
+
+                webViewClient = object : WebViewClient() {
+                    override fun onLoadResource(view: WebView?, url: String?) {
```
AwkwardPeak7: disable/comment out these overrides for logging/debugging

AwkwardPeak7 (changes requested):
impressive work honestly

TheUnlocked:
Thanks for the review. I have other dev tasks atm but I'll probably get around to making the updates in the next few days.

src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
TheUnlocked: I'm okay with removing this if you really want to (it's already gated behind a config option disabled by default), but just some context on why it's there—

BookWalker has two "tiers" of login session, where one is long-lived and gives access to browse the site and your library as yourself, and the other is short-lived and gives access to read content and make purchases.

What makes this awkward is that when you try to open a chapter, you might get a toast saying to open the WebView to log in, but when you open the WebView, it looks like you already are logged in. In order to log in "properly" you either need to try to open the web reader in the WebView, or click on the "Cart" button. Not the end of the world, but pretty confusing to someone who doesn't know how it works.

If I allow the page list to be acquired without verifying that pages can be loaded and just let the image loading fail, the button to open a single page in WebView will properly direct you to the login page. However, it also means that every page that the client tries to load (1-2 on-screen plus probably around 3 preloaded) will send another request to BookWalker and fail, requiring the user to manually refresh all of those images after logging in or to back out of the chapter and then re-open it. Again, not the end of the world, but also not great.

Automatically opening the login page in a WebView when opening a chapter while logged out felt like the best experience in my testing, though it does of course also have its own downsides (e.g. when trying to download multiple chapters at once, and also it only works on clients where it's possible to invoke the tachiyomi webview activity) so I made it configurable and disabled by default.

AwkwardPeak7: you can say "open cart in webview" or similar to tell users what to do. The current approach is brittle, it will not work on forks where the activity is located somewhere else, or not work at all on projects like suwayomi.

src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
TheUnlocked: I never encountered any issues with this in testing, but I've changed it to read the index from the existing (but now renamed) cache-busting query parameter rather than from the `Page` object.

src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
TheUnlocked: I get that it's all being run on an IO thread, but wouldn't a blocking call still prevent other coroutines from running on that same thread?

AwkwardPeak7: use `runBlocking` for coroutine context and execute any `async` calls. usage of rx.java is discouraged, since it is deprecated on the app side.

AwkwardPeak7 (approved):

src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
TheUnlocked: My concern with `runBlocking` isn't that calls within the block couldn't run asynchronously, but that it would lock down the thread for use by coroutines _outside_ of the block (e.g. other parts of the application couldn't use that thread). I recognize that Mihon has switched to coroutines, which is actually why I created the `rxSingle` function to wrap a single around a coroutine. If this repo ever adopts the extensions-lib 1.5 suspend API, you can remove the `rxSingle` calls, remove a level of indentation, and everything should just work™.

AwkwardPeak7:
merging now, the rx.Single comment can be addressed later

src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
AwkwardPeak7: Extensions don't need to worry about blocking other parts of the app from fetch* methods. They are only spawned in a separate coroutine IO thread.

Issue #10198, opened 2025-08-20
Artlapsa has started uploading smut to the site


FangMin2005:
### Source name

Art Lapsa

### Source language

English 

### Describe your suggested feature

Can it be marked as 18+ pls

### Other details

Here's the Smut I'm talking about, https://artlapsa.com/series/d8b25d80-c85e-4860-b1eb-e12aac202f09

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/mihonapp/mihon/issues/new/choose).
- [x] I have updated the app to version **[0.15.3](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

### 

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/keiyoushi/extensions-source/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

PR #10870, opened 2025-10-04
Keyoapp: tweak selector; Asmodeus Scans: filter novel,fix info

### LINKED ISSUE #9471: Asmodeus Scans to not show Novel entries
### Source name

Asmodeus Scans

### Source language

English

Asmodeus Scans currently parses the `Type` tag on the site as a normal tag.

As such, parse if the `Type` is set to "Novel" and if so, do not display it.

### Other details

Currently, Asmodeus Scans shows not just Comic/Manga/Manhwa/Mangatoon entries but also the Novel entries.

This can confuse any users who just add the entry to their library without properly checking the tags.

As an example for confusion, take [this](https://asmotoon.com/series/6356e391182/) entry.
It is a manga entry and works as such but then there's also [this](https://asmotoon.com/series/6379d4caa5d/) which is the same series but of the Novel entry instead and as such, could not be read.

I myself got confused by this as there were two of the same series in this source yet it has different entries.
I saw that it had more chapters so I migrated it back to that other entry.
Only to find out when trying to read it that it was a `Novel` entry and not a `Manga` entry.

### Acknowledgements

### 

---

Riztard:
Asmodeus Scans: filter novel, add/fix genres, clean description - Closes #9471

Keyoapp:
- previous type selector was mistakenly named genre
- make genre selector val
- tweak the genre selector to exclude manga status from genre

Checklist:

### DIFF
--- a/src/en/asmotoon/src/eu/kanade/tachiyomi/extension/en/asmotoon/Asmotoon.kt
+++ b/src/en/asmotoon/src/eu/kanade/tachiyomi/extension/en/asmotoon/Asmotoon.kt
@@ -17,13 +17,13 @@ class Asmotoon : Keyoapp(
 
     override val genreSelector: String = ".gap-3 .gap-1 a"
 
-    override fun mangaDetailsParse(document: Document): SManga = SManga.create().apply {
+    override fun mangaDetailsParse(document: Document): SManga {
         return super.mangaDetailsParse(document).apply {
             genre = buildList {
                 document.selectFirst(typeSelector)?.text()?.replaceFirstChar {
                     if (it.isLowerCase()) {
                         it.titlecase(
-                            Locale.getDefault(),
+                            Locale.ENGLISH,
                         )
                     } else {
                         it.toString()
--- a/src/en/asmotoon/src/eu/kanade/tachiyomi/extension/en/asmotoon/Asmotoon.kt
+++ b/src/en/asmotoon/src/eu/kanade/tachiyomi/extension/en/asmotoon/Asmotoon.kt
@@ -1,9 +1,36 @@
 package eu.kanade.tachiyomi.extension.en.asmotoon
 
 import eu.kanade.tachiyomi.multisrc.keyoapp.Keyoapp
+import eu.kanade.tachiyomi.source.model.SManga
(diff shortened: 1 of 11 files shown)


xdwil:
#10198

might as well do simple `false to true` so don't have to bump again

src/en/asmotoon/src/eu/kanade/tachiyomi/extension/en/asmotoon/Asmotoon.kt
```diff
@@ -1,9 +1,36 @@
...
+) {
+    // filtering novel entries
+    override fun popularMangaSelector() = "div:contains(Trending) + div .group.overflow-hidden.grid:not(:has(.capitalize:contains(Novel)))"
+    override fun latestUpdatesSelector() = "div.grid > div.group:not(:has(.capitalize:contains(Novel)))"
+    override fun searchMangaSelector() = "#searched_series_page > button:not(:has(.capitalize:contains(Novel)))"
+
+    override val genreSelector: String = ".gap-3 .gap-1 a"
+
+    override fun mangaDetailsParse(document: Document): SManga = SManga.create().apply {
+        return super.mangaDetailsParse(document).apply {
```
vetleledaal: Do you need to create 2 SManga instances here? Consider removing `SManga.create().apply ...`

src/en/asmotoon/src/eu/kanade/tachiyomi/extension/en/asmotoon/Asmotoon.kt
```diff
@@ -1,9 +1,36 @@
...
+            genre = buildList {
+                document.selectFirst(typeSelector)?.text()?.replaceFirstChar {
+                    if (it.isLowerCase()) {
+                        it.titlecase(
+                            Locale.getDefault(),
```
vetleledaal: Avoid relying on the user's locale. Use `Locale.ENGLISH` if you need one.

Riztard: i copied those from the multi src, should i change it as well?

AwkwardPeak7 (approved):

PR #10963, opened 2025-10-09
Asura hide cover

Naputt1:
Screenshot 2025-10-09 at 14 15 02

Checklist:


AwkwardPeak7:
why? one extra page isn't a big deal. besides, this work under the assumption the first page will always be the "cover page", which may or may not remain true in the future.

Naputt1:
It's optional so you just have to turn it off the they decided the move the "cover page".

src/en/asurascans/src/eu/kanade/tachiyomi/extension/en/asurascans/AsuraScans.kt
```diff
@@ -297,16 +297,18 @@ class AsuraScans : ParsedHttpSource(), ConfigurableSource {
...
+                        .fragment("pageListParse")
+                        .build()
+                        .toString()
+                }
 
-            Page(i, imageUrl = newUrl ?: page.url)
-        }
+                Page(i, imageUrl = newUrl ?: page.url)
+            }
     }
```
AwkwardPeak7, line 312:
```suggestion
        val pageListFiltered = if (preferences.hideCoverPages()) { pageList.drop(0) } else { pageList }
        return pageListFiltered.mapIndexed { i, page ->
                val newUrl = page.url.toHttpUrlOrNull()?.run {
                    newBuilder()
                        .fragment("pageListParse")
                        .build()
                        .toString()
                }

                Page(i, imageUrl = newUrl ?: page.url)
            }
}
```
avoid double iteration

Amejonah1200:
Instead on relying on only the position, why not taking the size into account? Banners usually are horizontal while pages themselves are vertical.

src/en/asurascans/src/eu/kanade/tachiyomi/extension/en/asurascans/AsuraScans.kt
cuong-tran: What if pageList has 0 size?

vetleledaal:
The cover is unique and containts its own chapter number. I don't think it should be removed.

If any, you should hide the last page (`/images/EndDesign.webp`). If you want to implement that please open another PR. You can implement this without a user-facing option/preference.

PR #11048, opened 2025-10-13
add use the "Reload broken image" source

wbmins:
Checklist:

### DIFF
--- a/src/en/kagane/build.gradle
+++ b/src/en/kagane/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Kagane'
     extClass = '.Kagane'
-    extVersionCode = 4
+    extVersionCode = 5
     isNsfw = true
 }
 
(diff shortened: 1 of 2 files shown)


src/all/ehentai/src/eu/kanade/tachiyomi/extension/all/ehentai/EHentai.kt
AwkwardPeak7, line 1: add alternative/failsafe url as fragment to primary image url. in an interceptor, detect if it fails and try the url from fragment. This avoids having to call HEAD on each image before loading.

wbmins: I'm sorry, I don't understand this language very well. I repaired it through chatgpt. I will try it according to your instructions. Thank you.

PR #11161, opened 2025-10-20
MangaDistrict: custom title regex

cuong-tran:
- Add custom regex for cleaning title
- Fix chapter upload date got reset for `New` chapters

Checklist:

### DIFF
--- a/src/en/mangadistrict/src/eu/kanade/tachiyomi/extension/en/mangadistrict/MangaDistrict.kt
+++ b/src/en/mangadistrict/src/eu/kanade/tachiyomi/extension/en/mangadistrict/MangaDistrict.kt
@@ -2,6 +2,7 @@ package eu.kanade.tachiyomi.extension.en.mangadistrict
 
 import android.annotation.SuppressLint
 import android.content.SharedPreferences
+import android.widget.Toast
 import androidx.preference.EditTextPreference
 import androidx.preference.ListPreference
 import androidx.preference.PreferenceScreen
@@ -172,21 +173,18 @@ class MangaDistrict :
         }
     }
 
-    private var titleRegex: Regex =
-        Regex("\\([^()]*\\)|\\{[^{}]*\\}|\\[(?:(?!]).)*]|«[^»]*»|〘[^〙]*〙|「[^」]*」|『[^』]*』|≪[^≫]*≫|﹛[^﹜]*﹜|〖[^〖〗]*〗|\uD81A\uDD0D.+?\uD81A\uDD0D|《[^》]*》|⌜.+?⌝|⟨[^⟩]*⟩|/Official|/ Official", RegexOption.IGNORE_CASE)
-
     private fun SManga.cleanTitleIfNeeded() = apply {
-        if (isRemoveTitleVersion()) {
-            title = title.let {
-                var tempTitle = it
-                if (customRemoveTitle().isNotEmpty()) {
-                    tempTitle = tempTitle.replace(Regex(customRemoveTitle()), "")
+        title = title.let { originalTitle ->
+            var tempTitle = originalTitle
+            customRemoveTitle().takeIf { it.isNotEmpty() }?.let { customRegex ->
+                runCatching {
+                    tempTitle = tempTitle.replace(Regex(customRegex), "")
                 }
-                if (isRemoveTitleVersion()) {
-                    tempTitle = tempTitle.replace(titleRegex, "")
-                }
-                tempTitle.trim()
             }
+            if (isRemoveTitleVersion()) {
+                tempTitle = tempTitle.replace(titleRegex, "")
+            }
+            tempTitle.trim()
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/en/mangadistrict/src/eu/kanade/tachiyomi/extension/en/mangadistrict/MangaDistrict.kt
```diff
@@ -47,16 +52,8 @@ class MangaDistrict :
...
-    override fun popularMangaFromElement(element: Element): SManga {
-        return super.popularMangaFromElement(element).cleanTitleIfNeeded()
-    }
-
     override fun popularMangaNextPageSelector() = "div[role=navigation] span.current + a.page"
 
-    override fun latestUpdatesFromElement(element: Element): SManga {
-        return super.latestUpdatesFromElement(element).cleanTitleIfNeeded()
-    }
-
```
AwkwardPeak7, line 59: j2k and yokai will not cleanup the title in manga details, so we need to do it here

src/en/mangadistrict/src/eu/kanade/tachiyomi/extension/en/mangadistrict/MangaDistrict.kt
```diff
@@ -168,11 +168,18 @@ class MangaDistrict :
...
     }
 
-    private val titleVersion = Regex("\\(.*\\)")
-
     private fun SManga.cleanTitleIfNeeded() = apply {
-        if (isRemoveTitleVersion()) {
-            title = title.replace(titleVersion, "").trim()
+        title = title.let { originalTitle ->
+            var tempTitle = originalTitle
+            customRemoveTitle().takeIf { it.isNotEmpty() }?.let { customRegex ->
```
AwkwardPeak7, line 186: why custom? imo if there exists a better regex, then extension should be updated to include that.

cuong-tran: No real needs yet. It's just that Bato.to has it so I just want to make they have the same option.

cuong-tran: If user only want to remove a specific tag such as `(yoai)` but want to keep `(official)` then they will need custom regex.

src/en/mangadistrict/src/eu/kanade/tachiyomi/extension/en/mangadistrict/MangaDistrict.kt
cuong-tran: then, currently Bato.to only applies cleaning title to getDetails will also failed? Then should I also add that to Bato.to?

AwkwardPeak7: > then, currently Bato.to only applies cleaning title to getDetails will also failed? Then should I also add that to Bato.to?

yeah

src/en/mangadistrict/src/eu/kanade/tachiyomi/extension/en/mangadistrict/MangaDistrict.kt
AwkwardPeak7: > If user only want to remove a specific tag such as `(yoai)` but want to keep `(official)` then they will need custom regex.

good point, ig it is fine then

src/en/mangadistrict/src/eu/kanade/tachiyomi/extension/en/mangadistrict/MangaDistrict.kt
```diff
@@ -83,9 +80,12 @@ class MangaDistrict :
 
     override fun chapterFromElement(element: Element): SChapter {
         return super.chapterFromElement(element).apply {
-            val urlKey = url.urlKey()
-            preferences.dates[urlKey]?.also {
-                date_upload = it
+            if (date_upload == 0L) {
```
AwkwardPeak7: also, remove the date from preference if it exists there, in else condition

src/en/mangadistrict/src/eu/kanade/tachiyomi/extension/en/mangadistrict/MangaDistrict.kt
```diff
@@ -302,6 +311,46 @@ class MangaDistrict :
...
+            setDefaultValue("")
+
+            val validate = { str: String ->
+                runCatching { Regex(str) }
+                    .map { true to "" }
+                    .getOrElse { false to it.message }
+            }
+
+            setOnBindEditTextListener { editText ->
+                editText.addTextChangedListener(
```
AwkwardPeak7, line 339: I am not sure if this would work on suwayomi, need to test. why not just do it in `setOnPreferenceChangeListner` and don't save new value if it is invalid regex

cuong-tran: this is to provide a feedback right when user edit so he has a chance to fix it, or else when dialog close it will lose the whole complex editing regex.

cuong-tran: Tested on Suwayomi, both `setOnBindEditTextListener` & `setOnPreferenceChangeListener` don't have any effect. They won't prevent bad format regex and allow setting preference always. The cleaning title won't work but also doesn't affect extension working.

AwkwardPeak7 (approved):

PR #11339, opened 2025-10-30
AsuraScans: add auth and premium support

Vacbo:
## Summary
- add WebView login activity with cookie fallback for Asura Scans
- integrate premium chapter unlock flow and hi-res image fallback
- adjust gradle + DTO wiring to support authenticated requests

## Testing
- [x] Tachimanga (iOS) premium chapter unlock
- [x] Android WebView login (not available in current test setup)


src/en/asurascans/src/eu/kanade/tachiyomi/extension/en/asurascans/AsuraScans.kt
```diff
@@ -65,6 +106,26 @@ class AsuraScans : ParsedHttpSource(), ConfigurableSource {
     private val json: Json by injectLazy()
 
     override val client = network.cloudflareClient.newBuilder()
+        .cookieJar(
+            object : CookieJar {
```
AwkwardPeak7, line 90: the default client should already have a cookie jar which is linked to webview cookies.

src/en/asurascans/src/eu/kanade/tachiyomi/extension/en/asurascans/AsuraScans.kt
```diff
@@ -75,14 +136,16 @@ class AsuraScans : ParsedHttpSource(), ConfigurableSource {
         val request = chain.request()
 
         if (preferences.forceHighQuality() && !failedHighQuality && request.url.fragment == "pageListParse") {
-            OPTIMIZED_IMAGE_PATH_REGEX.find(request.url.encodedPath)?.also { match ->
-                val (id, page) = match.destructured
-                val newUrl = request.url.newBuilder()
-                    .encodedPath("/storage/media/$id/$page.webp")
+            STANDARD_IMAGE_PATH_REGEX.find(request.url.encodedPath)?.also { match ->
```
AwkwardPeak7, line 131: why invert the logic?

AwkwardPeak7 (changes requested):
This is too complex.
- You should not add webview activity to the extension. throw exceptions asking users to login via webview in app. 
- Is the cookie/token based login necessary? Imo login via cookie from webview should be enough.

src/en/asurascans/src/eu/kanade/tachiyomi/extension/en/asurascans/AsuraScans.kt
Vacbo: Others extensions use it to keep the cookies synced with the WebView too:

- [Tapastic](https://github.com/keiyoushi/extensions-source/blob/main/src/en/tapastic/src/eu/kanade/tachiyomi/extension/en/tapastic/Tapastic.kt#L56) - "Syncs okhttp with webview cookies, allowing logged-in users do logged-in stuff"
- [ComX](https://github.com/keiyoushi/extensions-source/blob/main/src/ru/comx/src/eu/kanade/tachiyomi/extension/ru/comx/ComX.kt#L70) - "Syncs okhttp with WebView cookies, allowing logged-in users do logged-in stuff"

Vacbo:
I agree that the WebView launch code adds complexity, it was a leftover from trying to support iOS. But, it's worth keeping cookie/token paste for cross-platform support:
- at least for iOS/Tachimanga: WebView cookies aren't shared with extensions, so I used cookie paste as workaround to make it work.
- I am not sure but other clients for desktop (Tachidesk) may not share WebView cookies like in Android.

I will simplify the WebView login logic, as it makes more sense for Android.

Vacbo:
> I agree that the WebView launch code adds complexity, it was a leftover from trying to support iOS. But, it's worth keeping cookie/token paste for cross-platform support:
> 
> * at least for iOS/Tachimanga: WebView cookies aren't shared with extensions, so I used cookie paste as workaround to make it work.
> * I am not sure but other clients for desktop (Tachidesk) may not share WebView cookies like in Android.
> 
> I will simplify the WebView login logic, as it makes more sense for Android.

I fixed the WebView cookies not being shared. It was bug. Skill Issue lol. Since it worked, I will remove the cookie overide

src/en/asurascans/src/eu/kanade/tachiyomi/extension/en/asurascans/AsuraScans.kt
AwkwardPeak7: yeah but the app already syncs webview and okhttp cookies. does it not work without this cookiejar?
https://github.com/mihonapp/mihon/blob/f9793d33233ceb7729fa9cfe5ae55b45a372c54c/core/common/src/main/kotlin/eu/kanade/tachiyomi/network/NetworkHelper.kt#L24
https://github.com/mihonapp/mihon/blob/f9793d33233ceb7729fa9cfe5ae55b45a372c54c/core/common/src/main/kotlin/eu/kanade/tachiyomi/network/AndroidCookieJar.kt#L8


idk about those extensions in the repo, would need to invistigate.

src/en/asurascans/src/eu/kanade/tachiyomi/extension/en/asurascans/AsuraScans.kt
```diff
@@ -293,24 +372,157 @@ class AsuraScans : ParsedHttpSource(), ConfigurableSource {
...
+    private fun appendPageFragment(url: String): String {
+        return url.toHttpUrlOrNull()?.newBuilder()
+            ?.fragment("pageListParse")
+            ?.build()
+            ?.toString()
+            ?: url
+    }
+
+    private fun buildApiPostRequest(url: String, body: RequestBody): Request {
+        val builder = Request.Builder()
```
AwkwardPeak7, line 438: nitpick: use `POST` helper from the ext-lib

Vacbo: Sure

Vacbo: I will check if there is a way to do it without the jar

AwkwardPeak7 (approved):

AwkwardPeak7:
I  have reverted this pr in 2f71e9f8e349e9f3a306ff12e40d5350a6020665 as it was crashing the app for non logged in users. feel free to open a new pr to add it back with a fix.

PR #11408, opened 2025-11-03
Fix GreenShit: remove unnecessary x-client-hash and simplify authentication

FelipeGFA:
- Removed generateToken() function that generated HMAC hash for x-client-hash header
- API works with Bearer token only, x-client-hash is not required
- Removed unused imports (Base64, Mac, SecretKeySpec)
- Removed SECRET_KEY constant
- Added path field to PageDto to match API response
- Simplified pageListRequestMobile to use only authorization header
- User login functionality remains intact and automatic
Checklist:


lib-multisrc/greenshit/src/eu/kanade/tachiyomi/multisrc/greenshit/GreenShitDto.kt
```diff
@@ -226,6 +226,7 @@ class PageDto(
     val src: String,
     @SerialName("numero")
     val number: Float? = null,
+    val path: String? = null,
```
AwkwardPeak7: where is this being used? if not used, no need to put it in dto

src/pt/sussyscan/src/eu/kanade/tachiyomi/extension/pt/sussyscan/SussyToons.kt
```diff
@@ -11,7 +11,7 @@ class SussyToons : GreenShit(
 ) {
     override val id = 6963507464339951166
 
-    override val versionId = 2
+    override val versionId = 3
```
AwkwardPeak7: did they change the manga urls/ids? since bumping this `versionId` forces all users to migrate back to the same source to use it, as old mangas in library are invalidated

src/pt/sussyscan/build.gradle
```diff
@@ -3,7 +3,7 @@ ext {
     extClass = '.SussyToons'
     themePkg = 'greenshit'
     baseUrl = 'https://www.sussytoons.wtf'
-    overrideVersionCode = 56
+    overrideVersionCode = 57
```
AwkwardPeak7: you bumped the multisrc baseversion, no need to bump this too

AwkwardPeak7 (changes requested):

AwkwardPeak7 (approved):

PR #11523, opened 2025-11-10
Increment version code from 11 to 12

Fdjcm2:
Checklist:


src/es/leercapitulo/build.gradle
```diff
@@ -1,7 +1,7 @@
 ext {
     extName = 'LeerCapitulo'
     extClass = '.LeerCapitulo'
-    extVersionCode = 11
+    extVersionCode = 12
```
cuong-tran, line 4: Don't create 2 PR for 2 commits of same extension. Commit this in  #11524 instead.

PR #11588, opened 2025-11-12
Koharu: Add mirror selection to preference page

### LINKED ISSUE #11587: Koharu: Add mirror selection to preference page
### Source name

Koharu

### Source language

All

Add mirror selection to preference page. Current issue with the source is, schalenetwork will redirect to one of their mirrors everytime we open WebView. This is fine on desktop where it remembers the last redirected mirror, but on Mihon's WebView, it may redirect to a different domain every time and we have to wait for CloudFlare again, which wastes time.

### Other details

I'll make a PR for this.

### Acknowledgements

### 

---

anewadventure:
Closes #11587.

Checklist:

### DIFF
--- a/src/all/koharu/src/eu/kanade/tachiyomi/extension/all/koharu/Koharu.kt
+++ b/src/all/koharu/src/eu/kanade/tachiyomi/extension/all/koharu/Koharu.kt
@@ -505,7 +505,14 @@ class Koharu(
         const val PREFIX_ID_KEY_SEARCH = "id:"
         private const val PREF_MIRROR = "pref_mirror"
         private const val MIRROR_PREF_DEFAULT = "schale.network"
-        private val mirrors = arrayOf(MIRROR_PREF_DEFAULT)
+        private val mirrors = arrayOf(
+            MIRROR_PREF_DEFAULT,
+            "anchira.to",
+            "gehenna.jp",
+            "niyaniya.moe",
+            "shupogaki.moe",
+            "hdoujin.net",
+        )
         private const val PREF_IMAGERES = "pref_image_quality"
         private const val PREF_REM_ADD = "pref_remove_additional"
         private const val PREF_EXCLUDE_TAGS = "pref_exclude_tags"
--- a/src/all/koharu/src/eu/kanade/tachiyomi/extension/all/koharu/Koharu.kt
+++ b/src/all/koharu/src/eu/kanade/tachiyomi/extension/all/koharu/Koharu.kt
@@ -58,11 +58,27 @@ class Koharu(
     private val searchLang: String = "",
 ) : HttpSource(), ConfigurableSource {
 
-    private val preferences: SharedPreferences by getPreferencesLazy()
+    private val preferences: SharedPreferences by getPreferencesLazy {
+        val server = getString(PREF_MIRROR, MIRROR_PREF_DEFAULT) ?: MIRROR_PREF_DEFAULT
+        val isValidMirror = server in mirrors || server.toIntOrNull() != null
+        if (!isValidMirror) {
+            edit().also { editor ->
+                editor.putString(PREF_MIRROR, MIRROR_PREF_DEFAULT)
+            }.apply()
+        }
+    }
 
     override val name = "SchaleNetwork"
 
-    override val baseUrl = "https://${preferences.getString(PREF_MIRROR, MIRROR_PREF_DEFAULT)!!}"
+    override val baseUrl: String
+        get() {
+            val preferenceValue = preferences.getString(PREF_MIRROR, MIRROR_PREF_DEFAULT) ?: MIRROR_PREF_DEFAULT
+            val mirror = preferenceValue.toIntOrNull()?.let { index ->
+                mirrors[index.coerceAtMost(mirrors.lastIndex)]
+            } ?: preferenceValue.takeIf { it in mirrors } ?: MIRROR_PREF_DEFAULT
+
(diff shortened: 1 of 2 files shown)


src/all/koharu/src/eu/kanade/tachiyomi/extension/all/koharu/Koharu.kt
```diff
@@ -57,9 +58,11 @@ class Koharu(
     private val searchLang: String = "",
 ) : HttpSource(), ConfigurableSource {
 
+    private val preferences: SharedPreferences by getPreferencesLazy()
+
     override val name = "SchaleNetwork"
 
-    override val baseUrl = "https://schale.network"
+    override val baseUrl = "https://${preferences.getString(PREF_MIRROR, MIRROR_PREF_DEFAULT)!!}"
```
cuong-tran: Add a fall-back to default if the current one saved in preference not exists in the `mirrors` list (removed, incorrect...)

anewadventure: I think you might have pulled the wrong commit of the code? Latest commit in my branch already have the code like you suggested.

cuong-tran: I mean, say if user selected an alternative domain from the list. Then on next update of this extension, that domain might be invalid and removed, but then user stuck with theirs saved domain instead of reset to new domain.

Add an migrate for old invalid domain in preferences.

src/all/koharu/src/eu/kanade/tachiyomi/extension/all/koharu/Koharu.kt
```diff
@@ -57,9 +58,11 @@ class Koharu(
     private val searchLang: String = "",
 ) : HttpSource(), ConfigurableSource {
 
+    private val preferences: SharedPreferences by getPreferencesLazy()
```
cuong-tran:
```suggestion
    private val preferences: SharedPreferences by getPreferencesLazy {
        val server = getString(PREF_MIRROR, MIRROR_PREF_DEFAULT)

        if (server !in mirrors) {
            edit().also { editor ->
                editor.putString(PREF_MIRROR, MIRROR_PREF_DEFAULT)
            }.apply()
        }
    }
```

AwkwardPeak7:
please implement the domain preference where it saves the index instead of url directly in preference. 
https://github.com/keiyoushi/extensions-source/blob/main/src/all/comicklive/src/eu/kanade/tachiyomi/extension/all/comicklive/Comick.kt#L420-L427
https://github.com/keiyoushi/extensions-source/blob/main/src/all/comicklive/src/eu/kanade/tachiyomi/extension/all/comicklive/Comick.kt#L420-L427

or you can also do what Cuong suggested

cuong-tran:
Still if saving the index, need to have some sort of fallback or you will get out-of-bound crash if the saved index is bigger than the actual domain list, as shown in here:

https://github.com/keiyoushi/extensions-source/blob/eb52bf8393bce9befc7b60d998a8887da60d84d5/src/all/comicklive/src/eu/kanade/tachiyomi/extension/all/comicklive/Comick.kt#L52

src/all/koharu/src/eu/kanade/tachiyomi/extension/all/koharu/Koharu.kt
```diff
@@ -58,11 +58,27 @@ class Koharu(
     private val searchLang: String = "",
 ) : HttpSource(), ConfigurableSource {
 
-    private val preferences: SharedPreferences by getPreferencesLazy()
+    private val preferences: SharedPreferences by getPreferencesLazy {
+        val server = getString(PREF_MIRROR, MIRROR_PREF_DEFAULT) ?: MIRROR_PREF_DEFAULT
+        val isValidMirror = server in mirrors || server.toIntOrNull() != null
```
cuong-tran: If you save mirror's index in preference, then `server` here need to be changed.
Actually, you only need 1 way, either saving index plus using `coerceAtMost`, OR the preferences migration. No need using both.

lemcapybara:
If it hasn't already been done, would it be possible to manually add mirrors to an existing mirror list? So that there's no need to create pull requests just to add a mirror. For example, there's an existing mirror list, and it will be displayed by default when installing an extension, but the user can quickly add a new mirror as new mirror appears.

cuong-tran:
@lemcapybara that only applies for 1 user. a PR still need for all users.

src/all/koharu/src/eu/kanade/tachiyomi/extension/all/koharu/Koharu.kt
```diff
@@ -457,6 +466,20 @@ class Koharu(
...
+        ListPreference(screen.context).apply {
+            key = PREF_MIRROR
+            title = "Preferred Mirror"
+            entries = mirrors
+            entryValues = mirrors
+            setDefaultValue(MIRROR_PREF_DEFAULT)
+            summary = "%s"
+
+            setOnPreferenceChangeListener { _, _ ->
+                Toast.makeText(screen.context, "Restart the app to apply changes", Toast.LENGTH_LONG).show()
```
cuong-tran, line 478: Should be fine now. Just 1 improvement suggestion is the Restart can be avoided if you can make `lazyHeaders` also update when baseUrl is changed.

vetleledaal (approved):

PR #11648, opened 2025-11-14
Hentaifox: fix "Top" categories filter

SilverBeamx:
Checklist:

This PR fixes the "Top" categories filter as the site has recently enforced the usage of the "Referer" header for Sidebar requests.

### DIFF
--- a/src/all/hentaifox/src/eu/kanade/tachiyomi/extension/all/hentaifox/HentaiFox.kt
+++ b/src/all/hentaifox/src/eu/kanade/tachiyomi/extension/all/hentaifox/HentaiFox.kt
@@ -143,7 +143,7 @@ class HentaiFox(
         }
         return xhrHeaders.newBuilder()
             .add("X-Csrf-Token", csrfToken)
-            .add("Referer", baseUrl)
+            .add("Referer", "$baseUrl/")
             .build()
     }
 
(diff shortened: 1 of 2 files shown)


src/all/hentaifox/src/eu/kanade/tachiyomi/extension/all/hentaifox/HentaiFox.kt
```diff
@@ -143,6 +143,7 @@ class HentaiFox(
         }
         return xhrHeaders.newBuilder()
             .add("X-Csrf-Token", csrfToken)
+            .add("Referer", baseUrl)
```
vetleledaal: Minor thing, all browsers use a `/` suffix here.
```suggestion
            .add("Referer", "$baseUrl/")
```

SilverBeamx: I checked from Postman and the correct url is the one without the trailing slash. I don't know if the slash is for some kind of manipulation/formal notation but as far as i can see both Postamn and Chrome DevTools agree that the slash shouldn't be there. If i am wrong lmk and i'll update it.

vetleledaal: Checking from Brave:
image

MDN example for the `Referer` also has a trailing slash for the root page: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Referer#examples

SilverBeamx: Ok, i was confidently incorrect then, my bad, i'll just merge your suggestion

SilverBeamx: I even double checked, not sure what i saw, sorry once again

vetleledaal (approved):

PR #11702, opened 2025-11-17
Update Extension Spectral Scan to Nexus Toons

KaeneXx:
SpectralScan is no longer called that; it's now called Nexus Toons.

Change the extension names, icons, and URL.

Checklist:


src/pt/nexustoons/res/mipmap-xxxhdpi/ic_launcher.png
vetleledaal, line 1: Use the icon generator mentioned in the contributing guidelines.

src/pt/nexustoons/src/eu/kanade/tachiyomi/extension/pt/spectralscan/NexusToons.kt
vetleledaal, line 26: Since the name has been changed it will produce a new `id`. This means users have to migrate all existing manga in their library. If the paths are the same you should pin the `id` to avoid migration. When pinning `id` you can remove `versionId` -- it is no longer used.

If the paths are not the same you don't need to keep the `id`, you can also remove `versionId` since you starting from a new sequence.

src/pt/nexustoons/src/eu/kanade/tachiyomi/extension/pt/spectralscan/NexusToons.kt
```diff
@@ -1,4 +1,4 @@
-package eu.kanade.tachiyomi.extension.pt.spectralscan
+package eu.kanade.tachiyomi.extension.pt.nexustoons
```
vetleledaal, line 1: You changed the package name, but not the path. I'm surprised it compiled at all. Did you test it?

vetleledaal (changes requested):

PR #11710, opened 2025-11-18
Kagane: add support for Tachimanga

lamaxama:
Checklist:

### DIFF
--- a/src/en/kagane/src/eu/kanade/tachiyomi/extension/en/kagane/Kagane.kt
+++ b/src/en/kagane/src/eu/kanade/tachiyomi/extension/en/kagane/Kagane.kt
@@ -255,17 +255,15 @@ class Kagane : HttpSource(), ConfigurableSource {
         add("Referer", "$baseUrl/")
     }.build()
 
-    private fun getCertificate(): String {
-        return client.newCall(GET("$apiUrl/api/v1/static/bin.bin", apiHeaders)).execute()
+    private fun getCertificate(url: String): String {
+        return client.newCall(GET(url, apiHeaders)).execute()
             .body.bytes()
             .toBase64()
     }
 
-    private fun getFairPlayCertificate(): String {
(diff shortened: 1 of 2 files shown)


src/en/kagane/src/eu/kanade/tachiyomi/extension/en/kagane/Kagane.kt
```diff
@@ -261,6 +261,12 @@ class Kagane : HttpSource(), ConfigurableSource {
             .toBase64()
     }
 
+    private fun getFairPlayCertificate(): String {
```
vetleledaal: Static? Use `by lazy` to avoid multiple requests.

src/en/kagane/src/eu/kanade/tachiyomi/extension/en/kagane/Kagane.kt
```diff
@@ -314,21 +324,39 @@ class Kagane : HttpSource(), ConfigurableSource {
...
 
                     async function getData() {
-                        const g = base64ToArrayBuffer("${getCertificate()}");
-                        let t = await navigator.requestMediaKeySystemAccess("com.widevine.alpha", [{
-                          initDataTypes: ["cenc"],
-                          audioCapabilities: [],
-                          videoCapabilities: [{
-                            contentType: 'video/mp4; codecs="avc1.42E01E"'
-                          }]
+                        let widevine = detectDRMSupport() !== 'fairplay';
```
vetleledaal, line 325: If `detectDRMSupport()` returns `null` it will set `widevine` to `true`, seems incorrect?

lamaxama: This is intentional. According to the site's source code, the site only supports two DRM systems: Widevine for Chrome and FairPlay for Safari. The detectDRMSupport (which is extracted from the site's source code) can return "widevine", "fairplay", or null. In case of a null, default to the Widevine logic as a fallback.

AwkwardPeak7 (approved):

PR #11757, opened 2025-11-22
[RU]Nudemoon fix image list parse

abubaca4:
Checklist:

### DIFF
--- a/src/ru/nudemoon/build.gradle
+++ b/src/ru/nudemoon/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Nude-Moon'
     extClass = '.Nudemoon'
-    extVersionCode = 25
+    extVersionCode = 26
     isNsfw = true
 }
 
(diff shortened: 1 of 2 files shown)


AwkwardPeak7 (approved):

abubaca4:
@AwkwardPeak7 can you help me? Is any way to rotate image?
Site in server side rotate some images and the rotate it's back with css.
изображение

AwkwardPeak7:
you would have to write an okhttp interceptor to transform the image response. pass necessary data/parameters as url fragment in pagelistparse, and intercept those requests in the interceptor to apply the transform.

abubaca4:
> you would have to write an okhttp interceptor to transform the image response. pass necessary data/parameters as url fragment in pagelistparse, and intercept those requests in the interceptor to apply the transform.

What fun I should override? Is any example in other source?

AwkwardPeak7:
> What fun I should override? Is any example in other source?
https://github.com/keiyoushi/extensions-source/blob/71001bb959941df040242d879a9f88c132ca8693/src/th/manhuakey/src/eu/kanade/tachiyomi/extension/th/manhuakey/ManhuaKey.kt#L30-L108

your would obviously be simpler, as it would only need to flip the image response

PR #11721, opened 2025-11-19
Add Comic Grast and SeedRandom utils

manti-X:
geo-blocked in case you’re wondering

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [ ] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [ ] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [x] Have removed `web_hi_res_512.png` when adding a new extension

PR #11767, opened 2025-11-23
Ci: fix infinite recusrsion in getting dependent extensions

AwkwardPeak7:
should fix ci error in #11721

### DIFF
--- a/buildSrc/src/main/kotlin/Extensions.kt
+++ b/buildSrc/src/main/kotlin/Extensions.kt
@@ -7,34 +7,23 @@ var ExtensionAware.baseVersionCode: Int
     get() = extra.get("baseVersionCode") as Int
     set(value) = extra.set("baseVersionCode", value)
 
-private val reverseDependencyCache = mutableMapOf<String, Set<Project>>()
-private var reverseDependencyCacheInitialized = false
-
-private fun Project.buildReverseDependencyCache() {
-    if (reverseDependencyCacheInitialized) return
-    reverseDependencyCacheInitialized = true
-
-    // Build reverse dependency map
-    val map = mutableMapOf<String, MutableSet<Project>>()
-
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/Extensions.kt
```diff
@@ -7,30 +7,58 @@ var ExtensionAware.baseVersionCode: Int
     get() = extra.get("baseVersionCode") as Int
     set(value) = extra.set("baseVersionCode", value)
 
-fun Project.getDependents(): Set<Project> {
-    val dependentProjects = mutableSetOf<Project>()
+private val reverseDependencyCache = mutableMapOf<String, Set<Project>>()
+private var reverseDependencyCacheInitialized = false
```
vetleledaal: Could make `reverseDependencyCache` a nullable var instead, also not sure if it needs to be mutable -- you only initialize it once. Or you could possibly check the size of the map (only if `buildReverseDependencyCache` always adds at least 1 item).

---

Not sure how to test this part, but Gemini 3 Pro suggests:
```kt
private var reverseDependencyCache: Map<String, Set<Project>>? = null

fun Project.getDependents(): Set<Project> {
    if (reverseDependencyCache == null) {
        reverseDependencyCache = rootProject.allprojects
            .flatMap { p ->
                p.configurations.flatMap { config ->
                    config.dependencies
                        .filterIsInstance<ProjectDependency>()
                        .map { dep -> dep.path to p }
                }
            }
            .groupBy({ it.first }, { it.second })
            .mapValues { (_, projects) -> projects.toSet() }
    }

    return reverseDependencyCache?.get(path).orEmpty()
}
```

vetleledaal (dismissed):

vetleledaal (approved):

PR #11788, opened 2025-11-24
MangaFreak: Fix thumbnails and improve Latest Updates source

monel14:
Checklist:


src/en/mangafreak/src/eu/kanade/tachiyomi/extension/en/mangafreak/Mangafreak.kt
```diff
@@ -57,15 +57,37 @@ class Mangafreak : ParsedHttpSource() {
...
+        val url = if (page == 1) {
+            baseUrl
+        } else {
+            // Page 2 on Latest_Releases is actually /Latest_Releases/2
+            // But since we use Home for page 1, we might miss some items or have duplicates
+            // if we jump straight to page 2 of Latest_Releases.
+            // However, typically users just want the latest stuff.
+            "$baseUrl/Latest_Releases/$page"
+        }
+        return GET(url, headers)
```
vetleledaal, line 70: I don't really get this change. The manga on the home page seems to be a subset of `Latest_Releases`.

src/en/mangafreak/src/eu/kanade/tachiyomi/extension/en/mangafreak/Mangafreak.kt
```diff
@@ -57,15 +57,37 @@ class Mangafreak : ParsedHttpSource() {
...
+    override fun latestUpdatesSelector(): String = "div.latest_item, div.latest_releases_item"
     override fun latestUpdatesFromElement(element: Element): SManga = SManga.create().apply {
-        thumbnail_url = element.select("img").attr("abs:src").replace("mini", "manga").substringBeforeLast("/") + ".jpg"
-        element.select("a").apply {
-            title = first()!!.text()
-            url = attr("href")
+        // Fix thumbnail URL: replace "mini_images" with "manga_images" and remove dimensions (e.g. /175x245)
+        thumbnail_url = element.select("img").attr("abs:src")
+            .replace("mini_images", "manga_images")
+            .replace(Regex("/\\d+x\\d+$"), ".jpg")
```
vetleledaal: Extract regex to avoid recompilation.

src/en/mangafreak/src/eu/kanade/tachiyomi/extension/en/mangafreak/Mangafreak.kt
```diff
@@ -57,15 +57,37 @@ class Mangafreak : ParsedHttpSource() {
...
     override fun latestUpdatesNextPageSelector(): String = popularMangaNextPageSelector()
-    override fun latestUpdatesSelector(): String = "div.latest_releases_item"
+    override fun latestUpdatesSelector(): String = "div.latest_item, div.latest_releases_item"
     override fun latestUpdatesFromElement(element: Element): SManga = SManga.create().apply {
-        thumbnail_url = element.select("img").attr("abs:src").replace("mini", "manga").substringBeforeLast("/") + ".jpg"
-        element.select("a").apply {
-            title = first()!!.text()
-            url = attr("href")
+        // Fix thumbnail URL: replace "mini_images" with "manga_images" and remove dimensions (e.g. /175x245)
+        thumbnail_url = element.select("img").attr("abs:src")
```
vetleledaal: You only need to match one thumbnail, so you should use `.selectFirst()`.

src/en/mangafreak/src/eu/kanade/tachiyomi/extension/en/mangafreak/Mangafreak.kt
vetleledaal: Consider using `HttpUrl` to make these changes. You can check if `pathSegments[0]` is `mini_images` and then use `pathSegments[1]` as a value in the new request.

There are probably multiple ways to achieve this, but something like this could work:
```kt
val higherQualityImageUrl = url
    ?.takeIf { baseUrl.toHttpUrl().topPrivateDomain() == it.topPrivateDomain() }
    ?.takeIf { it.pathSegments.size >= 2 }
    ?.takeIf { it.pathSegments[0] == "mini_images" }
    ?.run {
        val slug = pathSegments[1]
        newBuilder()
            .encodedPath("/")
            .addPathSegment("manga_images")
            .addPathSegment("$slug.jpg")
            .build()
    }
```

src/en/mangafreak/src/eu/kanade/tachiyomi/extension/en/mangafreak/Mangafreak.kt
monel14: The main reason for this change is that the Latest_Releases page does not strictly respect the chronological order, which confuses users.

On Latest_Releases, the list is flat and we observed inconsistencies where items from 'Yesterday' could appear mixed with or before items from 'Today' depending on how the site bumps them. Since the extension parses the DOM sequentially and the site doesn't provide precise timestamps (only 'Today'/'Yesterday' labels), the app displays them in the wrong order.

The Home page (/), however, enforces a strict structural order:

TODAY'S MANGA section
YESTERDAY'S MANGA section
OLDER MANGA section
By parsing the Home page sequentially, we guarantee that all 'Today' items are extracted before 'Yesterday' items, resulting in a correct chronological display in the app.

Additionally, as mentioned, the Home page provides higher resolution thumbnails (175x245 vs 55x85) and more items per request (~60 vs 30), making it a superior source for the initial load.

vetleledaal (approved):

PR #12005, opened 2025-12-06
Mangapark | FIX: Implement server fallback for images using a network interceptor

ItachiUchiha-IU:
Refactors the image loading mechanism to resolve issues where pages would fail to load due to individual image server unavailability.

The previous implementation was brittle, as a failure on a single server (e.g., s02) would result in a broken image for the user.

This change introduces an `imageFallbackInterceptor` to the network client: when an image request fails, the interceptor now catches the failure and automatically retries the request using a list of alternative server domains. This process is transparent to the reader and significantly reduces the occurrence of non-loading pages. Crucially, this also resolves failures in the chapter download feature, which relies on the same image fetching logic.

---

Checklist:


src/all/mangapark/src/eu/kanade/tachiyomi/extension/all/mangapark/MangaPark.kt
```diff
@@ -268,6 +270,55 @@ class MangaPark(
...
     private val latch = CountDownLatch(1)
 
+    private fun imageFallbackInterceptor(chain: Interceptor.Chain): Response {
+        val request = chain.request()
+        var response = chain.proceed(request)
+
+        if (response.isSuccessful) return response
+
+        val urlString = request.url.toString()
+        val serverPattern = Regex("https://s\\d{2}")
```
vetleledaal: Move regex to class to avoid recompilation.

src/all/mangapark/src/eu/kanade/tachiyomi/extension/all/mangapark/MangaPark.kt
```diff
@@ -268,6 +270,55 @@ class MangaPark(
...
+                    // If this server also failed, close and loop to the next one
+                    newResponse.close()
+                } catch (e: Exception) {
+                    // Connection error on this mirror, ignore and loop to next
+                }
+            }
+        }
+
+        // If everything failed, re-run original request to return the standard error
+        return chain.proceed(request)
```
vetleledaal, line 318: Can't you just return `response`? If it doesn't work for closed responses maybe only close responses when you try a new URL.

src/all/mangapark/src/eu/kanade/tachiyomi/extension/all/mangapark/MangaPark.kt
ItachiUchiha-IU: You're absolutely right.
Done!

src/all/mangapark/src/eu/kanade/tachiyomi/extension/all/mangapark/MangaPark.kt
ItachiUchiha-IU: Hi @vetleledaal, thanks for the detailed feedback!

That's a great point about avoiding the extra network call. I built out that version to test it, and while it's possible, I ran into a tricky `OkHttp` behavior: you have to be extremely careful to close the previous response before proceeding with a new request, otherwise it can cause resource leaks or a crash if the app later tries to read a closed response body.

Here’s the most robust version I could come up with that follows the logic of keeping the initial response open:

```
    private fun imageFallbackInterceptor(chain: Interceptor.Chain): Response {
        val request = chain.request()
        var response = chain.proceed(request)

        if (response.isSuccessful) return response

        val urlString = request.url.toString()

        if (SERVER_PATTERN.containsMatchIn(urlString)) {
            val servers = listOf("s01", "s03", "s04", "s00", "s05", "s06", "s07", "s08", "s09", "s10", "s02")

            for (server in servers) {
                val newUrl = urlString.replace(SERVER_PATTERN, "https://$server")
                if (newUrl == urlString) continue

                // Close the PREVIOUS failed response before starting a new one.
                response.close()

                val newRequest = request.newBuilder().url(newUrl).build()

                try {
                    val newResponse = chain
                        .withConnectTimeout(5, java.util.concurrent.TimeUnit.SECONDS)
                        .withReadTimeout(10, java.util.concurrent.TimeUnit.SECONDS)
                        .proceed(newRequest)

                    response = newResponse

                    if (response.isSuccessful) {
                        return response
                    }
                } catch (e: Exception) {
                    // If this fails, 'response' is still pointing to the CLOSED object from the top of the loop.
                    // We just continue to the next server.
                }
            }
        }

        // Safety: If the last loop failed with an Exception, 'response' is closed.
        // We cannot return a closed response. We must re-run the original request
        // to return a valid (but failed) response to the user.
        // We check this by seeing if the response message is accessible or by a simple try-catch.
        return try {
            response.peekBody(1) // Check if alive
            response
        } catch (e: Exception) {
            chain.proceed(request)
        }
    }
```

While this version does work and is more performant in that specific failure scenario, I was concerned about a couple of things:
1. Readability and Maintainability: This approach requires carefully tracking the state of the mutable response variable through the loop, especially in the catch block. The final try/catch with peekBody(1) is a clever but non-obvious trick to check if the response is still open, which can make the code harder for the next developer to maintain.
2. Risk: The logic is more complex, which slightly increases the risk of accidentally introducing a resource leak in the future if the code is modified without fully understanding the state management.

The current implementation in the PR is stateless and linear. It's easy to follow: close the failed response, try the fallbacks, and if nothing works, get a fresh error.

My feeling was that the simplicity and guaranteed resource safety of the current version in the PR outweighs the small performance gain that the more complex version provides in a rare failure case.

What are your thoughts on this trade-off? I'm happy to go with the other approach if you still think it's the better direction.

ItachiUchiha-IU:
Hi @vetleledaal, thanks for the review! I've addressed all the feedback:

- The first point is fixed in the latest commit.

- For the second point, I've left a detailed comment explaining my reasoning for the current approach and would love to hear your thoughts on the trade-offs.

Ready for another look when you have a moment.

src/all/mangapark/src/eu/kanade/tachiyomi/extension/all/mangapark/MangaPark.kt
vetleledaal: If you just don't overwrite `response` it should be easier to handle. You don't really need to return the original request, any of the responses will work. I agree that using `.peekBody` would be a bad approach.

Could something like this work? It's not too dissimilar to your approach. Parts of the code written by GPT-4.1.
```kt
private fun imageFallbackInterceptor(chain: Interceptor.Chain): Response {
    val request = chain.request()

    if (response.isSuccessful) return response

    val urlString = request.url.toString()

    if (SERVER_PATTERN.containsMatchIn(urlString)) {
        listOf("s01", "s03", "s04", "s00", "s05", "s06", "s07", "s08", "s09", "s10", "s02")
            .map { urlString.replace(SERVER_PATTERN, "https://$it") }
            .filter { it != urlString }
            .forEach { newUrl ->
                val newRequest = request.newBuilder().url(newUrl).build()
                try {
                    val newResponse = chain
                        .withConnectTimeout(5, java.util.concurrent.TimeUnit.SECONDS)
                        .withReadTimeout(10, java.util.concurrent.TimeUnit.SECONDS)
                        .proceed(newRequest)

                    if (newResponse.isSuccessful) {
                        response.close()
                        return newResponse
                    }

                    newResponse.close()
                } catch (_: Exception) {}
            }
    }
    return response
}
```

If there are some problems with the longevity of the original response I would go for a different approach, such as making `response` a `var` and skipping `newResponse.close()` on the last iteration. I haven't looked too much into it.


Also, maybe don't rely too much on LLMs -- it's useful for prototyping and highlighting what you might not otherwise think to mention. It also makes it more difficult for others to gauge to what degree you reviewed the output.

Kerint11:
The same problem occurs with bato.to extension but it's server code is k02, k01 etc. And servers k00, k08, k09 don't respond.

src/all/mangapark/src/eu/kanade/tachiyomi/extension/all/mangapark/MangaPark.kt
ItachiUchiha-IU: Hi @vetleledaal,

The reason why I opted for that approach is that, as I discovered before even making the first pull request, `OkHttp` requires the previous response to be closed **before** `chain.proceed()` can be called again for a new request, likely due to how the connection pool is managed. **Through testing multiple different approaches, I confirmed that if we don't do this, the retry logic fails.**

Your `forEach` proposal unfortunately runs into the same underlying issue: the original response isn't closed before the new requests are made. Because of this, I've stuck with the standard `for` loop and a 'close-early' pattern, as it's the safest and most reliable way to guarantee resources are released correctly.

The alternative approach you suggested—making `response` a `var` and reassigning it inside the loop—is exactly what I built and demonstrated in my previous comment (the `peekBody` version). As we saw, the result is code that is significantly more complex to manage, especially if a request fails with an exception, as it risks leaving a closed response object that can't be returned.

For full transparency, I am using an LLM to help generate code, as I'm new to Kotlin (this was my first time even hearing about this language as I had never approached android before) and came here specifically to fix this extension myself. ***This issue is currently impacting a very large number of users, making the MangaPark source unusable due to constant image loading errors.*** My goal has been to adapt a fix I've already implemented for browsers to get the extension working again for everyone.

While I use a tool to help, I **fully** review, test, and fix the code myself. That's how I discovered this specific `OkHttp/response` behavior and developed what I believe is the easiest and safest working solution. Your proposal, in fact, had the same subtle bug as the first code generated by the LLMs, which reinforces that this is a tricky problem.

I absolutely agree that the code can likely be written to a higher standard, and I'm all for refactoring. However, I am bringing a ***hotfix*** that, in my opinion, should be merged as soon as possible to address the widespread user impact. Afterward, we could open a new issue to discuss refining the code further. I believe this is the best path forward to restore functionality for users right now.

Furthermore, this is another refactor of my previous working versions. Tell me what you think about this approach.
```Kotlin
    private fun imageFallbackInterceptor(chain: Interceptor.Chain): Response {
        val request = chain.request()
        val initialResponse = chain.proceed(request)

        if (initialResponse.isSuccessful || !SERVER_PATTERN.containsMatchIn(request.url.toString())) {
            return initialResponse
        }

        // We know the first attempt failed. Close the response body to release the
        // connection from the pool and prevent resource leaks before we start the retry loop.
        // This is critical; otherwise, new requests in the loop may hang or fail.
        initialResponse.close()

        val urlString = request.url.toString()
        val servers = listOf("s01", "s03", "s04", "s00", "s05", "s06", "s07", "s08", "s09", "s10", "s02")

        for (server in servers) {
            val newUrl = urlString.replace(SERVER_PATTERN, "https://$server")
            if (newUrl == request.url.toString()) continue

            val newRequest = request.newBuilder().url(newUrl).build()
            try {
                val newResponse = chain
                    .withConnectTimeout(5, java.util.concurrent.TimeUnit.SECONDS)
                    .withReadTimeout(10, java.util.concurrent.TimeUnit.SECONDS)
                    .proceed(newRequest)

                if (newResponse.isSuccessful) {
                    return newResponse
                }
                // This server also failed, close its response to free up the connection.
                newResponse.close()
            } catch (_: Exception) {
                // This server timed out or had a connection error. The loop will continue.
            }
        }

        // If we get here, all fallbacks failed. Re-run the original request to
        // get a fresh, standard error response to return to the app.
        return chain.proceed(request)
    }
```

ItachiUchiha-IU:
> The same problem occurs with bato.to extension but it's server code is k02, k01 etc. And servers k00, k08, k09 don't respond.

Arg... u're absolutely right!
Regarding this... wasn't it rumored that the same person owns both the sites and they (the owner) are currently unreachable?
I just hope they dont let these two sites die as they are literally at the top of the chain...

Anyway, thank you very much for reporting!
I will post the fix for browsers on reddit anytime soon, then ***maybe*** I will give a look at the extension too.
Tbvh, if this PL were to be merged anytime soon, maybe I would be more encouraged to look at it 😫~

vetleledaal (approved):

Kerint11:
Yeah he is unreacheable :( I Hope that too, but we can only observe for now.

PR #11654, opened 2025-11-14
BookWalker Global: high-res thumbnails and details from api

manti-X:
Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [ ] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [ ] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [ ] Have removed `web_hi_res_512.png` when adding a new extension

PR #12008, opened 2025-12-06
BookWalker Global: various fixes/improvements

TheUnlocked:
Checklist:

There were a couple presumably inadvertent changes introduced by #11654 around how series details were fetched, plus the BookWalker site recently changed the markup around pre-order chapters which broke detection. This PR addresses those issues and adds a new option for those who may want it.

Added:
* Option to (try to) use the first volume cover for the series thumbnail (disabled by default)

Fixed:
* Descriptions are extracted from the first chapter/volume again rather than from the latest one
* Series with different authors for different chapters now show all of the authors again
* "Pre-order" selector adjusted to detect pre-orders properly again after site changes

### DIFF
--- a/src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
+++ b/src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
@@ -1,6 +1,5 @@
 package eu.kanade.tachiyomi.extension.en.bookwalker
 
-import android.app.Application
 import android.text.Editable
 import android.text.TextWatcher
 import android.util.Log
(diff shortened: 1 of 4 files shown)


src/en/bookwalker/src/eu/kanade/tachiyomi/extension/en/bookwalker/BookWalker.kt
```diff
@@ -434,21 +427,42 @@ class BookWalker : ConfigurableSource, ParsedHttpSource(), BookWalkerPreferences
...
-            val firstItem = seriesPage.selectFirst(".o-tile:not(:has(.a-ribbon-pre-order)) .a-tile-ttl a")
-            val uuid = firstItem!!.absUrl("href").substringAfter("/de").substringBefore("/")
-            val bookUpdate = fetchBookUpdate(uuid)
+            val validChapters = seriesPage.select(".o-tile:not(:has($TILE_PREORDER_SELECTOR)) .a-tile-ttl a")
+
+            suspend fun linkElementToBookUpdate(link: Element): BookUpdateDto {
+                return link
+                    .absUrl("href")
+                    .substringAfter("/de")
+                    .substringBefore("/")
```
vetleledaal, line 436: Can you use `OkHttp`'s `HttpUrl` instead? Get the index of segment `de` and use the next segment. Or get the segment at position x, if it's static.

TheUnlocked: This is existing logic just reformatted, but `de` isn't a segment, it's prepended to the UUID segment. The full URL for a chapter would look like `https://global.bookwalker.jp/de00000000-0000-0000-0000-000000000000/`, or just `/de00000000-0000-0000-0000-000000000000/` when reading as a relative URL as happens elsewhere in the code.

vetleledaal (approved):

PR #12012, opened 2025-12-06
ScanR: Add filtering

CriosChan:
Checklist:

### DIFF
--- a/src/fr/scanr/src/eu/kanade/tachiyomi/extension/fr/scanr/ScanR.kt
+++ b/src/fr/scanr/src/eu/kanade/tachiyomi/extension/fr/scanr/ScanR.kt
@@ -28,6 +28,7 @@ class ScanR : HttpSource() {
         return FilterList(
             TypeFilter(),
             StatusFilter(),
+            AdultFilter(),
         )
     }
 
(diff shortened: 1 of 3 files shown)


src/fr/scanr/src/eu/kanade/tachiyomi/extension/fr/scanr/ScanR.kt
```diff
@@ -32,20 +41,25 @@ class ScanR : HttpSource() {
 
     // Search
     override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
-        val url = if (query.isNotBlank()) {
-            "$cdnUrl/index.json#$query"
-        } else {
-            "$cdnUrl/index.json"
+        val url = "$cdnUrl/index.json".toHttpUrl().newBuilder()
+        if (query.isNotBlank()) {
+            url.addQueryParameter("query", query)
```
vetleledaal: Is the query actually used by the remote? If it's only used in `searchMangaParse` you should modify the fragment instead.

CriosChan: Hi, thanks for the review. No, it's not used by the server. In fact, the filters aren't used either. Is there a better way to pass them to `searchMangaParse` ?

vetleledaal: Using the fragment as you do now is sufficient :+1:

vetleledaal (approved):

PR #12039, opened 2025-12-07
Update BatoTo.kt

### LINKED ISSUE #12035: Batoto image server issues
### Source information

Batoto 1.4.56

### Source language

Multi

### Steps to reproduce

Open any series and start reading

### Expected behavior

All pages should load normally

### Actual behavior

Some pages don't load due to image server issues

### Mihon/Tachiyomi version

Komikku 1.13.3

### Android version

11

### Other details

There is a workaround for desktop browsers: https://old.reddit.com/r/Batoto/comments/1pgb4rj/how_to_fix_panels_not_loading/

but for Tachi/Mihon reading it's broken. The extension could be altered to avoid the now broken image server, but I think a more robust solution would be to have a way to switch image servers that's controlled by the extension's logic instead of relying on what the source says. I'm unsure if there's a way to signal globally that "this image server is not responding, don't use it for a while", if there is that'd be a good way to make it so this is not just a quick fix

### Acknowledgements

### 

---

UjCbFwtBayFM:
Rewrites image URLs to avoid broken image servers.

Closes #12035 

Checklist:

### DIFF
--- a/src/all/batoto/build.gradle
+++ b/src/all/batoto/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Bato.to'
     extClass = '.BatoToFactory'
-    extVersionCode = 56
+    extVersionCode = 57
     isNsfw = true
 }
 
--- a/src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
+++ b/src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
@@ -612,7 +612,7 @@ open class BatoTo(
                 it
             }
 
-            Page(i, imageUrl = url)
+            Page(i, imageUrl = url.replace("https://k", "https://n"))
         }
     }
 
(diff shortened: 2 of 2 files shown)


Smol-Ame:
Need to bump versionCode so it shows up as an update

mm12:
Just my opinion, but this should be a fallback that activates on fail, not a hard-coded URL. about 95% of the images do not need this, so doing it for all images would just put strain on the `k` server. 

Additional context Their discord says that their image servers are having issues, but that their dev is away with no known return date so they are unsure when it will be fixed. 

edit: also, we should generalize this a bit more to fallback to another server, instead of just `n` and `k` 
edit2: do we still update the `changelog.md` for individual extensions?

UjCbFwtBayFM:
> Just my opinion, but this should be a fallback that activates on fail, not a hard-coded URL. about 95% of the images do not need this, so doing it for all images would just put strain on the `k` server.

I did not put a fallback because I don't know how to write one but I agree it would be a good thing to rely on `n` servers only when needed as from my own observations only `k03`, `k06` and `k07` URLs seems to be working. 

> we should generalize this a bit more to fallback to another server, instead of just `n` and `k`

Do they even have servers other than `n` and `k`?

mm12:
> I did not put a fallback because I don't know how to write one but I agree it would be a good thing to rely on `n` servers only when needed as from my own observations only `k03`, `k06` and `k07` URLs seems to be working.

I think we can override `imageRequest` to do this better. If I figure it out I'll make a PR on your repo so you can add it to this. 

 
> Do they even have servers other than `n` and `k`?

they have at least `n`,`k`,`b`,`d` I think.

AwkwardPeak7:
write an okhttp interceptor which changes the image subdomain only if request fails

PR #12045, opened 2025-12-07
feat(logger): Added Chunked logger and Detailed HTTP Logger

Starmania:
I written theses loggers to provide detailed insight on what happens and which requests are fired because I don't have a working http proxy setup. 

Compared to a traditional setup, it has the advantage to be really fast to deploy.

The chunked logger has the purpose to log any big chunks of text in logcat while not being over the 4000ish character limit. 

They are not intended to be shipped within an app.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,6 +2,7 @@
 kotlin = "1.7.21"
 coroutines = "1.6.4"
 serialization = "1.4.0"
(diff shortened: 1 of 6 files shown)


AwkwardPeak7:
the app already have http logging when you enable verbose logging. If more detailed request/response logging is needed, it is better to use tools like mitmproxy/mitmweb.

PR #12062, opened 2025-12-08
BatoTo | HOTFIX: Implement server fallback for images using a network interceptor

### LINKED ISSUE #12056: Batoto 503 error
### Source information

Batoto

### Source language

Multi

### Steps to reproduce

open any manga
some random pages will have 503 error

### Expected behavior

no 503 error

### Actual behavior

503 error

### Mihon/Tachiyomi version

r7480

### Android version

Android 14

### Other details

https://github.com/keiyoushi/extensions-source/pull/12005 this fix could be applied to batoto

### Acknowledgements

### 

---

ItachiUchiha-IU:
Closes #12056, #12035.

This PR ports the image fallback interceptor from the MangaPark extension (#12005) to BatoTo. It also incorporates the fix from #12055 and adapts the server-matching regex to be more flexible for BatoTo's URLs. As this implements a more robust solution, it also supersedes the approach taken in PR #12039.

The image loading mechanism has been refactored to resolve issues where pages would fail to load due to individual image server unavailability.

The previous implementation was brittle, as a failure on a single server (e.g., k02) would result in a broken image for the user.

This change introduces an `imageFallbackInterceptor` to the network client: when an image request fails, the interceptor now catches the failure and automatically retries the request using a list of alternative server domains. This process is transparent to the reader and significantly reduces the occurrence of non-loading pages. Crucially, this also resolves failures in the chapter download feature, which relies on the same image fetching logic.

Checklist:


src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
```diff
@@ -1094,6 +1151,7 @@ open class BatoTo(
     }
 
     companion object {
+        private val SERVER_PATTERN = Regex("https://[a-zA-Z]\\d{2}")
```
mm12, line 1154: Is there a reason to use `\\` instead of `\`? This doesn't seem like it would work, matching `https://k\dd` instead of `https://k00`

fix: ItachiUchiha-IU/extensions-source#1

src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
```diff
@@ -633,6 +638,56 @@ open class BatoTo(
...
+        val urlString = request.url.toString()
+
+        // We know the first attempt failed. Close the response body to release the
+        // connection from the pool and prevent resource leaks before we start the retry loop.
+        // This is critical; otherwise, new requests in the loop may hang or fail.
+        response.close()
+
+        if (SERVER_PATTERN.containsMatchIn(urlString)) {
+            // Sorted list: Most reliable servers FIRST
+            val servers = listOf("k03", "k06", "k07", "k00", "k01", "k02", "k04", "k05", "k08", "k09", "n03", "n00", "n01", "n02", "n04", "n05", "n06", "n07", "n08", "n09", "n10")
```
mm12, line 656: What is reliable will likely change over time. In addition, this could strain 'reliable' servers, making them less reliable.

mm12 (commented):
Was going to do something like this, but didn't know how to make an interceptor. Anyway, heres the feedback.

src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
bapeey: In a regular string literal you need to escape the backslash, so `\\d` in the source code becomes `\d` in the actual regex

mm12: For consistency and readability, it should be `"""https://[a-zA-Z]\d{2}"""` then. And, I tested `"https://[a-zA-Z]\d{2}"` and it worked fine.

src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
redeven: This might be too jank but I would add random shuffling to the array of options before iterating, to spread the fallback servers' usage. I would probably also match and remove the failed server from the list of fallbacks.

ItachiUchiha-IU: Im sorry but it's not that simple, some servers have WAY more reachable images and bandwidth than others, and also a faster connection, which also changes based on your traceroute and location. Some (the very first ones in the list), and only them, have ALL the images reachable, at least in the current state of the servers. If we just do the fallback randomly, we may end up having to wait an average of `5 * 10` seconds PER IMAGE before we get to the most reliable servers and finally get the single image loaded. Spreading the fallback servers' usage will only cause significant slowdowns for the user, while those servers have already enough bandwidth to cover all the fallbacks anyway. In fact, if u were to check all the images in all the chapters, u would notice that many of them are loaded from the same very few servers.
In short, this fix needs to be the most effective, even for the worst situations, like the one we are currently in, and the only thing Im certain of is that server `k03` has all the images reachable, at least for now.
In a normal state of the servers, the fallback should never be necessary if not in rare cases, and overengineering it would only make it less effective, given that we have no way of knowing which servers will be down and which will not, we only know which ones are more reliable. For this exact reason, the reliability possibly changing over time simply doesnt matter, it's the same as when the way for fetching the chapters changes, or the naming, or the addresses/name of the servers and even domains. When that happens, we just change them.

ItachiUchiha-IU:
Hi @vetleledaal,
I wanted to bring this PR back to your attention as the user impact is extremely high right now. The BatoTo source is in the same unusable state that MangaPark was, with widespread image loading failures.
This hotfix would resolve the issue for everyone. Please let me know if there's anything I can do to help get this merged.
Thank you for your time and consideration.

slg407:
what about adding the option for the user to set the fallback order themselves? or a helper script button in the extension settings that runs a traceroute or ping on all the servers and sets a fallback order based on connection speed x the server image availability? (availability statistics get updated per version of the extension, the traceroute would only need to be run once when the user installs it, but lets you run it whenever necessary to update the server order)

LetianLi:
I think that would be overcomplicating. This is a hot fix fallback that ideally is never triggered, a workaround for the base site having issues. Ideally the base site gets fixed. 

We can consider that proposal if the image servers become even less reliable and fallbacks get hit more often.

BillyBob1354:
Instead of the fixed list of fallbacks, could it be made to remember the last few servers that worked (like 3-5 recent good ones) and try those first when an image fails?

I tested it manually (I tried the cover art server and the image server before the broken image) and it worked for me (the first broken image on seven different comics).

Note: I manually checked k03 and it did not work for me (all the k servers are down, at least for me)

GGG-KILLER:
Instead of bike shedding, can't this be merged and then a better approach be discussed once no users are being impacted anymore please?

It's been over a week since I've been unable to use this source.

slg407:
> Instead of bike shedding, can't this be merged and then a better approach be discussed once no users are being impacted anymore please?
> 
> It's been over a week since I've been unable to use this source.

i ended up building the extension myself, usually PRs take pretty long to be accepted here, every time an extension breaks i just review the changes myself in PRs to see if they are good and build the PR for my own use, i can confirm the hotfix works perfectly as intended.

Enigmatic-Onion:
> > Instead of bike shedding, can't this be merged and then a better approach be discussed once no users are being impacted anymore please?
> > It's been over a week since I've been unable to use this source.
> 
> i ended up building the extension myself, usually PRs take pretty long to be accepted here, every time an extension breaks i just review the changes myself in PRs to see if they are good and build the PR for my own use, i can confirm the hotfix works perfectly as intended.

could ya host the file somewhere or provide compile instructions? :D

slg407:
> > > Instead of bike shedding, can't this be merged and then a better approach be discussed once no users are being impacted anymore please?
> > > It's been over a week since I've been unable to use this source.
> > 
> > 
> > i ended up building the extension myself, usually PRs take pretty long to be accepted here, every time an extension breaks i just review the changes myself in PRs to see if they are good and build the PR for my own use, i can confirm the hotfix works perfectly as intended.
> 
> could ya host the file somewhere or provide compile instructions? :D

use adoptium jdk 21 LTS for the gradle build, install android studio, download the pull repo, delete all extensions in the repo other than the one you want to build (otherwise it takes AGES for gradle to sync) open the repo on android studio and build

AwkwardPeak7 (approved):

PR #12155, opened 2025-12-13
RaijinScans: Fix "Mihon detection" (Final)

CriosChan:
### Message from one of Raijin's developers found in the source code.
```
/**
 * THE END.
 *
 * * Honestly, after much consideration, this will be the last update I will produce.
 *
 * I have locally tested many methods to break scrapers/extensions:
 * - Cloudflare Invisible Turnstile (Server-side validation)
 * - Full AES-GCM encryption (Instead of XOR)
 * - Serving raw image bytes via keys (.bin files)
 * - Canvas auto-scrambling with daily key updates
 * - WebSockets for image streaming
 * ...and many other creative ways.
 *
 * * Unfortunately, every effective solution came with
 * unacceptable performance costs for legitimate users.
 * * The main trigger for this specific update was a ban from our ad network 
 * due to "fake/bot traffic" caused entirely by similar apps and programs.
 *
 * * I am writing this here because I'm sure you read this code.
 * If you patch this version again: Congratulations, you win. 
 * You can be proud of your achievement ! even though all this time only parsing 
 * was used and one other "hooked atob".
 * * I won't bother with public repositories anymore.
 * 
 * it was fun.
 */
```
### My answer, because I think he'll watch it anyway
```
I'm glad to see how much you care about your users.
And I'm really sorry about your ad network. Maybe we
could find a solution to all this that works for both of us,
but to be honest, apart from an API, I can't see anything
else. And I obviously understand your desire not to be
scraped, but now that my solution is public, it wouldn't
be surprising if some people used it to scrape and re-upload,
unfortunately.
I'm glad I spent this week trying to find and rediscover
solutions to each of your patches. And I'll be honest, I
have immense respect for that.
As you said, it was fun, a really, really fun game.
Just know that it was pure chance that I started looking
for ways to solve all this. If it had been done a week earlier,
I probably wouldn't have touched anything.
I'm glad I got this far.

I might drop by your Discord so we can talk, without
fighting of course. I think we both have a certain respect
for each other and can have a calm discussion.

Open source always wins.
```

As you can see, this should be my last PR for Mihon detection, and I hope that will be the case. This PR fixes Raijin's final patch to make images more difficult to fetch. I hope I have understood their code correctly and won't need to redo a patch anytime soon.

Checklist:

### DIFF
--- a/src/fr/raijinscans/build.gradle
+++ b/src/fr/raijinscans/build.gradle
@@ -2,7 +2,7 @@ ext {
     extName = 'Raijin Scans'
     extClass = '.RaijinScans'
     baseUrl = 'https://raijin-scans.fr'
-    extVersionCode = 53
+    extVersionCode = 54
     isNsfw = false
 }
 
--- a/src/fr/raijinscans/src/eu/kanade/tachiyomi/extension/fr/raijinscans/RaijinScans.kt
+++ b/src/fr/raijinscans/src/eu/kanade/tachiyomi/extension/fr/raijinscans/RaijinScans.kt
@@ -1,6 +1,6 @@
 package eu.kanade.tachiyomi.extension.fr.raijinscans
 
-import android.webkit.CookieManager
+import android.util.Base64
 import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.network.POST
 import eu.kanade.tachiyomi.source.model.FilterList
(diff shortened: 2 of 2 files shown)


Riztard:
noo. plz dont scrape our pirated content :(

btw you can also just add rate limit

AwkwardPeak7 (approved):

PR #12160, opened 2025-12-14
Fix Manhuarm chapter loading and translation issues

SuperMario22922:
## Issue
Chapters fail to load in Mihon with error:
> Unexpected JSON token at offset 0: Expected start of the array '[', but had '‹'
The extension was unable to fetch OCR translation data from the server, causing chapter loading to fail completely.

## Root Cause
1. Server returns gzip-compressed JSON responses that weren't being decompressed
2. Using `cloudflareClient` for JSON requests caused 403 Forbidden responses from Cloudflare

## Changes
- Use `network.client` (plain HTTP) for JSON requests to avoid Cloudflare blocks
- Add gzip decompression for compressed server responses

Checklist:

### DIFF
--- a/src/all/manhuarm/src/eu/kanade/tachiyomi/extension/all/manhuarm/Manhuarm.kt
+++ b/src/all/manhuarm/src/eu/kanade/tachiyomi/extension/all/manhuarm/Manhuarm.kt
@@ -31,6 +31,7 @@ import okhttp3.Headers
 import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.Response
+import okhttp3.brotli.BrotliInterceptor
 import org.jsoup.nodes.Document
 import org.jsoup.nodes.Element
 import java.util.Calendar
@@ -146,6 +147,11 @@ class Manhuarm(
             .connectTimeout(1, TimeUnit.MINUTES)
             .readTimeout(2, TimeUnit.MINUTES)
             .rateLimit(2, 1)
+            // Fix disk cache / decompression issues
+            .apply {
+                val index = networkInterceptors().indexOfFirst { it is BrotliInterceptor }
+                if (index >= 0) interceptors().add(networkInterceptors().removeAt(index))
+            }
             .addInterceptor(warmupInterceptor)
             .addInterceptorIf(
                 !disableTranslator && language.lang != language.origin,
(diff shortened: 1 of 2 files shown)


src/all/manhuarm/src/eu/kanade/tachiyomi/extension/all/manhuarm/Manhuarm.kt
```diff
@@ -189,16 +190,31 @@ class Manhuarm(
...
+            .add("Referer", chapterUrl)
+            .add("Accept", "application/json")
             .build()
 
-        val dialog = client.newCall(GET("$baseUrl/wp-content/uploads/ocr-data/$chapterId.json", headers))
-            .execute()
-            .parseAs<List<PageDto>>()
+        val dialog = try {
+            // Use network.client (without Cloudflare interceptors) for JSON
+            val response = network.client.newCall(GET("$baseUrl/wp-content/uploads/ocr-data/$chapterId.json", jsonHeaders))
```
bapeey: In [Mihon](<https://github.com/mihonapp/mihon/blob/2ecb70a590b89fb2c75e6b7992d463366389e633/core/common/src/main/kotlin/eu/kanade/tachiyomi/network/NetworkHelper.kt#L76>) `client` and `cloudflareClient` are the same. You probably want to remove the `BrotliInterceptor` like [this](<https://github.com/keiyoushi/extensions-source/blob/a22582111c917172e656e35f49ae68ceb4aa290c/src/en/kagane/src/eu/kanade/tachiyomi/extension/en/kagane/Kagane.kt#L78>) and see if that alone fixes the issue

SuperMario22922: applied your suggestion and tested. all working

src/all/manhuarm/src/eu/kanade/tachiyomi/extension/all/manhuarm/Manhuarm.kt
```diff
@@ -315,6 +336,20 @@ class Manhuarm(
     // Prevent bad fragments
     fun String.toFragment(): String = "#${this.replace("#", "*")}"
 
+    /**
+     * Decompresses gzip response if needed by checking for gzip magic number.
+     * Some servers return gzip-compressed data without OkHttp's automatic decompression.
+     */
+    private fun Response.decompressGzipIfNeeded(): String {
```
bapeey: Check if this is still needed, removing `BrotliInterceptor` should disable the compression

SuperMario22922: it was indeed not needed anymore, removed it and tested and it works. thanks

AwkwardPeak7 (approved):

PR #12190, opened 2025-12-15
Manga Up: add preferences and small fixes

manti-X:
Checklist:

### DIFF
--- a/src/all/mangaup/src/eu/kanade/tachiyomi/extension/all/mangaup/MangaUp.kt
+++ b/src/all/mangaup/src/eu/kanade/tachiyomi/extension/all/mangaup/MangaUp.kt
@@ -135,7 +135,8 @@ class MangaUp(override val lang: String) : HttpSource(), ConfigurableSource {
                 val isSecretValid = !newSecret.isNullOrEmpty() && newSecret != failedSecret
 
                 if (!isSecretValid && isMyPage) {
-                    throw IOException("Log in via WebView to access your favorites")
+                    val target = request.url.fragment
+                    throw IOException("Log in via WebView to access your $target")
                 }
 
                 val newUrl = request.url.newBuilder()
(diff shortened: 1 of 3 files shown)


src/all/mangaup/src/eu/kanade/tachiyomi/extension/all/mangaup/MangaUp.kt
```diff
@@ -273,33 +346,80 @@ class MangaUp(override val lang: String) : HttpSource() {
...
+                "Press \"Reset\" in filters or restart the app to apply changes."
+            setDefaultValue(false)
+        }.also(screen::addPreference)
+
+        SwitchPreferenceCompat(screen.context).apply {
+            key = LOGGED_IN_SESSION_PREF
+            title = "Logged-in Session"
+            summary = "If enabled, uses the session from the WebView to access your content.\nDisable to browse as a guest."
+            setDefaultValue(true)
+        }.also(screen::addPreference)
```
AwkwardPeak7: I don't think such preferences are necessary. always show favorites and history in filters, throw if not logged in. and always browse as logged in user if logged in via webview

src/all/mangaup/src/eu/kanade/tachiyomi/extension/all/mangaup/MangaUp.kt
```diff
@@ -74,27 +84,64 @@ class MangaUp(override val lang: String) : HttpSource() {
         return secret
     }
 
+    @SuppressLint("SetJavaScriptEnabled")
+    private fun flushSecret(target: String) {
```
AwkwardPeak7, line 89: `@Synchronized` ?

src/all/mangaup/src/eu/kanade/tachiyomi/extension/all/mangaup/MangaUp.kt
manti-X: >Logged-in Session

When logged in, it’s possible to hide entries on the website therefore the API won’t show those entries. When logged out, the api will show all entires and all entries are visible to the user.

AwkwardPeak7: if user hid some entries in their account, that's on them. or am i misunderstanding?

manti-X: Yes, but this allows easy browsing in two states:
– When the preference is off and logged in: show all entries (like a guest)
– When it’s on and logged in: show all the entries you haven’t hidden on the website
(Logged out users won't see any differences when pref is on/off)
The API reflects the same behavior as the website

AwkwardPeak7: Hmm Alright

AwkwardPeak7 (approved):

PR #12349, opened 2025-12-24
fix(MangaPoisk): query url detection

EnergoStalin:
query may not always be first


src/ru/mangapoisk/src/eu/kanade/tachiyomi/extension/ru/mangapoisk/MangaPoisk.kt
```diff
@@ -89,7 +89,8 @@ class MangaPoisk : ParsedHttpSource() {
 
     override fun searchMangaParse(response: Response): MangasPage {
         val document = response.asJsoup()
-        val mangas = if (document.location().contains("search?q")) {
+        val location = document.location()
+        val mangas = if (location.contains("?q=") || location.contains("&q=")) {
```
vetleledaal: You should instead use `HttpUrl` and check if `.queryParameter("q") != null`.

EnergoStalin: Fair point fixed

vetleledaal (approved):

PR #12320, opened 2025-12-23
feat(batotov4): add Bato.to (v4) extension

Sgwa:
Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [ ] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [ ] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [x] Have removed `web_hi_res_512.png` when adding a new extension

Closes https://github.com/keiyoushi/extensions-source/issues/11781

Issue #12344, opened 2025-12-24
Bato.to: `HTTP 404` error


rrenee449-byte:
### Source information

all.batoto 1.4.57

### Source language

Multiple

### Steps to reproduce

1. Click Browse
2. Click latest in Bato
3. HTTP 404 error appears
4. Click open in we view
5. A white screen saying "hello" appears 

![Image](https://github.com/user-attachments/assets/439b5cbc-1f36-42b3-a319-195642a00161)
![Image](https://github.com/user-attachments/assets/ec6adca0-45f5-492c-942b-f059dc3d9203)
![Image](https://github.com/user-attachments/assets/da9d793e-eec0-4092-86c6-65c8f360793a)

### Expected behavior

When I click the latest button on Bato, all the titles that has been updated on Bato should appear. 

### Actual behavior

However, when I clicked latest, an HTTP 404 error 

![Image](https://github.com/user-attachments/assets/1928fe18-4864-45a5-8ab4-962e794a5d65)

appeared. 

### Mihon/Tachiyomi version

0.19.3

### Android version

Android 13

### Other details

Bato has released a statement explaining their all of their mirrors sites are giving trouble currently. If the issue persists, use bato.si or bato.ing. However, I am not seeing these versions on the Bato extension I have. 

![Image](https://github.com/user-attachments/assets/875881de-6af3-4358-bcee-aafda3a30025)
![Image](https://github.com/user-attachments/assets/1becab02-79a8-4758-b55f-54a3e851cfb5)

![Image](https://github.com/user-attachments/assets/eafd97d0-f16a-4b14-837d-870f867400ea)
![Image](https://github.com/user-attachments/assets/a23f6a85-b47d-454c-9575-70a22b72a32c)
![Image](https://github.com/user-attachments/assets/1cac79f9-349c-49d6-81a6-f2987d103f2b)
![Image](https://github.com/user-attachments/assets/9279545c-9280-4838-90a7-2d51d2459c2d)
![Image](https://github.com/user-attachments/assets/ca9385be-3528-46c6-a7c4-d8c792bc8847)

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to version **[0.15.3](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I have updated all installed extensions.
[...]

PR #12350, opened 2025-12-24
feat(batotov4): add new Bato.to V4 extension

LetianLi:
Adds a new extension for Bato.to V4.

This PR is independent of the one by @Sgwa in https://github.com/keiyoushi/extensions-source/pull/12320

I've tested popular, latest, search, and a lot of the filters. I've also updated the genre and language tags to what's listed on the site.  

Closes #12320, #12344

https://github.com/user-attachments/assets/51e7cc2b-e896-46f6-8cfb-0dcf7036251d

Checklist:


LetianLi:
I plan to move search and obtainment of manga details to the GraphQL api. Currently search is done through `/comics` query parameters (which surprising supports all of the filters) plus scraping, and manga details is done by scraping the title's page.

LetianLi:
This PR started off a copy of the Bato.to extension. 
The GraphQL is inspired by MangaPark as well as the old [BatoV3 rewrite](https://github.com/AwkwardPeak7/tachiyomi-extensions/tree/batov3) by @AwkwardPeak7.

LetianLi:
The only issue at this moment is lack of signed in Utils and History filters. I don't quite understand what those even are.

LetianLi:
But everything else has been moved to GraphQL

LetianLi:
Everything now runs through graphql api (except for getting user ID which is done via scraping).
I have finished porting all the v2 features.

- [x] Search and filtering and sorting
- [x] Logged in History list and Updates list
- [x] Utils list with auto-detect user id (or provide it in preferences)
- [x] Multi language
- [x] Manga details and pagination
- [x] GraphQL

I think the PR is ready for review.

aoba-seragakii:
Hello! I built the extension for testing. In the latest version, the status doesn't get fetched. Also, maybe it would be good adding the scanlator name as well, just like v2.

LetianLi:
Fixed the status parsing and added the scanlator's name.

However, I've mapped the status "pending" to `Smanga.UNKNOWN`, as I'm unsure what exactly that means. Let me know if it should map to ongoing instead.

cuong-tran:
Why made a separated one from the one by Sgwa, care to explain?

LetianLi:
I started working on this December 22nd after getting frustrated that no one else had made an updated extension. 2 days later I had a working extension (mix of scraping and graphql), but realized late that Sgwa had also made a PR (created around the time I started work). So it's a case of unfortunate timing.

Regardless, I created my PR so my work isn't wasted, finished my graphql migration as I was already so close to finishing the migration from scraping to graphql, and installed it onto my phone so I can use it (I trust the code I had already written myself more).

LetianLi:
The API doesn't support the user facing ascending/descending toggles that we currently have inside our SortFilter. 

Should we switch to a SelectFilter instead, or try to messily implement reverse pagination (sending first a request for how many pages are in the search, then doing a search starting from the last page, then reversing the result's order)?

cuong-tran:
> The API doesn't support the user facing ascending/descending toggles that we currently have inside our SortFilter. 
> 
> Should we switch to a SelectFilter instead, or try to messily implement reverse pagination (sending first a request for how many pages are in the search, then doing a search starting from the last page, then reversing the result's order)?

I think remove the order. If you have to track the current max page, it won't support concurrent searching

LetianLi:
Or do two requests per search page, without tracking. But I'd agree that removing is simpler, plus it's a low priority feature.

AwkwardPeak7 (approved):

PR #12453, opened 2025-12-30
BatoTo (v4) | HOTFIX: Reorder server fallbacks

DTM9025:
The current implementation lists the k subdomains first and iterates through the subdomains until it finds a successful image url. The problem is that right now the k subdomains are mostly failing and it takes 5 seconds to timeout, meaning it takes a long time to get a successful image url and thus load times can be over a minute.

We do this quick fix to reorder the subdomains to have the n subdomains first as those are mostly working. Ideally a more sophisticated solution is done, but as this is a problem happening now, we do the simplest, quickest fix to get this out as soon as possible.

Checklist:

### DIFF
--- a/src/all/batoto/build.gradle
+++ b/src/all/batoto/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Bato.to'
     extClass = '.BatoToFactory'
-    extVersionCode = 57
+    extVersionCode = 58
     isNsfw = true
 }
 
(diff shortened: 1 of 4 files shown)


DTM9025:
I'm not entirely sure why the original fix had the k subdomains listed first as currently all the problematic image sub domains start with k (and pretty much all of them are down), but this should hopefully be sufficient for now. Hopefully we can get this out as soon as possible to finally resolve this issue for users.

AwkwardPeak7 (approved):
Please do the same for mangapark extension too

DTM9025:
I don't believe this is needed for mangapark as they exclusively use s subdomains (or at least the fix does), but I can doublecheck.

PR #12480, opened 2026-01-01
Add U-Next

manti-X:
Reading paid chapters is not supported atm. I think there are some issues with signed cookies (or something else) for auth users, I tried but dunno how to resolve this.

There is a hidden wasm (ulibd.wasm) which does the signing (keys, signature). Couldn't figure out how to access/request it, so getting keys via webview.

Reading is geo-restricted to japan.

Checklist:


src/ja/unext/assets/key-extractor.js
vetleledaal, line 1: Use a more descriptive name.

src/ja/unext/assets/script.js
```diff
@@ -0,0 +1,74 @@
+(function() {
+    if (window._keyCheckRunning) return;
+    window._keyCheckRunning = true;
+
+    const interval = setInterval(async function() {
+        try {
+            const el = document.querySelector(".swiper");
+            if (!el) return;
```
vetleledaal: Instead of using a timer to check for DOM changes, can you use a [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)? If you'd rather not you can ignore this suggestion.

src/ja/unext/assets/script.js
```diff
@@ -0,0 +1,74 @@
...
+
+            const getFiber = n => {
+                const k = Object.keys(n).find(x =>
+                    x.startsWith("__reactFiber$") ||
+                    x.startsWith("__reactInternalInstance$")
+                );
+                return n ? n[k] : null;
```
vetleledaal:
```suggestion
                return n?.[k];
```

Changes the signature, but `undefined` is still falsey.

src/ja/unext/assets/script.js
```diff
@@ -0,0 +1,74 @@
...
+            };
+
+            let curr = getFiber(el);
+            let mgr = null;
+            while (curr) {
+                if (curr.memoizedProps?.manager) {
+                    mgr = curr.memoizedProps.manager;
+                    break;
+                }
```
vetleledaal: I think this is doable:
```suggestion
                if (mgr = curr.memoizedProps?.manager) break;
```

src/ja/unext/assets/script.js
```diff
@@ -0,0 +1,74 @@
...
+                curr = curr.return;
+            }
+
+            if (!mgr || !mgr.parser || !mgr.parser.drmContext) return;
+            if (!mgr.parser.drmParser || !mgr.parser.drmParser.drmHeader) return;
```
vetleledaal: Can this be chained?
```suggestion
            if (!mgr?.parser?.drmContext) return;
            if (!mgr?.parser?.drmParser?.drmHeader) return;
```

src/ja/unext/assets/script.js
```diff
@@ -0,0 +1,74 @@
...
+
+            const missingKeyIds = new Set();
+            const keyIdToFilePath = {};
+
+            for (const [path, info] of Object.entries(fileList)) {
+                if (info.keyId && !loadedKeyIds.includes(info.keyId)) {
+                    missingKeyIds.add(info.keyId);
+                    if (!keyIdToFilePath[info.keyId]) {
+                        keyIdToFilePath[info.keyId] = path;
+                    }
```
vetleledaal:
```suggestion
                    keyIdToFilePath[info.keyId]) ??= path;
```

src/ja/unext/assets/script.js
```diff
@@ -0,0 +1,74 @@
...
+                }
+                curr = curr.return;
+            }
+
+            if (!mgr || !mgr.parser || !mgr.parser.drmContext) return;
+            if (!mgr.parser.drmParser || !mgr.parser.drmParser.drmHeader) return;
+
+            const fileList = mgr.parser.drmParser.drmHeader.encryptedFileList;
+            const contextKeys = mgr.parser.drmContext.keys || {};
+            const loadedKeyIds = Object.keys(contextKeys);
```
vetleledaal: Make this a set too, then you can can avoid the $$O(n)$$ of `.includes` and instead use the $$O(1)$$ of `.has`.

src/ja/unext/src/eu/kanade/tachiyomi/extension/ja/unext/UNext.kt
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    companion object {
+        private const val HIDE_PAID_PREF = "HIDE_PAID"
+        private const val POPULAR_QUERY_HASH = "1e1e84fd9b5718c37ef030ea8230bbf9ddd1e5b86f5b8ce2c224b3704f0468ec"
+        private const val LATEST_QUERY_HASH = "0570a586caa9869bd5eb0b05a59bdfec853f92dc6cf280ebd583df2cc93e1c21"
+        private const val SEARCH_QUERY_HASH = "2ec7804350bf993678c92a5d79f20812b3b0d5b38aaba2603c9dd291c6df927e"
+        private const val DETAILS_QUERY_HASH = "99f21ebea20b64b11ef5d3b811c2b3fa5b4dbd8c5d2933baadf9c26fc60b35d1"
+        private const val CHAPTER_LIST_QUERY_HASH = "66f0c600259b82a4826fba7be2ace33726f3ec09735e65a421f8f602b481487d"
+        private const val PLAYLIST_QUERY_HASH = "f8a851c14ec61eb42dff966570b2ad49f86eeec7f39d2d32ab0ec58cad268fc1"
```
vetleledaal, line 367: You should add a comment hinting at where these are sourced from, in case these needs to be updated.

src/ja/unext/src/eu/kanade/tachiyomi/extension/ja/unext/Dto.kt
```diff
@@ -0,0 +1,320 @@
...
+
+    @Serializable
+    class BookFile(
+        val code: String,
+    )
+
+    fun toSChapter(sakuhinCode: String): SChapter = SChapter.create().apply {
+        val lock = if (isFree != true && isPurchased != true) "🔒 " else ""
+        name = "$lock${this@Book.name}"
+        val bfc = bookContent?.mainBookFile?.code
```
vetleledaal: Not sure what a `bfc` is. Is that how the site itself refers to this value?

src/ja/unext/src/eu/kanade/tachiyomi/extension/ja/unext/UNext.kt
```diff
@@ -0,0 +1,422 @@
...
+    override val name = "U-NEXT"
+    private val domain = "unext.jp"
+    override val baseUrl = "https://video.$domain"
+    override val lang = "ja"
+    override val supportsLatest = true
+
+    private val apiUrl = "https://cc.$domain"
+    private val preferences: SharedPreferences by getPreferencesLazy()
+
+    override val client = network.client.newBuilder()
```
vetleledaal:
```suggestion
    override val client = network.cloudflareClient.newBuilder()
```

src/ja/unext/src/eu/kanade/tachiyomi/extension/ja/unext/UNext.kt
```diff
@@ -0,0 +1,422 @@
...
+        .addInterceptor(ImageInterceptor())
+        .addInterceptor { chain ->
+            var request = chain.request()
+            var response = chain.proceed(request)
+
+            // API request for details sometimes gives "PersistedQueryNotFound" but works after 1-2 retries
+            while (request.url.host == "cc.unext.jp" && response.code == 200) {
```
vetleledaal:
```suggestion
            while (request.url.host == "cc.unext.jp" && response.isSuccessful) {
```

src/ja/unext/src/eu/kanade/tachiyomi/extension/ja/unext/UNext.kt
```diff
@@ -0,0 +1,422 @@
...
+                val contentType = response.body.contentType()
+                if (contentType?.subtype == "json") {
+                    val bodyString = response.peekBody(64 * 1024).string()
```
vetleledaal: If you're reading the whole response (i.e. every response is <64KiB) then you should use `Long.MAX_VALUE` instead of a magic value.
```suggestion
                    val bodyString = response.peekBody(Long.MAX_VALUE).string()
```

src/ja/unext/src/eu/kanade/tachiyomi/extension/ja/unext/UNext.kt
```diff
@@ -0,0 +1,422 @@
...
+
+    override fun searchMangaParse(response: Response): MangasPage {
+        val result = response.parseAs<SearchResponse>().data.search
+        val mangas = result.books.map { it.toSManga() }
+        val hasNextPage = result.pageInfo.let { it.page * it.pageSize < it.results }
+        return MangasPage(mangas, hasNextPage)
+    }
+
+    override fun mangaDetailsRequest(manga: SManga): Request {
+        val btd = manga.url.substringAfterLast("/")
```
vetleledaal: Unknown acronym.

vetleledaal: Avoid using string operations directly -- turn into `HttpUrl` then look at `.pathSegments`.
Comment applies for similar patterns in this PR.

src/ja/unext/src/eu/kanade/tachiyomi/extension/ja/unext/UNext.kt
```diff
@@ -0,0 +1,422 @@
...
+
+                val requestData = ImageRequestData(
+                    realZipFile.absolutePath,
+                    entryName,
+                    key,
+                    drmData.iv,
+                )
+
+                val fragment = requestData.toJsonString()
+                Page(i, "http://127.0.0.1/#$fragment")
```
vetleledaal: No image URL?

manti-X: Images are in a zip file (.ubook)

manti-X: Images are in a zip file (.ubook)

vetleledaal: I think it would still be more correct to set `imageUrl` instead of `url`, Normally `imageUrl` is a reference to the resource itself and `url` is a reference to additional context (URL the resource was seen on).

src/ja/unext/src/eu/kanade/tachiyomi/extension/ja/unext/Dto.kt
manti-X: bookFileCode

src/ja/unext/src/eu/kanade/tachiyomi/extension/ja/unext/UNext.kt
```diff
@@ -0,0 +1,441 @@
...
+                val contentType = response.body.contentType()
+                if (contentType?.subtype == "json") {
+                    val bodyString = response.peekBody(Long.MAX_VALUE).string()
+                    if (bodyString.contains("PersistedQueryNotFound")) {
+                        response.close()
+                        request = request.newBuilder().build()
+                        response = chain.proceed(request)
+                        continue
+                    }
+                }
```
AwkwardPeak7: add a limit, so as to not end up in an infinite loop

src/ja/unext/src/eu/kanade/tachiyomi/extension/ja/unext/UNext.kt
```diff
@@ -0,0 +1,441 @@
...
+            val zipFile = File(cacheDir, "unext_${playlistParse.playToken}.ubook")
+
+            val zipRequest = GET(zipUrl, headers)
+            client.newCall(zipRequest).execute().use { zipResponse ->
+                zipResponse.body.byteStream().use { input ->
+                    zipFile.outputStream().use { output ->
+                        input.copyTo(output)
+                    }
+                }
+            }
```
AwkwardPeak7: how big can these usually be? since it is downloading all of it in one go. would it be possible to stream it? like what pandachaika extension does.

manti-X: chapters are around ~3-7 MB and whole volume books vary between ~60-100 MB

vetleledaal (approved):

PR #12470, opened 2025-12-31
PhiliaScans: Handle thumbnail lazy load + attempt to get full cover from thumbnail url

FlaminSarge:
Philia thumbnails are now lazy-loaded, so swapping to imageFromElement. Also, their thumbnail URLs seem to all be the same as the main cover URL but with `-280x400` appended, so this also handles converting to full cover urls in popularMangaParse().

Nobody had opened an issue for thumbnails not loading, so no linked issue.

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [ ] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [ ] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [ ] Have removed `web_hi_res_512.png` when adding a new extension

PR #12492, opened 2026-01-02
Madara: Add method processThumbnail() to make changes to thumbnail_url, implement for PhiliaScans

FlaminSarge:
PhiliaScans: Implement processThumbnail() to remove -280x400 suffix on search thumbnail URLs.

Only including Philia for now but several other Madara sites have these suffixes and would benefit from this being available, hence why this PR isn't just an override of imageFromElement/etc in Philia.

Checklist:

### DIFF
--- a/lib-multisrc/madara/build.gradle.kts
+++ b/lib-multisrc/madara/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 46
+baseVersionCode = 45
 
 dependencies {
     api(project(":lib:cryptoaes"))
--- a/lib-multisrc/madara/build.gradle.kts
+++ b/lib-multisrc/madara/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 45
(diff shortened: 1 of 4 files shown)


FlaminSarge:
This is the followup from #12470 
Tested on emulator, the initial covers loaded during latest/popular search are the full resolution images rather than the smaller versions.

There's _probably_ a regex that can handle this more generally for more Madara exts but some of them do some weird stuff like using 'sample-optimized' for the suffix for the full-resolution covers instead of just removing the -resolution suffix.

lib-multisrc/madara/build.gradle.kts
```diff
@@ -2,7 +2,7 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 45
+baseVersionCode = 46
```
AwkwardPeak7: this isn't changing the behaviour of madara, so don't bump it.

FlaminSarge:
Considering inverting fromSearch param to fromMangaDetails, or not even checking the boolean param for Philia.
I included the param in order to give better flexibility for anyone overriding this function.

AwkwardPeak7 (approved):

PR #12547, opened 2026-01-04
AsuraScans: Add cached auth check and add lock emoji to premium chapters

### LINKED ISSUE #10691: [AsuraScans] Add lock emoji 🔒besides the ch name in the locked Chapters
### Source name

AsuraScans

### Source language

English

if i choose to show the premium chapters / or Locked Chapters, it will be helpful & nice to differentiate between the free Chapters and the locked chapters 🔒 with this.

The Current behavior:
If i don't hide the premium Chapters,
It’s displayed similar to the free Chapters, so this is a problem as the locked chapters

We needs 🔒 to differentiate between them, if we choose to not hide them to see just the latest releases, i know without read/download until the 🔒 is removed when the time finished/be opened.

Will remove the hassle & stress of needing to search the webview everytime to see the lock 🔒removed or not. Or here the time is finished or not.

### Other details

You already implemented something like that, with Qi Scans

Image

### Acknowledgements

### 

---

colblitz:
Changes:
- Add `hasPremiumAccess()` to check premium status, cache results (building from #11339)
- Add 🔒 emoji to premium chapter names when applicable (closes #10691)
- Use `hasPremiumAccess()` in unlock flow for early check

After logging in the chapter titles only update after a little bit (probably the cached auth check expiring and/or having to refresh the chapter list), but I figured that was fine to keep the flow simple and not have to try to detect auth state changes or hook into webview events

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions (updates to **53 instead of 52** because of https://github.com/keiyoushi/extensions-source/pull/12534, could also just combine into one PR if needed)

Testing: Verified with Mihon on Android (non-logged-in user sees lock emojis, can still access non premium chapters, after logging in can access all chapters)

### DIFF
--- a/src/en/asurascans/src/eu/kanade/tachiyomi/extension/en/asurascans/AsuraScans.kt
+++ b/src/en/asurascans/src/eu/kanade/tachiyomi/extension/en/asurascans/AsuraScans.kt
@@ -16,6 +16,8 @@ import keiyoushi.utils.getPreferences
 import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
 import kotlinx.serialization.json.put
 import okhttp3.HttpUrl.Companion.toHttpUrl
 import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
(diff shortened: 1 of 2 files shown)


src/en/asurascans/src/eu/kanade/tachiyomi/extension/en/asurascans/AsuraScans.kt
```diff
@@ -399,7 +413,7 @@ class AsuraScans : ParsedHttpSource(), ConfigurableSource {
             throw Exception(errorMsg)
         }
 
-        val responseBody = body?.string() ?: throw Exception("$errorPrefix: Empty response")
+        val responseBody = body.string()
```
colblitz, line 416: this was because of compiler warning

```
w: .../AsuraScans.kt: (416, 32): Unnecessary safe call on a non-null receiver of type ResponseBody
```

vetleledaal (approved):

Issue #5195, opened 2024-09-24
Pixiv users url changed


StarTwi:
### Source information

Pixiv 1.4.9

### Source language

all

### Source new URL

https://www.pixiv.net/search/users

### Other details

Pixiv users url changed, the users filter doesnt work anymore

### Acknowledgements

- [X] I have updated all installed extensions.
- [X] I have opened WebView and checked that the source URL has not changed yet.
- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

StarTwi:
I dont know where this issue goes, this was what I thought was best please go easy on me

AwkwardPeak7:
this should just be bug report

FengTeLai:
Still hoping someone can take over this issue, there are quite a few renamed works on Pixiv and need to rely on Users to sift through them.
#1968

ZenVinny:
Resolved in [#12615](https://github.com/keiyoushi/extensions-source/pull/12615)

PR #12615, opened 2026-01-08
[pixiv] Fix user search + filters, add image quality setting

### LINKED ISSUE #481: Pixiv: Add option to load preview image (smaller size image)
### Source name

Pixiv

### Source language

All

The current implementation loads full size image, which some individual image can be multiple of megabytes.

By adding option to load preview image instead, it can reduce the loading times.

### Other details

_No response_

### Acknowledgements

### LINKED ISSUE #1968: pixiv artist search based on user ID
### Source name

Pixiv

### Source language

English/Multi

current artist search is non-fuctional, please correct me if i don't quite catch how the feature works. but isn't it supposed to search for artworks caption then narrow the results down to match artist/user name? if so, i'm still unable to search for specific artist works.
wouldn't it be nice also if we could search for artist from the filter menu without a search query and it would return their works, currently doing exactly this would yield no results.

### Other details

Screenshot_2024-03-19-04-53-11-295_eu kanade tachiyomi sy
Screenshot_2024-03-19-04-53-01-351_eu kanade tachiyomi sy

### Acknowledgements

- [ ] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/mihonapp/mihon/issues/new/choose).
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements.

colblitz:
Changes:

 * Change username search to use /search/users endpoint (closes #1968, #5195)
 * Add setting for image quality and get urls based on setting (closes #481)
 * Add back deeplink support
 * Add some smarts when getting filtered results (and avoid searches that spin forever)
 * Make user filter case insensitive
 * Rename some variable shadowing, minor formatting things

Checklist:

Testing:

Tested on Android with Mihon

 * verified popular/latest still work
 * verified deeplinks work (for user and aid)
 * verified user filter works
 * verified user filter + search works

### DIFF
--- a/src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
+++ b/src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
@@ -1,12 +1,16 @@
 package eu.kanade.tachiyomi.extension.all.pixiv
 
+import android.app.Application
+import android.content.SharedPreferences
+import androidx.preference.ListPreference
+import androidx.preference.PreferenceScreen
+import eu.kanade.tachiyomi.source.ConfigurableSource
 import eu.kanade.tachiyomi.source.model.FilterList
 import eu.kanade.tachiyomi.source.model.MangasPage
 import eu.kanade.tachiyomi.source.model.Page
 import eu.kanade.tachiyomi.source.model.SChapter
 import eu.kanade.tachiyomi.source.model.SManga
 import eu.kanade.tachiyomi.source.online.HttpSource
-import eu.kanade.tachiyomi.util.asJsoup
 import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.decodeFromJsonElement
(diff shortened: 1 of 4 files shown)


src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -16,15 +20,21 @@ import okhttp3.HttpUrl.Companion.toHttpUrl
...
+class Pixiv(override val lang: String) : ConfigurableSource, HttpSource() {
     override val name = "Pixiv"
     override val baseUrl = "https://www.pixiv.net"
     override val supportsLatest = true
 
     private val json: Json by injectLazy()
 
+    private val preferences: SharedPreferences by lazy {
+        Injekt.get<Application>().getSharedPreferences("source_$id", 0x0000)
+    }
```
AwkwardPeak7:
```suggestion
    private val preferences: SharedPreferences by getPreferencesLazy()
```
need to import

AwkwardPeak7 (approved):

PR #12727, opened 2026-01-13
Mangabox: Update API Fetching for Chapters

JakeeLuwi:
Closes #12729

Decided to leave out `chapterFromElement` incase they decided to revert it back from the old way

TODO: 
- [x] Update chapter requests
- [x] Refactor / cleanup
- [x] Include issues related to Mangabox

Checklist:

- [ ] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [x] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [ ] Have removed `web_hi_res_512.png` when adding a new extension

PR #12785, opened 2026-01-16
fix: fix_chapter_load_manganelo

Relcai:
Checklist:


JakeeLuwi:
Already fixed in #12727

Also, 
- This will only solve manganato/manganelo, other similar extensions will be left unfixed.
- You only fetched first 500 of the list. In chapters with more than 500 such as `Martial Peak`, only the last 500 chapters will be fetched.

PR #12942, opened 2026-01-22
XBatCat: small improvments

AwkwardPeak7:
- remove unused dependency
- add legacy mirrors for deeplink
  - they have bunch of old urls in various manga descriptions, so this helps with clicking on them to open in extension 
- add full title in description
- show refresh message on old chapter urls

Checklist:

### DIFF
--- a/src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/XBatCat.kt
+++ b/src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/XBatCat.kt
@@ -91,16 +91,13 @@ class XBatCat(
 
     override fun fetchSearchManga(page: Int, query: String, filters: FilterList): Observable<MangasPage> {
         // Handle URL and ID search
-        val idMatch = when {
-            query.startsWith("https://") -> {
-                val id = urlIdRegex.find(query)?.groupValues?.get(1)
-                    ?: return Observable.error(Exception("Unknown url"))
-                idQueryRegex.matchEntire("id:$id")
-            }
-            else -> idQueryRegex.matchEntire(query.trim())
-        }
-        if (idMatch != null) {
-            val id = idMatch.groupValues[1]
+        if (query.startsWith("https://") || query.startsWith("id:")) {
+            val id = when {
+                query.startsWith("https://") -> {
+                    urlIdRegex.find(query)?.groupValues?.get(1)
+                }
+                else -> idQueryRegex.matchEntire(query.trim())?.groupValues?.get(1)
+            } ?: return Observable.error(Exception("Unknown url or id"))
 
             if (id.toIntOrNull() == null) {
                 return Observable.error(Exception("comic id must be integer"))
@@ -377,6 +374,10 @@ class XBatCat(
 
     // ************ Page List ************ //
     override fun pageListRequest(chapter: SChapter): Request {
+        if (chapter.url.toIntOrNull() == null) {
+            throw Exception("Refresh Chapter List")
+        }
+
         val apiVariables = ApiChapterNodeVariables(id = chapter.url)
 
         return graphQLRequest(apiVariables, CHAPTER_NODE_QUERY)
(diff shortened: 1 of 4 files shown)


src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/XBatCat.kt
```diff
@@ -91,16 +91,13 @@ class XBatCat(
...
-        }
-        if (idMatch != null) {
-            val id = idMatch.groupValues[1]
+        if (query.startsWith("https://") || query.startsWith("id:")) {
+            val id = when {
+                query.startsWith("https://") -> {
+                    urlIdRegex.find(query)?.groupValues?.get(1)
+                }
+                else -> idQueryRegex.matchEntire(query.trim())?.groupValues?.get(1)
+            } ?: return Observable.error(Exception("Unknown url or id"))
```
vetleledaal, line 100: Probably just shuffling stuff around, but weird that you're checking for `query.startsWith("https://")` twice.
```suggestion
        val id = when {
            query.startsWith("https://") -> urlIdRegex.find(query)?.groupValues?.get(1)
            query.startsWith("id:")) -> idQueryRegex.matchEntire(query.trim())?.groupValues?.get(1)
            else -> null
        } ?: return Observable.error(Exception("Unknown url or id"))
```

src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/XBatCat.kt
```diff
@@ -377,6 +374,10 @@ class XBatCat(
 
     // ************ Page List ************ //
     override fun pageListRequest(chapter: SChapter): Request {
+        if (chapter.url.toIntOrNull() == null) {
+            throw Exception("Refresh Chapter List")
+        }
```
vetleledaal, line 379: You can abuse the elvis operator a bit here:
```suggestion
        chapter.url.toIntOrNull() ?: Exception("Refresh Chapter List")
```

vetleledaal (commented):
Should probably wait until maintenance is over before merging.

src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/XBatCat.kt
```diff
@@ -91,16 +91,13 @@ class XBatCat(
...
-            query.startsWith("https://") -> {
-                val id = urlIdRegex.find(query)?.groupValues?.get(1)
-                    ?: return Observable.error(Exception("Unknown url"))
-                idQueryRegex.matchEntire("id:$id")
-            }
-            else -> idQueryRegex.matchEntire(query.trim())
-        }
-        if (idMatch != null) {
-            val id = idMatch.groupValues[1]
+        if (query.startsWith("https://") || query.startsWith("id:")) {
```
LetianLi, line 94: We should make the “id” check case insensitive. `str.startsWith(“id:”, ignoreCase = true)` should work, the regex already handles case insensitivity.

src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/XBatCat.kt
```diff
@@ -538,11 +539,6 @@ class XBatCat(
             entryValues = Array(mirrors.size) { it.toString() }
             summary = "%s"
             setDefaultValue("0")
-
-            setOnPreferenceChangeListener { _, _ ->
-                Toast.makeText(screen.context, "Restart the app to apply changes", Toast.LENGTH_LONG).show()
```
LetianLi, line 543: I don’t think we should delete this. A restart is necessary to update the `baseUrl`.

LetianLi: Nevermind, should be fine considering baseUrl.get()

PR #12995, opened 2026-01-25
DocTruyen3Q: Fix images load

FiorenMas:
Checklist:

### DIFF
--- a/src/vi/doctruyen3q/build.gradle
+++ b/src/vi/doctruyen3q/build.gradle
@@ -3,7 +3,7 @@ ext {
     extClass = '.DocTruyen3Q'
     themePkg = 'wpcomics'
     baseUrl = 'https://doctruyen3qhubz.com'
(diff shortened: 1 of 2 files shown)


dejavui:
I think using it this way is still better. ( ͡° ͜ʖ ͡°) 
```Kotlin
        return document.select(".page-chapter[id] img").mapIndexed { index, element ->
            val rawUrl = when { element.hasAttr("data-src") -> element.attr("abs:data-src") else -> element.attr("abs:src") }
            val fixedUrl = when { rawUrl.startsWith("//") -> "https:$rawUrl" else -> rawUrl }
            Page(index, imageUrl = fixedUrl)
        }.distinctBy { it.imageUrl }
    }
```

AwkwardPeak7 (approved):

AwkwardPeak7:
`abs:src` should handle the // case

PR #13197, opened 2026-02-05
Run spotless

### LINKED ISSUE #13327: Sparse checkouts broken by spotless lint
### Describe why this should be added

PR #13170 added spotless to linting, which I believe broke builds in sparse checkouts. Specifically, the ratcheting seems to be causing the issue. This means contributors need to clone the entire repo which slows development.

If spotless ratcheting is to be used only in CI, it might be better to gate the `ratchetFrom` inside [keiyoshi.lint.gradle.kts](https://github.com/keiyoushi/extensions-source/blob/main/buildSrc/src/main/kotlin/keiyoushi.lint.gradle.kts) with 
```kotlin
if (System.getenv("CI") == "true") {
    ratchetFrom = "68ed874a42038f4efd001c555cba9bf7de8474d7"
}
```

### Other details

I'm getting the following error for the tasks spotlessGradle, spotlessKotlin, spotlessXml whenever I try to build an extension.
```
> org.eclipse.jgit.errors.CorruptObjectException: Invalid path: .github/
```

### Acknowledgements

### 

---

AwkwardPeak7:
first commit is entirely from `spotlessApply`
second is manual changes which spotless complained about

This commit can be added to `.git-blame-ignore-revs`

closes #13356
closes #13327


FourTOne5:
Maybe look at some more rules to disable from https://github.com/mihonapp/mihon/blob/main/.editorconfig

PR #13243, opened 2026-02-07
Fix chapter number handling in MediocreToonsDto

rod-moraes:
Fix chapter number when number is Float. Example: 0.5 or 1.5

Checklist:


src/pt/mediocretoons/src/eu/kanade/tachiyomi/extension/pt/mediocretoons/MediocreToonsDto.kt
```diff
@@ -142,9 +142,9 @@ fun MediocreChapterSimpleDto.toSChapter(): SChapter {
 
 fun MediocreChapterDetailDto.toPageList(): List<Page> {
     val obraId = manga?.id ?: 0
-    val capituloNome = number?.toInt()?.toString() ?: name
+    val chapterNumber = number?.toString() ?: name
```
vetleledaal: This will turn into chapter numbers such as `1.0`, `2.0`, etc. Normally you would just remove that suffix.
```suggestion
    val chapterNumber = number?.toString()?.removeSuffix(".0") ?: name
```

AwkwardPeak7 (approved):

PR #13371, opened 2026-02-14
CI: resolve dependencies directly in python

AwkwardPeak7:
previous method of calling gradle is slow, as it first restores gradle cache and then runs it. It also requires we disable gradle's `configureondemand`. 
Resolving dependencies in the script via regex skips gradle and allows us to enable `configureondemand` globally, which is especially helpful for local builds

### DIFF
--- a/.github/scripts/generate-build-matrices.py
+++ b/.github/scripts/generate-build-matrices.py
@@ -23,6 +23,43 @@ def run_command(command: str) -> str:
     return result.stdout.strip()
 
 
+def resolve_dependent_libs(libs: set[str]) -> set[str]:
+    """
+    returns all libs which depend on any of the passed libs (/lib),
+    recursively resolving transitive dependencies
+    """
+    if not libs:
+        return set()
+    
+    all_dependent_libs = set()
+    to_process = set(libs)
+    
+    while to_process:
+        current_libs = to_process
+        to_process = set()
+        
+        lib_dependency = re.compile(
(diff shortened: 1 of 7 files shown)


.github/scripts/generate-build-matrices.py
```diff
@@ -22,6 +22,99 @@ def run_command(command: str) -> str:
...
+        for lib in Path("lib").iterdir():
+            if lib.name in all_dependent_libs or lib.name in libs:
+                continue
+                
+            build_file = lib / "build.gradle.kts"
+            if not build_file.is_file():
+                continue
+            
+            with open(build_file) as f:
+                content = f.read()
```
vetleledaal: You can use `.read_text()` to avoid using the context manager.
You should also explicitly set the encoding.

Suggestion applies throughout this PR.
```suggestion
            content = build_file.read_text("utf-8")
```

.github/scripts/generate-build-matrices.py
```diff
@@ -22,6 +22,99 @@ def run_command(command: str) -> str:
...
+    passed multisrcs or libs
+    """
+    
+    multisrc_dependency = re.compile(
+        rf"themePkg\s*=\s*['\"]({'|'.join(map(re.escape, multisrcs))})['\"]"
+    )
+    
+    lib_dependency = re.compile(
+        rf"project\([\"']:(?:lib):({'|'.join(map(re.escape, libs))})[\"']\)"
+    )
```
vetleledaal: Slight optimization: combine these. Might make the code less readable so leaving as-is is fine too.

vetleledaal (dismissed):

PR #13431, opened 2026-02-17
MangaGo: fix missing image urls

### LINKED ISSUE #7138: Mangago only loads a few pages
### Source information

Mangago 1.4.24

### Source language

English

### Steps to reproduce

1. Open Eleceed with the mangago extension
2. Wait for pages to load
3. Any page after 5 does not work even though I can see it through webview

### Expected behavior

All pages should load

### Actual behavior

No pages load after page 5

### Mihon/Tachiyomi version

0.17.1

### Android version

Android 13

### Other details

_No response_

### Acknowledgements

### 

---

valonsodev:
Checklist:

---

Fixes https://github.com/keiyoushi/extensions-source/issues/7138 

---

Summary:
- Mangago returns a decrypted imgsrcs list from chapter HTML
- Normal manga return image urls for every page.
- In webcomics the imgsrcs only contains urls for 5 images (eg: getting page 3 returns images from 3 to 7), while the other array elements are empty. Thus to get all image urls we need to fetch HTML `pages / 5` times.
- Something in the URL format changed from the previous version
- It's better to use the page selection dropdown for links since it was already used to check how many pages a chapter has.

---

There is some special `synchronized` logic to make the reader experience better since without it the chapter would query all pages from the beginning and take a while to start reading or query unneeded HTML


valonsodev:
Some mangas use a different URL without a URL part for the first page like `https://.../chapter/79860/2183173/` and were failing, now fixed

src/en/mangago/src/eu/kanade/tachiyomi/extension/en/mangago/Mangago.kt
```diff
@@ -29,18 +29,52 @@ import okhttp3.HttpUrl.Companion.toHttpUrl
...
+        val iv: ByteArray,
+        val time: Long,
+    )
+
+    private val chapterStates = mutableMapOf<String, ChapterState>()
+    private val pageContexts = WeakHashMap<Page, PageContext>()
+    private val chunkStateLock = Any()
+    private val chapterJsCache = mutableMapOf<String, ChapterJsDecryptCache>()
+    private val chapterJsCacheLocks = mutableMapOf<String, Any>()
+    private val chapterJsCacheLock = Any()
```
AwkwardPeak7: this kind of state can be removed if user restarts the app, while the page list is still cached. 
I would suggest putting all the necessary information in the page/image url as json encoded fragment, this way you have the relivant/necessary information in the url and can access it in `fetchImageUrl`

valonsodev: I took a slightly different approach with this. Basically, I removed the URL parsing since it was a bit flaky and instead opted to use the page URLs from the dropdown that were already being used, passing them as `Page.url`

Now, in `fetchImageUrl`, if the cache already has the `imageUrl` (from being part of a previously requested chunk), it's returned immediately. If not, Page.url is requested and the cache is filled for that chunk of images. Since it caches all non-empty URLs returned, the chunk size is no longer hardcoded.

If the app is restarted it doesn't matter since we can continue from whatever `Page.url` we are currently on and request the HTML and JS to decrypt again.

The cache is there to not request the HTML many times over when one chunk already contains URLs to `CHUNK_SIZE` images

AwkwardPeak7 (approved):

PR #13531, opened 2026-02-22
Fix Les Poroiniens not working

Nyantad:
Checklist:

### DIFF
--- a/src/fr/lesporoiniens/src/eu/kanade/tachiyomi/extension/fr/lesporoiniens/LesPoroiniens.kt
+++ b/src/fr/lesporoiniens/src/eu/kanade/tachiyomi/extension/fr/lesporoiniens/LesPoroiniens.kt
@@ -1,12 +1,123 @@
 package eu.kanade.tachiyomi.extension.fr.lesporoiniens
 
+import eu.kanade.tachiyomi.multisrc.scanr.PageData
 import eu.kanade.tachiyomi.multisrc.scanr.ScanR
+import eu.kanade.tachiyomi.network.GET
+import eu.kanade.tachiyomi.source.model.MangasPage
+import eu.kanade.tachiyomi.source.model.Page
+import eu.kanade.tachiyomi.util.asJsoup
+import keiyoushi.utils.jsonInstance
+import keiyoushi.utils.parseAs
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
+import okhttp3.MediaType.Companion.toMediaType
+import okhttp3.OkHttpClient
+import okhttp3.Response
+import okhttp3.ResponseBody.Companion.toResponseBody
 
 class LesPoroiniens :
     ScanR(
         name = "Les Poroiniens",
         baseUrl = "https://lesporoiniens.org",
         lang = "fr",
-        useHighLowQualityCover = false,
-        slugSeparator = "_",
-    )
+    ) {
+
+    override val client: OkHttpClient = super.client.newBuilder()
+        .addInterceptor { chain ->
+            val response = chain.proceed(chain.request())
+
+            val encodedPath = response.request.url.encodedPath
+            val isSeriesJson = encodedPath.startsWith("/data/series/") && (encodedPath.endsWith(".json") || response.request.url.fragment?.endsWith(".json") == true || encodedPath == "/data/series/")
+
+            if ((response.code == 403 || response.code == 404) && isSeriesJson) {
+                val bodyString = response.peekBody(1024 * 1024).string()
+
+                if (bodyString.contains("Accès refusé", true) || bodyString.contains("Erreur 404", true)) {
+                    response.close()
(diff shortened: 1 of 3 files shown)


src/fr/lesporoiniens/src/eu/kanade/tachiyomi/extension/fr/lesporoiniens/LesPoroiniens.kt
```diff
@@ -1,12 +1,123 @@
...
+    }
+
+    override fun pageListParse(response: Response): List<Page> {
+        val document = response.asJsoup()
+        val chapterNumber = document.location().substringAfterLast("/")
+        val jsonData = document.selectFirst("#reader-data-placeholder")!!.html()
+
+        val jsonElement = jsonInstance.parseToJsonElement(jsonData).jsonObject
+        val series = jsonElement["series"]!!.jsonObject
+        val chaptersElement = series["chapters"]!!
```
vetleledaal: Use `kotlinx-serialization` for this, this is a relatively common pattern used in other extensions.

src/fr/lesporoiniens/src/eu/kanade/tachiyomi/extension/fr/lesporoiniens/LesPoroiniens.kt
```diff
@@ -1,12 +1,123 @@
...
+        .build()
+
+    override fun searchMangaParse(response: Response): MangasPage {
+        val page = super.searchMangaParse(response)
+        return MangasPage(page.mangas.filter { it.title != "DUMMY_ERROR_403" }, page.hasNextPage)
+    }
+
+    override fun pageListParse(response: Response): List<Page> {
+        val document = response.asJsoup()
+        val chapterNumber = document.location().substringAfterLast("/")
```
vetleledaal: Consider using `HttpUrl` and `.pathSegments`. You can use a fixed offset, to avoid issues with links such as `https://example.com/manga/123/`

src/fr/lesporoiniens/src/eu/kanade/tachiyomi/extension/fr/lesporoiniens/LesPoroiniens.kt
```diff
@@ -1,12 +1,123 @@
...
+                    val dummyJson = """{"title":"DUMMY_ERROR_403","description":null,"artist":null,"author":null,"cover":null,"cover_low":null,"cover_hq":null,"tags":null,"release_status":null,"alternative_titles":null,"chapters":null}"""
+                    return@addInterceptor response.newBuilder()
+                        .code(200)
+                        .message("OK")
+                        .body(dummyJson.toResponseBody("application/json".toMediaType()))
+                        .build()
+                }
+            }
+
+            if (response.code == 200 && isSeriesJson) {
```
vetleledaal: Check if response is successful instead of specifically `200`.

Nyantad:
Ok it should be good now if there's nothing else

src/fr/lesporoiniens/src/eu/kanade/tachiyomi/extension/fr/lesporoiniens/LesPoroiniens.kt
```diff
@@ -1,12 +1,112 @@
...
+                if (bodyString.contains("Accès refusé", true) || bodyString.contains("Erreur 404", true)) {
+                    response.close()
+                    val dummyJson = """{"title":"DUMMY_ERROR_403","description":null,"artist":null,"author":null,"cover":null,"cover_low":null,"cover_hq":null,"tags":null,"release_status":null,"alternative_titles":null,"chapters":null}"""
+                    return@addInterceptor response.newBuilder()
+                        .code(200)
+                        .message("OK")
+                        .body(dummyJson.toResponseBody("application/json".toMediaType()))
+                        .build()
+                }
+            }
```
AwkwardPeak7, line 42: why?

src/fr/lesporoiniens/src/eu/kanade/tachiyomi/extension/fr/lesporoiniens/LesPoroiniens.kt
```diff
@@ -1,12 +1,112 @@
...
+                            return@addInterceptor response.newBuilder()
+                                .body(newRoot.toString().toResponseBody(response.body.contentType()))
+                                .build()
+                        }
+                    }
+                } catch (e: Exception) {
+                    // Ignore parse errors, fallback to original body
+                }
+                return@addInterceptor response.newBuilder()
+                    .body(bodyString.toResponseBody(response.body.contentType()))
```
AwkwardPeak7: do this in the appropriate chaperlistparse method. (you can rebuild response and call super method)

src/fr/lesporoiniens/src/eu/kanade/tachiyomi/extension/fr/lesporoiniens/LesPoroiniens.kt
Nyantad: It's because the site sometimes returns 403 for some series which isn't Cloudflare-related. Without checking this the app tries to use FlareSolverr for nothing. I've simplified it to just return a 404 so those broken series are skipped properly.

AwkwardPeak7 (approved):

PR #13760, opened 2026-03-03
Refactor RaijinScans into multisrc + Add Pornhwa.io

### LINKED ISSUE #11186: [FR] Source Request: Pornhwa.io
### Source name

Pornhwa.io

### Source link

https://pornhwa.io/

### Source language

French 

### Other details

18+/NSFW= Yes 

### Acknowledgements

- [x] I have checked that the extension does not already exist by searching the [extension listing](https://keiyoushi.github.io/extensions/) and verified it does not appear there.
- [x] I have written a meaningful title with the source name.

### 

---

luigidotmoe:
Closes #11186 

Checklist:


AwkwardPeak7 (commented):
they are operated by the same people, I would suggest putting both in a source factory instead of multisrc.

luigidotmoe:
> they are operated by the same people, I would suggest putting both in a source factory instead of multisrc.

I see your point, but is there really a significant benefit to using the factory approach instead of multisrc?

I’m personally not a fan of the idea of an H-source being silently installed alongside an SFW source, even if +18 sources can be disabled. Furthermore, the only factory I’ve found that exists for reasons other than language-specific extensions is ThunderScans, and I don't think it's comparable to this case.

That said, the factory refactor is ready to be pushed if you still believe that is the best path forward.

AwkwardPeak7:
> is there really a significant benefit to using the factory approach instead of multisrc?

just my opinion, that multisrc should be generic for various sites. plus sites by same people serving sameish content should be together. 
Thunderscan case was simply that their ar branch renamed, it was previously thunderscan ar and not lavascan

> H-source being silently installed alongside an SFW source

I see your point

AwkwardPeak7 (approved):

PR #13785, opened 2026-03-04
MyComic: add date_upload and note

hualiong:
Checklist:

### DIFF
--- a/src/zh/mycomic/src/eu/kanade/tachiyomi/extension/zh/mycomic/MyComic.kt
+++ b/src/zh/mycomic/src/eu/kanade/tachiyomi/extension/zh/mycomic/MyComic.kt
@@ -34,9 +34,15 @@ class MyComic :
 
     override fun headersBuilder() = super.headersBuilder().add("Referer", "$baseUrl/")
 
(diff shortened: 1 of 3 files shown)


src/zh/mycomic/src/eu/kanade/tachiyomi/extension/zh/mycomic/MyComic.kt
```diff
@@ -180,7 +180,7 @@ class MyComic :
                 title = "設置首選語言"
                 summary = "當前：%s"
                 entries = arrayOf("繁體中文", "简体中文")
-                entryValues = arrayOf("zh-hant", "zh-hans")
+                entryValues = arrayOf("", "cn")
```
AwkwardPeak7, line 189: changing the preference value like this needs a migration. (you can add one in `getPreferencesLazy { }`)

AwkwardPeak7: should also revert the key change. (PREF_LANGUAGE to pref_key_lang)

AwkwardPeak7 (approved):

PR #14060, opened 2026-03-15
Refactor `:lib:randomua` to set header in `headersBuilder`

AwkwardPeak7:
Setting header in interceptor means the new header will not be available in webview screen, so manual cloudflare bypass breaks.
Now it is set via `headersBuider` meaning it will be in the default `headers` object and thus available to the webview.

Also added a custom lint rule to force override of `getMangaUrl`, since [default implantation ](https://github.com/mihonapp/mihon/blob/77a86d486fc1d1ec2968a51dbda6c8286bb20b55/source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/online/HttpSource.kt#L437-L439)depend on `mangaDetailsRequest` which in turns depends on `headers`. This means we get a network call upon opening manga screen in app, as app needs it for webview button. ([ref](https://github.com/mihonapp/mihon/blob/77a86d486fc1d1ec2968a51dbda6c8286bb20b55/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreen.kt#L106))


buildSrc/src/main/kotlin/RandomUaCheck.kt
```diff
@@ -0,0 +1,25 @@
...
+        "randomua-requires-getMangaUrl",
+        State(),
+        State::toFormatter,
+    )
+
+    private class State : Serializable {
+        fun toFormatter() = FormatterFunc { content ->
+            if ("package keiyoushi.lib.randomua" !in content &&
+                "keiyoushi.lib.randomua" in content &&
+                "override fun getMangaUrl(" !in content
```
vetleledaal, line 16: Would this catch commented out code? Can some AST be used instead? Might be unreasonable to implement.

lib/randomua/src/keiyoushi/lib/randomua/UserAgentPreference.kt
```diff
@@ -0,0 +1,97 @@
...
+    val preferences = source.getPreferences()
+    val customUaPref = EditTextPreference(context).apply {
+        key = PREF_KEY_CUSTOM_UA
+        title = "Custom user agent string"
+        summary = "Leave blank to use the default user agent string"
+        setEnabled(preferences.getPrefUAType() == UserAgentType.OFF)
+        setOnPreferenceChangeListener { _, newValue ->
+            try {
+                Headers.headersOf("User-Agent", newValue as String)
+                Toast.makeText(context, "Restart the app to apply changes", Toast.LENGTH_SHORT).show()
```
vetleledaal, line 70: Can this be avoided in a similar manner we switch `baseUrl` without requiring a restart in some extensions?

vetleledaal (approved):

lib/randomua/src/keiyoushi/lib/randomua/UserAgentPreference.kt
AwkwardPeak7: no unfortunately, [`headers` is hardcoded](https://github.com/mihonapp/mihon/blob/main/source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/online/HttpSource.kt#L62) and extensions can't override it

buildSrc/src/main/kotlin/RandomUaCheck.kt
AwkwardPeak7: spotless passes the raw file, and if comments contain this pattern, it will fail. I'll leave it up to the pr reviewer to catch such case.

If we want AST, `detekt` would be better as a linter but it comes with it's own issues

PR #14087, opened 2026-03-16
Kagane: Allow selecting which filter method to use for popular tab

Amqx:
Checklist:

### DIFF
--- a/src/en/kagane/build.gradle
+++ b/src/en/kagane/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Kagane'
     extClass = '.Kagane'
-    extVersionCode = 23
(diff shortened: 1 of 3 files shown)


AwkwardPeak7 (dismissed):

AwkwardPeak7:
on second thought, you can already use whatever filter you want via the filters (and save them for reuse on some forks (sy or komikku)), so keeping popular for all views is very reasonable. plus I am not a fan of too many settings in an extension.

Amqx:
Fair enough, I just thought it'd be somewhat nice for the popular page to change occasionally. It gets somewhat bland for the opening page to always be the same few titles.

PR #14202, opened 2026-03-21
Comic Fury: Fix chapterListParse not parsing all pages.

### LINKED ISSUE #14147: Comic Fury: Comics missing if there are more than 4 pages in the page list
### Source information

Comic Fury 1.4.6

### Source language

Multi

### Steps to reproduce

Open a series that has more than 4 pages of comics (e.g. https://comicfury.com/read/normalsville/archive)

### Expected behavior

Comics from all pages should appear in the chapter list.

### Actual behavior

Only comics from pages 1, 2, 3, and the last page will appear.

### Mihon/Mihon fork version

Mihon 0.19.4

### Android version

Android 16

### Other details

This is because the page selector element, which the Comic Fury extension uses to find the URL for each page, gets truncated once the page count exceeds 4. In this instance it leads to all the comics on page 4 being skipped.

Image

### Acknowledgements

### 

---

ColouMods:
Currently, the Comic Fury extension uses the page selector element to fetch the URLs for all chapter pages at once. However this breaks if a series has more than 5 pages of comics, since the element will be truncated (in this example page 4 will be missed).

image

To fix this, I reworked the code for parsing the pages so it now iterates through the pages one at a time, only fetching the URL of the page directly after the current (since that will always be present in the page selector).

Closes #14147

Checklist:

### DIFF
--- a/src/all/comicfury/src/eu/kanade/tachiyomi/extension/all/comicfury/ComicFury.kt
+++ b/src/all/comicfury/src/eu/kanade/tachiyomi/extension/all/comicfury/ComicFury.kt
@@ -20,6 +20,7 @@ import okhttp3.HttpUrl
 import okhttp3.HttpUrl.Companion.toHttpUrl
 import okhttp3.Request
 import okhttp3.Response
+import org.jsoup.nodes.Document
 import org.jsoup.nodes.Element
 import java.text.SimpleDateFormat
 import java.util.Locale
@@ -66,9 +67,7 @@ class ComicFury(
      */
     private val archiveSelector = "a:has(div.archive-chapter)"
     private val chapterSelector = "a:has(div.archive-comic)"
-    private val nextArchivePageSelector = "#scroll-content > .onsite-viewer-back-link + .archive-pages a"
-    private val nextChapterPageSelector = "#scroll-content > .webcomic-title-content-inner + .archive-pages a"
-    private lateinit var currentPage: org.jsoup.nodes.Document
+    private val nextPageSelector = "span.vfpagecurrent + a.vfpage"
 
     private fun Element.toSManga(): SChapter = SChapter.create().apply {
         setUrlWithoutDomain(this@toSManga.attr("abs:href"))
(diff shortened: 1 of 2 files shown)


src/all/comicfury/src/eu/kanade/tachiyomi/extension/all/comicfury/ComicFury.kt
```diff
@@ -66,41 +67,46 @@ class ComicFury(
...
         val chapters = mutableListOf<SChapter>()
 
         if (jsp.selectFirst(archiveSelector) != null) {
-            jsp.select(archiveSelector).eachAttr("abs:href").map { url ->
-                chapters.addAll(collect(url))
-                currentPage.select(nextArchivePageSelector).eachAttr("abs:href")
-                    .mapNotNull { nextUrl -> chapters.addAll(collect(nextUrl)) }
+            jsp.select(archiveSelector).eachAttr("abs:href").forEach { url ->
+                val currentPage = client.newCall(GET(url, headers)).execute().asJsoup()
+                chapters.addAll(collect(currentPage))
```
AwkwardPeak7, line 106: I haven't looked at the site but shouldn't the collect method get all chapters? why iterate over archiveselector list?

ColouMods: Some series on the site have the list of comic pages directly on the archive page, however other series split the pages into groups (which the site calls 'chapters' but I think are called 'archives' here to avoid confusion with Mihon's chapters - I wasn't the one who named the selectors so I dunno). So for those series you need iterate through list of chapters, _then_ iterate through the list of comics _inside_ those chapters (which is what the collect method does).

Now there _is_ an issue with the way this works (applies to both the existing code and my reworked one) where, presumably, if a series has more than 100 chapters the chapter list also gets split into pages, so this `forEach` would only parse the first page of chapters. I would have fixed this myself, however when I looked on the site I couldn't actually find any series with anywhere close to 100 chapters (though I don't know the site that well either so maybe there are ones out there). I'd basically have to guess how the page is laid out in order to fix it, and I'd have no way to test it to see if it actually works, so I just left it as-is. It'd probably be better to fix in a seperate PR anyway.

AwkwardPeak7 (approved):

PR #14286, opened 2026-03-25
Vcomycs: Fix thumbail in search can't load in some case

FiorenMas:
Checklist:


src/vi/vcomycs/src/eu/kanade/tachiyomi/extension/vi/vcomycs/Vcomycs.kt
```diff
@@ -107,14 +108,48 @@ class Vcomycs : HttpSource() {
...
+                when {
+                    response.isSuccessful -> true
+                    response.code == 403 || response.code == 405 ->
+                        client.newCall(GET(url, requestHeaders)).execute().use { getResponse ->
+                            getResponse.isSuccessful
+                        }
+                    else -> false
+                }
+            }
+        }.getOrDefault(false)
```
vetleledaal: I think an interceptor could work better. Then you don't need 3x requests per image.

vetleledaal (approved):

PR #14323, opened 2026-03-28
MangaLivre: dont use Madara Load More

zLuckKing:
Manga Livre is blocking Mihon's WebView requests via Cloudflare TLS fingerprint detection. Adding sec-fetch and sec-ch-ua headers makes requests look more like a real browser. Firefox bypasses Cloudflare without cf_clearance, suggesting these headers help distinguish legitimate browser traffic. Library updates work partially with Edge UA but Explore tab fails completely.

Checklist:


zLuckKing:
Update: After testing, library updates and reading work correctly now. However, the Explore/Browse tab still fails with Cloudflare error. The browse endpoint may need additional headers.

Smol-Ame:
Not a knowledgeable contributor, but all parts of the extension works just fine on my end currently.

If you are blocked in WebView/you're getting the "Failed to bypass Cloudflare" error:
1. In the Manga Livre extension settings > Language settings, turn off "Random UA" and clear the "Custom UA" setting in there. Remember to restart your app after changing anything in those settings.
2. If you are still getting the error (either after doing the step above, or not needing to do it), change your app's user-agent.
3. After changing your user-agent, restart the app, then check WebView to see if you pass Cloudflare verification. If you don't pass/still get the blocked error, try a different user-agent.

zLuckKing:
After following your steps (disabled Random UA, cleared Custom UA), the error changed from "Failed to bypass Cloudflare" to "HTTP 400" on the Explore tab. Library still works fine. Could the interceptor headers be conflicting with the browse endpoint request?

Smol-Ame:
Are you still using the testing version with your changes? Try with the current extension if you haven't already.

Since your changes are incompletely/still having issues, I'm converting this into a draft instead.

zLuckKing:
Also worth noting: with the original extension, library updates and reading work correctly now (same as with my changes). The HTTP 400 error only occurs on the Explore/Browse tab. So my PR changes don't seem to break anything, they just don't fix the browse endpoint issue either.

zLuckKing:
I'm from Brazil and I believe this might be a regional issue. Other Brazilian users are also reporting the same problem. The library works fine and manga reading works normally, but the Explore/Browse tab fails with HTTP 400. Cloudflare might have stricter rules for Brazilian IPs.

FelipeGFA:
> problem

thats a problem with User Agent itself, try others User Agents until it works.

this pr does not resolve anything, adding `.header("sec-ch-ua", "\"Microsoft Edge\";v=\"146\", \"Chromium\";v=\"146\", \"Not/A)Brand\";v=\"24\"")` will cause more fingerprint conflicts with the apps user agent

zLuckKing:
I also discovered that the issue might be related to ISP-level DNS blocking in Brazil. Without configuring Cloudflare DNS (1.1.1.1) on my device, the site doesn't resolve at all. With Cloudflare DNS enabled, library works but Explore tab returns HTTP 400. Could the browse endpoint be using a different subdomain that isn't being resolved correctly?

I also tested all Android User Agents listed on whatismybrowser.com and none of them fixed the Explore tab. Edge-based UAs pass the Cloudflare check but still return HTTP 400 on browse. Firefox UAs get stuck in an infinite Cloudflare verification loop. The issue seems to be specific to the browse endpoint, not the User Agent.

zLuckKing:
Found the root cause: the Explore tab uses POST requests to /wp-admin/admin-ajax.php with XHR headers, while the library uses GET requests to normal endpoints. Cloudflare seems to be blocking POST requests to /wp-admin/admin-ajax.php specifically for Brazilian IPs. This explains why library works but Explore doesn't.

zLuckKing:
Confirmed via direct testing: POST requests to /wp-admin/admin-ajax.php return 403 with Cloudflare challenge page ("Just a moment..."). The Mihon cloudflareClient partially bypasses this returning HTTP 400 instead. The browse endpoint requires a valid cf_clearance cookie that the app can't obtain for POST requests.

vetleledaal (approved):

PR #14360, opened 2026-03-29
Komiic: Optimize error message display

hualiong:
Checklist:

### DIFF
--- a/src/zh/komiic/src/eu/kanade/tachiyomi/extension/zh/komiic/Komiic.kt
+++ b/src/zh/komiic/src/eu/kanade/tachiyomi/extension/zh/komiic/Komiic.kt
@@ -33,6 +33,7 @@ class Komiic :
             val origin = chain.request()
             chain.proceed(origin).also {
                 if (it.code == 402 && origin.url.toString().contains("api/image")) {
+                    it.close()
                     throw IOException("今日圖片讀取次數已達上限，請登录或明天再來！")
                 }
             }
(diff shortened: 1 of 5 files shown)


src/zh/komiic/src/eu/kanade/tachiyomi/extension/zh/komiic/Komiic.kt
```diff
@@ -30,9 +30,13 @@ class Komiic :
     override val client = network.cloudflareClient.newBuilder()
         .addInterceptor { chain ->
             refreshToken(chain)
-            chain.proceed(chain.request())
-        }
-        .build()
+            val origin = chain.request()
+            chain.proceed(origin).also {
+                if (it.code == 402 && origin.url.toString().contains("api/image")) {
+                    throw IOException("今日圖片讀取次數已達上限，請登录或明天再來！")
```
hualiong, line 37: Throwing an Exception inside the `Interceptor` will be caught and converted into an `IOException` by `UncaughtExceptionInterceptor`, resulting in the displayed error message being prefixed with "java.lang.Exception: ". Therefore, an `IOException` is directly thrown here.
```kotlin
class UncaughtExceptionInterceptor : Interceptor {
    override fun intercept(chain: Interceptor.Chain): Response {
        return try {
            chain.proceed(chain.request())
        } catch (e: Exception) {
            if (e is IOException) {
                throw e
            } else {
                throw IOException(e)
            }
        }
    }
}
```

AwkwardPeak7, line 37:
```suggestion
                if (it.code == 402 && origin.url.toString().contains("api/image")) {
                    it.close()
                    throw IOException("今日圖片讀取次數已達上限，請登录或明天再來！")
```

AwkwardPeak7 (dismissed):

vetleledaal (approved):

PR #14417, opened 2026-04-01
BiliManga: Optimize the display of manga details

hualiong:
Checklist:


src/zh/bilimanga/src/eu/kanade/tachiyomi/extension/zh/bilimanga/BiliManga.kt
```diff
@@ -50,9 +50,21 @@ class BiliManga :
...
+            name = element.text().toHalfWidthDigits()
+            date_upload = date
+            scanlator = volume
+            setUrlWithoutDomain(url ?: getChapterUrlByContext(i, this@mapToChapter))
+        }
+    }
     private fun String.toHalfWidthDigits(): String = this.map { if (it in '０'..'９') it - 65248 else it }.joinToString("")
 
     companion object {
+        val URL_REGEX = Regex("https?://(?:www\\.)?([-a-zA-Z0-9@:%._+~#=]{2,256}\\.[a-z]{2,6}\\b)*(/[/\\w.-]*)*[?]*(.+)*", RegexOption.IGNORE_CASE)
```
vetleledaal, line 66: Seems like a messy way to match URLs. the TLD matching is for example lacking. Can this be done without a regex? Otherwise I would make it much laxer.

hualiong: This is a rather comprehensive and stringent regular expression I found online for matching valid URLs. I'm concerned that if it's too lenient, it might cause issues with the displayed text in the description.

AwkwardPeak7 (approved):

PR #14563, opened 2026-04-05
Luminare Translations: api changes

manti-X:
Checklist:


src/en/luminaretranslations/src/eu/kanade/tachiyomi/extension/en/luminaretranslations/Dto.kt
```diff
@@ -41,80 +45,60 @@ class Details(
...
-        chapter_number = chapterNumber
+        name = title
+        chapter_number = number
+        date_upload = dateFormat.tryParse(publishedAt)
     }
 }
 
+private val dateFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ", Locale.ROOT).apply {
+    timeZone = TimeZone.getTimeZone("UTC")
+}
```
vetleledaal: Since you already use `Z` (not `'Z'`) in the pattern it should use the correct timezone already.
```suggestion
private val dateFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ", Locale.ROOT)
```

vetleledaal (dismissed):

vetleledaal (approved):

PR #14709, opened 2026-04-11
Revert "Hive Scans: fix Missing FieldException"

MuhamadSyabitHidayattulloh:
Resolved in #14706

### DIFF
--- a/src/en/infernalvoidscans/build.gradle
+++ b/src/en/infernalvoidscans/build.gradle
@@ -3,7 +3,7 @@ ext {
     extClass = '.HiveScans'
     themePkg = 'iken'
     baseUrl = 'https://hivetoons.org'
-    overrideVersionCode = 41
+    overrideVersionCode = 42
     isNsfw = false
 }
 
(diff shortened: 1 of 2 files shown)


src/en/infernalvoidscans/build.gradle
```diff
@@ -3,7 +3,7 @@ ext {
     extClass = '.HiveScans'
     themePkg = 'iken'
     baseUrl = 'https://hivetoons.org'
-    overrideVersionCode = 42
+    overrideVersionCode = 41
```
AaronDeimos:
```suggestion
    overrideVersionCode = 42
```
Even if an update is unnecessary.
Don't lower this keep it the same, ~or anyone who already updated won't get the next actual update.~

AaronDeimos: Oh nevermind my bad
you bumped the multisrc

Now it will send a update but i suppose that's better anyways

MuhamadSyabitHidayattulloh: @AaronDeimos No, you're right. If I revert and override the version, even if I bump the base version, the app version will remain the same! And it's likely that users won't get any updates on the app side.

AaronDeimos: Yes I know I just meant if you didn't intend to update the extension since current also works

AwkwardPeak7 (approved):

PR #14788, opened 2026-04-14
Update CONTRIBUTING.md

darwin-256:
Outdated in some parts

### Update CONTRIBUTING.md

This PR significantly expands and modernizes the `CONTRIBUTING.md` guide to reflect current Keiyoushi repository standards. It introduces full documentation for the reusable `lib/` modules, the `keiyoushi.utils` core utilities library, and adds important performance, memory-safety, and best-practice guidelines for extension developers.

### Changes

**General & File Structure**
- Updated title and several references from “Tachiyomi” to “Keiyoushi” where appropriate.
- Fixed grammar, wording, spelling, and typo issues throughout the document.
- Updated the example extension file structure to include the optional helper files (`Dto.kt`, `Filters.kt`, `UrlActivity.kt`).
- Added an explicit note enforcing the new convention: **do not repeat the extension name** in helper files (with a note that older extensions may still use the old pattern).

**New Section: `lib tools`**
- Added a complete new top-level section documenting the `lib/` directory and all reusable Gradle modules.
- Included a table listing every available library with a short description and link.
- Explained how to add a dependency and how to create a new lib (structure, `lib-android` plugin, KDoc recommendations).

**New Section: `keiyoushi.utils` (Core Utilities)**
- Comprehensive documentation for the shared utilities available to every extension without extra dependencies.
- Covered:
  - JSON handling (`parseAs` / `toJsonString`, why you should **not** create your own `Json` instance, preference for regular `class` + `@SerialName` over `data class` for DTOs).
  - Date parsing (`tryParse` + reusable class/file-level `SimpleDateFormat`).
  - Filter helpers (`firstInstance` / `firstInstanceOrNull`).
  - Preferences (`getPreferences` / `getPreferencesLazy`).
  - Next.js data extraction (`extractNextJs` / `extractNextJsRsc`).
  - URL helpers (recommended use of `setUrlWithoutDomain` + `absUrl()`).

**New Section: HTML and Image Processing**
- New best-practice guidelines:
  - Use `Jsoup.parseBodyFragment(html, baseUrl)` for partial HTML returned by APIs.
  - Simple chapter number formatting with `.toString().removeSuffix(".0")`.
  - Elegant `Page` list creation with `mapIndexed`.
- **Memory-efficient image interceptors** – critical new guidance on stream-based processing (`byteStream()`, Okio `Buffer`, `cipherSource`, `bitmap.recycle()`) to avoid `OutOfMemoryError` on low-end devices.

**New Section: OkHttp and Network**
- New recommendations:
  - Extract tokens (CSRF, nonces, etc.) inside interceptors using `response.peekBody(Long.MAX_VALUE).string()`.
  - GraphQL queries using Kotlin raw multi-dollar strings (`$$"""..."""`).
  - Use `.isNotEmpty()` with Jsoup `.text()` (it is already trimmed).

**Misc notes & Advanced features (expanded)**
- Added many new best-practice bullets:
  - Store minimal unique path segments in `SManga.url` / `SChapter.url` to avoid forcing library migrations on URL changes.
  - Never store request state in the source class (concurrency issues).
  - Prefer `HttpUrl.Builder`, `absUrl()`, `response.asJsoup()`, cached `Regex`, `buildString {}`, etc.
  - Avoid unnecessary overrides of default `HttpSource` methods.
  - Do not hardcode `User-Agent` unless strictly required.
- New subsection **“Extension logic and app features”**:
  - `title` and `url` on `SManga` are mandatory – fail loudly if missing.
  - When (and when not) to bump `versionId`.
  - Implement `UnmeteredSource` for self-hosted servers.
  - `ConfigurableSource` automatically saves preference values (no manual handling needed in listeners).

**URL Intent Filter**
- Small improvement in the `UrlActivity` example (catch `Throwable` instead of only `ActivityNotFoundException`).

**Other**
- Minor wording and consistency fixes in the rest of the document (Running, Debugging, Building, PR checklist, etc.).


CONTRIBUTING.md
FelipeGFA, line 230: this isnt outdated? a lot of extensions use smth like: `MyExtentionDto.kt`

darwin-256: Does it matter? It's not like it affect runtime behavior. The only reason to changes this would be Consistency: helps keep naming style uniform across extensions.

CONTRIBUTING.md
```diff
@@ -324,7 +462,34 @@ either `SourceFactory` or extend one of the `Source` implementations: `HttpSourc
...
+- **Formatting Chapter Numbers:** Do not write custom `DecimalFormat` logic just to remove trailing zeros from float chapter numbers. Simply use `.toString().removeSuffix(".0")`.
+
+- **Generating Page lists:** Use Kotlin's `mapIndexed` to generate your `Page` lists rather than maintaining your own counter or index variable:
+```
+  return document.select(".pages img").mapIndexed { index, img ->
+      Page(index, imageUrl = img.attr("abs:src"))
+  }
+```
+
+* **Memory-efficient Image Interceptors:** If you write an interceptor to descramble or stitch images, do not load the entire image into a `ByteArray`. Use streams to avoid `OutOfMemory` errors on lower-end devices:
```
manti-X: ```okio.cipherSource``` for decryption

CONTRIBUTING.md
luigidotmoe: > this isnt outdated? a lot of extensions use smth like: `MyExtentionDto.kt`

fairly new, see #14484
i try to rename files when working on old exts

CONTRIBUTING.md
```diff
@@ -445,25 +610,37 @@ will be cached.
 - While a chapter is open in the reader or is being downloaded, `fetchImageUrl` will be called to get
 URLs for each page of the manga if the `Page.imageUrl` is empty.
 - If the source provides all the `Page.imageUrl`'s directly, you can fill them and let the `Page.url`
-empty, so the app will skip the `fetchImageUrl` source and call directly `fetchImage`.
+empty, so the app will skip the `fetchImageUrl` step and call directly `fetchImage`.
 - The `Page.url` and `Page.imageUrl` attributes **should be set as an absolute URL**.
 - Chapter pages numbers start from `0`.
 - The list of `Page`s should be returned already sorted, the `index` field is ignored.
 
 ### Misc notes
```
manti-X, line 680: consider to save only unique pathsegments e.g. id in ```getUrlWithoutDomain``` / ```url``` and build url by overriding the functions so users maybe don't have to migrate

```url = "/manga/some-id"``` -> ```url = "some-id"```

File: CONTRIBUTING.md [hunk 3]
```diff
@@ -445,25 +610,37 @@ will be cached.
...
+- **Jsoup `.text()` is already trimmed:** Calling `element.text().trim()` is redundant because Jsoup automatically normalizes and trims whitespace. Just use `element.text()`.
+- **Use named parameters for `Page`:** When instantiating `Page` objects, use the named parameter for the image URL: `Page(index, imageUrl = url)` instead of passing an empty string as the second argument (`Page(index, "", url)`).
+- **Throw `UnsupportedOperationException`:** If a source uses an API and doesn't need to parse HTML for images, override `imageUrlParse(response: Response)` and throw `UnsupportedOperationException()` instead of returning an empty string. Also use this pattern for unused inherited methods.
+- **Cache Regex instances:** Define `Regex` instances at the class level or in a `companion object` so they aren't recompiled on every method call.
+- **Do not hardcode `User-Agent`:** Unless absolutely necessary to bypass Cloudflare/protection, do not hardcode a specific `User-Agent`. Calling `super.headersBuilder()` already provides the app's default User-Agent.
```
manti-X: not always because of cf, sometime also because of different selector etc...

CONTRIBUTING.md
```diff
@@ -287,6 +287,144 @@ dependencies {
...
+
+```kotlin
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+// Bad: Using data class and snake_case variable names
+@Serializable
+data class MyDto(val manga_id: Int, val cover_img: String)
+
+// Good: Regular class, camelCase variables mapped with @SerialName
```
manti-X: recommend to make fields private if you pass it to e.g ```SManga()``` etc... in dto

CONTRIBUTING.md
```diff
@@ -324,7 +462,34 @@ either `SourceFactory` or extend one of the `Source` implementations: `HttpSourc
...
+```
+  return document.select(".pages img").mapIndexed { index, img ->
+      Page(index, imageUrl = img.attr("abs:src"))
+  }
+```
+
+* **Memory-efficient Image Interceptors:** If you write an interceptor to descramble or stitch images, do not load the entire image into a `ByteArray`. Use streams to avoid `OutOfMemory` errors on lower-end devices:
+
+  * **Read:** Use `response.body.byteStream()` with `BitmapFactory.decodeStream()`.
+  * **Write:** Use Okio's `Buffer` (`val output = Buffer()`) and `output.outputStream()` to compress the new `Bitmap`, then return `output.readByteArray().toResponseBody(mediaType)`.
```
manti-X: why ```readByteArray()```?

Also use ```asResponseBody()``` instead of ```toResponseBody()```.

And use ```recycle()``` if possible

CONTRIBUTING.md
```diff
@@ -324,7 +471,36 @@ either `SourceFactory` or extend one of the `Source` implementations: `HttpSourc
...
+
+* **Memory-efficient Image Interceptors:** When implementing interceptors for descrambling, stitching, or decrypting images, avoid loading the entire image into a `ByteArray`, as this can cause `OutOfMemoryError` on low-end devices. Prefer stream-based processing instead:
+
+  * **Read:** Use `response.body.byteStream()` with `BitmapFactory.decodeStream()` to decode images directly from the stream.
+  * **Write:** Write the processed bitmap into an Okio `Buffer` via `output.outputStream()` and convert it using `asResponseBody(mediaType)`.
+  * **Decryption:** Use Okio's `cipherSource` extension for stream-based decryption rather than decrypting a full byte array in memory.
+  * Note: `readByteArray()` should generally be avoided here because it forces full in-memory buffering of the image. Streaming directly keeps memory usage lower and more stable.
+  * If applicable, call `bitmap.recycle()` after you're done with it to free native memory early.
+
+### OkHttp and Network
```
cuong-tran, line 550: Suggest to use `response.use { ... }` to avoid memory leaks.

CONTRIBUTING.md
```diff
@@ -287,6 +291,149 @@ dependencies {
     implementation(project(':lib-i18n'))
 }
 ```
+#### keiyoushi.utils (core utilities)
+
+The `core/utils` module provides a set of shared extension functions that are available to all extensions
+without any extra Gradle dependency. You should use these helpers instead of rolling your own equivalents —
+reviewers will ask you to switch if you don't. The utilities live in the `keiyoushi.utils` package and are
+imported individually.
```
Starmania:
```suggestion
without any extra Gradle dependency. You should use these helpers instead of rolling your own equivalents.
The utilities live in the `keiyoushi.utils` package and are imported individually.
```

CONTRIBUTING.md
```diff
@@ -287,6 +291,149 @@ dependencies {
...
+
+// With a transform applied before parsing:
+val dto = response.parseAs<MyDto> { it.substringAfter("callback(").dropLast(1) }
+```
+
+**Do not** create a local `private val json: Json by injectLazy()` — the global instance is already
+available via `jsonInstance` and the `parseAs` helpers use it automatically. Repeated in several PRs
+(e.g. [#13619](https://github.com/keiyoushi/extensions-source/pull/13619),
+[#13813](https://github.com/keiyoushi/extensions-source/pull/13813),
+[#13968](https://github.com/keiyoushi/extensions-source/pull/13968)).
```
Starmania: This may be needed in some case extreme case. I think that you already covered the point by giving some example code.

CONTRIBUTING.md
```diff
@@ -287,6 +291,149 @@ dependencies {
...
+
+```kotlin
+import keiyoushi.utils.extractNextJs
+
+val data = response.extractNextJs<MyDto>()
+// Or with an explicit predicate:
+val data = document.extractNextJs<MyDto> { element ->
+    element is JsonObject && "slug" in element
+}
+```
```
Starmania: This PR implementing that is not merged, but be sure that I will cherry pick this later

CONTRIBUTING.md
```diff
@@ -445,25 +622,38 @@ will be cached.
...
+- **Media Types:** `application/json` is intrinsically UTF-8. Avoid using `application/json; charset=utf-8`. Use `"application/json".toMediaType()`.
+- **Use `getUrlWithoutDomain` carefully:** It can be useful when parsing target source URLs, but note a current issue with spaces—replace them with URL-encoded characters (e.g., `%20`).
+- **Follow `HttpSource` workflow:** Stick to the general workflow from this base class when possible; deviating may introduce unnecessary complexity.
+- **Configurable sources:** By implementing `ConfigurableSource`, you can add settings backed by `SharedPreferences`.
 
 ### Advanced Extension features
 
+#### Extension logic and app features
+
+- **Mandatory SManga fields:** A manga's `title` and `url` are **mandatory**. Do not provide generic fallbacks like `"Untitled"` or `"Unknown"` if the site fails to provide a title, as this breaks downloads and library management. If the title is missing, it is better to drop the entry entirely or throw an exception.
```
Starmania: Throw, silent error should be documented

CONTRIBUTING.md
```diff
@@ -445,25 +622,38 @@ will be cached.
...
+- **When to bump `versionId`:** The `versionId` property dictates how the app tracks the source. **Only override and bump `versionId` if the source's URL structure fundamentally changes** (e.g., old manga URLs no longer work and do not redirect). Bumping this forces all users to re-migrate their bookmarks.
```
Starmania:
```suggestion
- **When to bump `versionId`:** The `versionId` property dictates how the app tracks the source. **Only override and bump `versionId` if the source's URL structure fundamentally changes** (e.g., old manga URLs no longer work and there is no way to create a redirect). Bumping this forces all users to re-migrate their bookmarks. 
```

Starmania (commented):
AI will go brrr after this merge

CONTRIBUTING.md
```diff
@@ -214,9 +214,12 @@ src/<lang>/<mysourcename>/
                 └── extension
                     └── <lang>
                         └── <mysourcename>
-                            └── <MySourceName>.kt
+                            ├── <MySourceName>.kt
+                            ├── <Dto>.kt (optional)
+                            ├── <Filters.kt (optional)
```
al-shamiry:
```suggestion
                            ├── <Filters>.kt (optional)
```

CONTRIBUTING.md
```diff
@@ -214,9 +214,12 @@ src/<lang>/<mysourcename>/
...
+                            └── <UrlActivity>.kt (optional)
 
-13 directories, 9 files
+14 directories, 7(+4) files
```
al-shamiry:
```suggestion
14 directories, 7(+3) files
```
3 optional files?

FelipeGFA:
> AI will go brrr after this merge

what about something like [this](https://github.com/cohen-liel/hivemind/blob/main/CONTRIBUTING.md#ai-generated-contributions)

CONTRIBUTING.md
```diff
@@ -287,6 +291,149 @@ dependencies {
...
+// Or with an explicit predicate:
+val data = document.extractNextJs<MyDto> { element ->
+    element is JsonObject && "slug" in element
+}
+```
+
+For client-side navigation responses (`text/x-component` content type), pass the `rsc: 1`
+request header and use `extractNextJsRsc` on the response body string.
+See [#14266](https://github.com/keiyoushi/extensions-source/pull/14266) and
+[#14446](https://github.com/keiyoushi/extensions-source/pull/14446) for real-world usage.
```
al-shamiry, line 484:
```suggestion

**Extracting URLs — `setUrlWithoutDomain` + `absUrl`**
When extracting URLs from HTML, prefer `element.absUrl("href")` or `element.attr("abs:href")` over manually concatenating `baseUrl` + `path`. Combined with `setUrlWithoutDomain()`, this safely handles both absolute and relative links.

```

CONTRIBUTING.md
al-shamiry:
```suggestion
**Mandatory SManga fields:** A manga's `title` and `url` are **mandatory**. Do not provide generic fallbacks like `"Untitled"` or `"Unknown"` if the site fails to provide a title, as this breaks downloads and library management. If the title is missing, let it throw an exception.
```
```kotlin
// Bad: silent fallback hides broken selectors
title = doc.selectFirst("h1")?.text() ?: "Unknown"
title = doc.selectFirst("h1")?.text().orEmpty()
// Good: fail loudly so broken selectors are caught
title = doc.selectFirst("h1")!!.text()
```

CONTRIBUTING.md
```diff
@@ -445,25 +622,38 @@ will be cached.
...
-- By implementing `ConfigurableSource` you can add settings to your source, which is backed by
-[`SharedPreferences`](https://developer.android.com/reference/android/content/SharedPreferences).
+- **Store minimal unique URLs:** Consider storing only unique path segments (e.g., the ID) in the `SManga.url` or `SChapter.url` properties and building the full URL by overriding request methods (`mangaDetailsRequest`, `chapterListRequest`, `getMangaUrl`, etc.). For example, instead of `url = "/manga/some-id"`, use `url = "some-id"`. This prevents users from being forced to migrate their library if the site's base URL or routing path changes.
+- **Do not store state in the source class:** Extensions are instantiated once per session. Storing request state (like `lastQuery`, `page`, or auth tokens that need rapid refreshing) in class-level variables can cause concurrency bugs. Pass state locally through the fetch methods instead.
+- **Use `HttpUrl` for URL building:** Rely on `okhttp3.HttpUrl.Builder` to build URLs and encode query parameters safely instead of string concatenation. Use `HttpUrl.pathSegments` instead of `String.split("/")` when parsing URLs.
```
al-shamiry:
```suggestion
- **Use `HttpUrl` for URL building:** Rely on `okhttp3.HttpUrl.Builder` to build URLs and encode query parameters safely instead of string concatenation. Use `HttpUrl.pathSegments` and `HttpUrl.queryParameter()` instead of `String.split()` or regex when parsing URLs and extracting query values.
```

al-shamiry (commented):
bits and bobs

CONTRIBUTING.md
cuong-tran: Other than mandatory fields, avoid using double-bang operator (!!) or causing exception on other fields for more stable parsing (I can't find the right words here)

al-shamiry: For all other fields, prefer safe calls `?.` over the double-bang operator `!!`. A missing thumbnail or description should not crash the entire parse, only `title` and `url` warrant a hard failure.

CONTRIBUTING.md
darwin-256: 4. 1 xml file and 3 kt files

CONTRIBUTING.md
darwin-256: > ```suggestion
> without any extra Gradle dependency. You should use these helpers instead of rolling your own equivalents.
> The utilities live in the `keiyoushi.utils` package and are imported individually.
> ```

This may need to be worded differently  like saying dont implement  your own solution  the reviews  will ask you to switch to utils

CONTRIBUTING.md
luigidotmoe: Not a universal rule, there is technically no limit to optional files (some extensions need interceptors that could be in separate file(s), for example)

CONTRIBUTING.md
darwin-256:
```suggestion
without any extra Gradle dependency. Prefer using these helpers instead of implementing your own equivalents, as they provide standardized and maintained solutions.
The utilities live in the `keiyoushi.utils` package and are imported individually.
```

I think this wording is better

CONTRIBUTING.md
darwin-256: Resolved in a011e8a2103f6e01f652a9f2e5c880a3b9a0dd6b

darwin-256: Resolved in a011e8a2103f6e01f652a9f2e5c880a3b9a0dd6b

CONTRIBUTING.md
darwin-256:
```suggestion
**Do not** create a local `private val json: Json by injectLazy()` unless you specifically need a custom JSON configuration (e.g., `isLenient = true` or custom serializers). For standard parsing, the global instance is already available via `jsonInstance` and the `parseAs` helpers use it automatically.
```

Maybe this would be better?

CONTRIBUTING.md
darwin-256:
```suggestion
```

Do we even need to include the directory and file count? This block is only meant to show the basic directory structure and the types of files that might be there. I suggest we remove that line entirely to avoid confusion.

CONTRIBUTING.md
Starmania: This may require a Member's opinion, but this isn't conditional anymore for us.

KarinaFromAespa, line 1102: In settings.gradle.kts, the dev can mention the single extension in the load individual extension function. It helps to avoid loading unnecessary modules and making the build process more efficient and preventing the CPU go brrrrrrr.

KarinaFromAespa: Include this too.

CONTRIBUTING.md
```diff
@@ -1,10 +1,10 @@
 # Contributing
 
-This guide have some instructions and tips on how to create a new Tachiyomi extension. Please **read
+This guide has some instructions and tips on how to create a new Tachiyomi extension. Please **read
```
manti-X: > Tachiyomi 

Tachiyomi is dead

CONTRIBUTING.md
```diff
@@ -445,25 +623,40 @@ will be cached.
 - While a chapter is open in the reader or is being downloaded, `fetchImageUrl` will be called to get
 URLs for each page of the manga if the `Page.imageUrl` is empty.
 - If the source provides all the `Page.imageUrl`'s directly, you can fill them and let the `Page.url`
-empty, so the app will skip the `fetchImageUrl` source and call directly `fetchImage`.
+empty, so the app will skip the `fetchImageUrl` step and call directly `fetchImage`.
 - The `Page.url` and `Page.imageUrl` attributes **should be set as an absolute URL**.
 - Chapter pages numbers start from `0`.
 - The list of `Page`s should be returned already sorted, the `index` field is ignored.
 
 ### Misc notes
```
manti-X, line 680: recommend if needed/possible pass additional needed data via fragment and no need to remove it afterwards since okhttp doesn't send it to the server https://square.github.io/okhttp/5.x/okhttp/okhttp3/-http-url/index.html#:~:text=may%20be%20repeated.-,Fragment,-The%20fragment%20is

CONTRIBUTING.md
```diff
@@ -324,7 +472,37 @@ either `SourceFactory` or extend one of the `Source` implementations: `HttpSourc
...
-| `id`      | Identifier of your source, automatically set in `HttpSource`. It should only be manually overriden if you need to copy an existing autogenerated ID.            |
+| `id`      | Identifier of your source, automatically set in `HttpSource`. It should only be manually overridden if you need to copy an existing autogenerated ID.           |
+
+### HTML and Image Processing
+
+- **Parsing partial HTML:** If an API returns a JSON response containing an HTML string, use `Jsoup.parseBodyFragment(html, baseUrl)` instead of `Jsoup.parse(html)`. Passing the `baseUrl` ensures that `abs:href` and `absUrl()` can correctly resolve relative links.
+
+- **Formatting Chapter Numbers:** Do not write custom `DecimalFormat` logic just to remove trailing zeros from float chapter numbers. Simply use `.toString().removeSuffix(".0")`.
+
+- **Generating Page lists:** Use Kotlin's `mapIndexed` to generate your `Page` lists rather than maintaining your own counter or index variable:
```
manti-X: use the index from the response if it provides one

manti-X:
are you gonna rewrite the ```lib``` and ```lib-multisrc``` parts too?👀

CONTRIBUTING.md
darwin-256:
```suggestion
This guide has some instructions and tips on how to create a new Keiyoushi extension. Please **read
```

CONTRIBUTING.md
darwin-256:
```suggestion
- **Generating Page lists:** Prefer using the index provided by the source (API/HTML) when available, as it ensures correct ordering and consistency. Only use Kotlin's `mapIndexed` to generate indices when the source does not provide one:
```

this better?

darwin-256:
> are you gonna rewrite the `lib` and `lib-multisrc` parts too?👀

lib-multisrc I don't think, so I had hard time building one.

As for lib I think I can, It's going to take some time though.

CONTRIBUTING.md
manti-X: > to generate indices

kinda redundant because ```mapIndexed``` already implies it

```Fall back to Kotlin’s mapIndexed only if no index is provided.``` ?

**Comment by darwin-256** [comment]
@luigidotmoe @Starmania @bapeey @vetleledaal  need some opinions on [2a2a0f6](https://github.com/keiyoushi/extensions-source/pull/14788/commits/2a2a0f67c152cfe5eea5983ccf9c0330d3a3133d) I'm pretty sure its overkill to explain all the libs so tell me which to remove which to keep

**Comment by Starmania** [comment]
> I'm pretty sure its overkill to explain all the libs so tell me which to remove which to keep

(Srx for the ping)

Why not move specific library documentation in his folder ?

So this way you could just say that every library has his documentation where it is written, and you could do a really fast forward the explanation on what's the library do, just enough to know it exists.

File: CONTRIBUTING.md [hunk 19]
```diff
@@ -264,30 +267,235 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+| [`lib-randomua`](https://github.com/keiyoushi/extensions-source/tree/main/lib/randomua) | Fetches and rotates real-world User-Agent strings |
+| [`lib-seedrandom`](https://github.com/keiyoushi/extensions-source/tree/main/lib/seedrandom) | Seeded PRNG matching JavaScript's `seedrandom` library |
+| [`lib-speedbinb`](https://github.com/keiyoushi/extensions-source/tree/main/lib/speedbinb) | SpeedBinb manga viewer protocol and image descrambling |
+| [`lib-synchrony`](https://github.com/keiyoushi/extensions-source/tree/main/lib/synchrony) | JavaScript deobfuscation via the Synchrony engine (QuickJS sandbox) |
+| [`lib-textinterceptor`](https://github.com/keiyoushi/extensions-source/tree/main/lib/textinterceptor) | Renders plain text or HTML as a PNG image page |
+| [`lib-unpacker`](https://github.com/keiyoushi/extensions-source/tree/main/lib/unpacker) | Unpacks Dean Edwards–packed JavaScript; substring extraction helpers |
+| [`lib-zipinterceptor`](https://github.com/keiyoushi/extensions-source/tree/main/lib/zipinterceptor) | Extracts and stitches images delivered as ZIP archives |
```
AwkwardPeak7: I think we should not mention all available ones, only mention most used/helpful ones like cookie interceptor or dataimage one? the line above saying to check `lib/` directory should be enough

CONTRIBUTING.md
```diff
@@ -264,30 +267,235 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+
+### Adding a lib dependency
+
+Declare the module in your extension's `build.gradle`:
+
+```groovy
+dependencies {
+    implementation(project(':lib-<name>'))
```
AwkwardPeak7:
```suggestion
    implementation(project(':lib:<name>'))
```

CONTRIBUTING.md
```diff
@@ -264,30 +267,235 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+}
+```
+
+For example:
 
 ```groovy
 dependencies {
     implementation(project(':lib-dataimage'))
```
AwkwardPeak7:
```suggestion
    implementation(project(':lib:dataimage'))
```

CONTRIBUTING.md
AwkwardPeak7: app ignores any index passed to to `Page`, so it can be anything. developers need to make sure the page list is sorted correctly (according to source) and not rely on the index

CONTRIBUTING.md
```diff
@@ -324,7 +532,37 @@ either `SourceFactory` or extend one of the `Source` implementations: `HttpSourc
...
+  * **Write:** Write the processed bitmap into an Okio `Buffer` via `output.outputStream()` and convert it using `asResponseBody(mediaType)`.
+  * **Decryption:** Use Okio's `cipherSource` extension for stream-based decryption rather than decrypting a full byte array in memory.
+  * Note: `readByteArray()` should generally be avoided here because it forces full in-memory buffering of the image. Streaming directly keeps memory usage lower and more stable.
+  * Always wrap network responses in `response.use { ... }` to ensure the response body is properly closed and to prevent memory leaks.
+  * If applicable, call `bitmap.recycle()` after you're done with it to free native memory early.
+
+### OkHttp and Network
+
+* **Extracting tokens via Interceptors:** If you need to grab a CSRF token or nonce from a website's HTML to use in an API request, do not make a separate `GET` request just to find it. Instead, capture it in an `Interceptor` during the normal browsing flow using `response.peekBody(Long.MAX_VALUE).string()`. Peeking ensures the body isn't consumed and can still be parsed downstream.
```
AwkwardPeak7: this kind of depends on the specific case, I would say remove from the guide

CONTRIBUTING.md
```diff
@@ -445,25 +683,40 @@ will be cached.
...
+- **Do not store state in the source class:** Extensions are instantiated once per session. Storing request state (like `lastQuery`, `page`, or auth tokens that need rapid refreshing) in class-level variables can cause concurrency bugs. Pass state locally through the fetch methods instead.
+- **Use `HttpUrl` for URL building:** Rely on `okhttp3.HttpUrl.Builder` to build URLs and encode query parameters safely instead of string concatenation. Use `HttpUrl.pathSegments` and `HttpUrl.queryParameter()` instead of `String.split()` or regex when parsing URLs and extracting query values.
+- **Do not override default methods needlessly:** Do not override methods like `mangaDetailsRequest`, `chapterListRequest`, `pageListRequest`, or `getMangaUrl` if you are only returning the default behavior (e.g., `GET(baseUrl + manga.url, headers)`).
```
AwkwardPeak7: this is not necessary, i would say remove it

CONTRIBUTING.md
```diff
@@ -445,25 +683,40 @@ will be cached.
...
+- **Follow `HttpSource` workflow:** Stick to the general workflow from this base class when possible; deviating may introduce unnecessary complexity.
+- **Configurable sources:** By implementing `ConfigurableSource`, you can add settings backed by `SharedPreferences`.
 
 ### Advanced Extension features
 
+#### Extension logic and app features
+
+- **Mandatory SManga fields:** A manga's `title` and `url` are **mandatory**. Do not provide generic fallbacks like `"Untitled"` or `"Unknown"` if the site fails to provide a title, as this breaks downloads and library management.
+  Prefer failing loudly (e.g., throwing an exception) so broken selectors are detected early. Silent fallbacks or empty values can hide issues and make debugging harder. If the title or url is missing, it is better to throw or skip the entry entirely.
+- **Optional fields:** For all other fields, prefer safe calls (`?.`) and avoid using the double-bang operator (`!!`). Missing data like thumbnails or descriptions should not crash the entire parsing process.
```
AwkwardPeak7:
```suggestion
- **Optional fields:** For all other fields, prefer safe calls (`?.`) and avoid using non-null assertion (`!!`). Missing data like thumbnails or descriptions should not crash the entire parsing process.
```

CONTRIBUTING.md
```diff
@@ -680,10 +933,10 @@ multisrc
                 └── ThemeSourceGenerator.kt
 ```
 
-- `multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/<themepkg>/<Theme>.kt` defines the the theme's
+- `multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/<themepkg>/<Theme>.kt` defines the theme's
 default implementation.
-- `multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/<theme>/<Theme>Generator.kt` defines the the
-theme's generator class, this is similar to a `SourceFactory` class.
+- `multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/<theme>/<Theme>Generator.kt` defines the theme's
```
AwkwardPeak7: current multisrc section is outdated, we don't have generators anymore. consider updating this too. 
basically no we have lib-multisrc which is gradle libraries, which you can apply via `themePkg` in build.gradle and inherit from them.

CONTRIBUTING.md
vetleledaal: Most used:
```bash
$ rg --no-filename -o -g'build.gradle' ':lib[^\'"]+' | sort | uniq -c | sort -nr
     22 :lib:randomua
     19 :lib:cookieinterceptor
     12 :lib:textinterceptor
     11 :lib:cryptoaes
     10 :lib:unpacker
      9 :lib:i18n
      6 :lib:speedbinb
      6 :lib:publus
      4 :lib:synchrony
      2 :lib:seedrandom
      2 :lib:dataimage
      2 :lib:clipstudioreader
      1 :lib:zipinterceptor
      1 :lib:lzstring
```

I would say to keep randomua, cookieinterceptor, cryptoaes, unpacker, synchrony. Maybe textinterceptor. I'm not that familiar with the others.

CONTRIBUTING.md
MuhamadSyabitHidayattulloh: > 

Touching on the use of MapNotNull might be better

AwkwardPeak7 (dismissed):

CONTRIBUTING.md
```diff
@@ -264,30 +264,225 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+Place your code in the package `keiyoushi.lib.<mylibname>`. Document public API with KDoc so
+contributors can understand the lib without needing to read `CONTRIBUTING.md`.
+
+#### keiyoushi.utils (core utilities)
+
+The `core/utils` module provides a set of shared extension functions that are available to all extensions
+without any extra Gradle dependency. Prefer using these helpers instead of implementing your own equivalents, as they provide standardized and maintained solutions.
+The utilities live in the `keiyoushi.utils` package and are imported individually.
+
+**JSON parsing — `parseAs`**
```
vetleledaal: Avoid em-dash. Not easy to type on a normal keyboard.
```suggestion
**JSON parsing - `parseAs`**
```

CONTRIBUTING.md
```diff
@@ -264,30 +264,225 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+
+Use `keiyoushi.utils.toJsonString` to serialize an object to a JSON string.
+
+```kotlin
+import keiyoushi.utils.toJsonString
+
+val body = myRequestDto.toJsonString().toRequestBody("application/json".toMediaType())
+```
+
+**JSON models (DTOs) and serialization**
```
vetleledaal: Try to avoid using emphasis when you intend to use a heading. You usually want sections to be visible from the ToC.
```suggestion
##### JSON models (DTOs) and serialization
```

CONTRIBUTING.md
```diff
@@ -324,7 +519,35 @@ either `SourceFactory` or extend one of the `Source` implementations: `HttpSourc
...
+- **Parsing partial HTML:** If an API returns a JSON response containing an HTML string, use `Jsoup.parseBodyFragment(html, baseUrl)` instead of `Jsoup.parse(html)`. Passing the `baseUrl` ensures that `abs:href` and `absUrl()` can correctly resolve relative links.
+
+- **Formatting Chapter Numbers:** Do not write custom `DecimalFormat` logic just to remove trailing zeros from float chapter numbers. Simply use `.toString().removeSuffix(".0")`.
+
+- **Generating Page lists:** The app ignores the `index` passed to the `Page` object, but you must ensure the list itself is sorted correctly according to the source. You can use Kotlin's `mapIndexed` to easily instantiate `Page` objects, or rely on the index provided by the source API if available:
+```kotlin
+  return document.select(".pages img").mapIndexed { index, img ->
+      Page(index, imageUrl = img.attr("abs:src"))
+  }
+```
```
vetleledaal: Indent the codeblock so it renders properly inside this bullet point. Try to indent by an even amount of spaces (not 3), even if it doesn't line up with the bullet point.

vetleledaal (dismissed):

CONTRIBUTING.md
```diff
@@ -1,10 +1,10 @@
 # Contributing
 
-This guide have some instructions and tips on how to create a new Tachiyomi extension. Please **read
+This guide has some instructions and tips on how to create a new Keiyoushi extension. Please **read
```
vetleledaal: 1 trailing space is the same as 0 trailing spaces. Did you intend to use 2 trailing spaces to force a line break? Unlikely in the middle of emphasized text.
```suggestion
This guide has some instructions and tips on how to create a new Keiyoushi extension. Please **read
```

CONTRIBUTING.md
```diff
@@ -264,30 +267,225 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+
+```kotlin
+import keiyoushi.utils.toJsonString
+
+val body = myRequestDto.toJsonString().toRequestBody("application/json".toMediaType())
+```
+
+##### JSON models (DTOs) and serialization
+
+When defining `@Serializable` classes for JSON parsing, **do not** use `data class` unless you actually need data class features (like `copy()` or destructuring). Use a regular `class` instead to reduce the generated bytecode size.
```
vetleledaal:
```suggestion
When defining `@Serializable` classes for JSON parsing, **do not** use `data class` unless you actually need data class features (like `copy()` or destructuring). Use a regular `class` instead to reduce the generated bytecode size.
```

CONTRIBUTING.md
```diff
@@ -443,27 +670,39 @@ will be cached.
...
+
+- **Mandatory SManga fields:** A manga's `title` and `url` are **mandatory**. Do not provide generic fallbacks like `"Untitled"` or `"Unknown"` if the site fails to provide a title, as this breaks downloads and library management.
+  Prefer failing loudly (e.g., throwing an exception) so broken selectors are detected early. Silent fallbacks or empty values can hide issues and make debugging harder. If the title or url is missing, it is better to throw or skip the entry entirely.
+- **Optional fields:** For all other fields, prefer safe calls (`?.`) and avoid using the non-null assertion (`!!`). Missing data like thumbnails or descriptions should not crash the entire parsing process. Consider using Kotlin's `mapNotNull` when parsing lists of elements so that if a single item fails, the rest of the list can still be loaded successfully.
+- **When to bump `versionId`:** The `versionId` property dictates how the app tracks the source. **Only override and bump `versionId` if the source's URL structure fundamentally changes** (e.g., old manga URLs no longer work and there is no way to create a redirect). Bumping this forces all users to re-migrate their bookmarks.
```
vetleledaal:
```suggestion
- **When to bump `versionId`:** The `versionId` property dictates how the app tracks the source. **Only override and bump `versionId` if the source's URL structure fundamentally changes** (e.g., old manga URLs no longer work and there is no way to create a redirect). Bumping this forces all users to re-migrate their bookmarks.
```

CONTRIBUTING.md
```diff
@@ -255,7 +258,7 @@ apply from: "$rootDir/common.gradle"
 The extension's version name is generated automatically by concatenating `1.4` and `extVersionCode`.
 With the example used above, the version would be `1.4.1`.
 
-### Core dependencies
+## Core dependencies
 
 #### Extension API
```
vetleledaal: Heading should only increment by one at a time. With this change (going from `###`  to `##`) it will skip over one heading level.

CONTRIBUTING.md
```diff
@@ -503,15 +742,15 @@ To do this, you need two files:
 </manifest>
 ```
 
-The `AndroidManifest.xml` file will contain an `android:name` attribute that refers to the “path” of your `UrlActivity.kt` file. For example, if the extension is Riztranslation, the `android:name` will be `.id.riztranslation.UrlActivity`.
+The `AndroidManifest.xml` file will contain an `android:name` attribute that refers to the path of your `UrlActivity.kt` file. For example, if the extension is Riztranslation, the `android:name` will be `.id.riztranslation.UrlActivity`.
 
-Next, you have the `<data android:scheme=“https” android:host=“host” android:pathPattern=“/..*” />` element; you can have it multiple times, which allows you to specify the URL that can be opened in Mihon. You can read more about this [here](https://developer.android.com/guide/topics/manifest/data-element). 
+Next, you have the `<data android:scheme="https" android:host="host" android:pathPattern="/..*" />` element; you can have it multiple times, which allows you to specify the URL that can be opened in Mihon. You can read more about this [here](https://developer.android.com/guide/topics/manifest/data-element).
```
vetleledaal: Avoid non-descriptive text such as `here` in links.

```suggestion
Next, you have the `<data android:scheme="https" android:host="host" android:pathPattern="/..*" />` element; you can have it multiple times, which allows you to specify the URL that can be opened in Mihon. You can read more about this in Android's [`<data>` documentation](https://developer.android.com/guide/topics/manifest/data-element).
```

CONTRIBUTING.md
```diff
@@ -264,30 +267,225 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+    fun toSManga() = SManga.create().apply {
+        url = mangaId.toString()
+        thumbnail_url = coverImg
+        this.title = title
+    }
+}
+```
+
+
+**Date parsing — `tryParse`**
```
vetleledaal: Avoid using em-dash.
```suggestion
**Date parsing - `tryParse`**
```

vetleledaal: Avoid faux headings.
```suggestion
###### Date parsing — `tryParse`
```

CONTRIBUTING.md
```diff
@@ -324,7 +522,36 @@ either `SourceFactory` or extend one of the `Source` implementations: `HttpSourc
...
+  * **Decryption:** Use Okio's `cipherSource` extension for stream-based decryption rather than decrypting a full byte array in memory.
+  * Note: `readByteArray()` should generally be avoided here because it forces full in-memory buffering of the image. Streaming directly keeps memory usage lower and more stable.
+  * Always wrap network responses in `response.use { ... }` to ensure the response body is properly closed and to prevent memory leaks.
+  * If applicable, call `bitmap.recycle()` after you're done with it to free native memory early.
+
+### OkHttp and Network
+
+* **GraphQL Queries:** If you are sending GraphQL requests, use Kotlin's raw multi-dollar string interpolation (`$$"""..."""`) for your queries. This prevents having to escape every JSON variable `$` symbol manually.
+
+* **Empty checks on `.text()`:** Because Jsoup's `.text()` automatically trims whitespace, you can use `.isNotEmpty()` instead of `.isNotBlank()` when checking for empty strings.
```
vetleledaal: Mixing `*` and `-` unordered lists. `-` is used elsewhere.

CONTRIBUTING.md
```diff
@@ -787,14 +956,16 @@ For other builds, replace  `app.mihon.dev` with the corresponding package IDs:
...
 
-If the extension builds and runs successfully then the code changes should be ready to test in your local app.
+If the extension builds and runs successfully, then the code changes should be ready to test in your local app.
 
 > [!IMPORTANT]
 > If you're deploying to Android 11 or higher, enable the `Always install with package manager` option in the run configurations. Without this option enabled, you might face issues such as Android Studio running an older version of the extension without the modifications you might have done.
 
 ## Debugging
 
 ### Android Debugger
```
vetleledaal, line 966: Headings should be surrounded by blank lines.

```suggestion
### Android Debugger

```

CONTRIBUTING.md
```diff
@@ -264,30 +267,225 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+#### Available libs
+
+| Module | Description |
+|---|---|
+| [`lib-cookieinterceptor`](https://github.com/keiyoushi/extensions-source/tree/main/lib/cookieinterceptor) | Injects cookies into OkHttp requests for a given domain |
+| [`lib-cryptoaes`](https://github.com/keiyoushi/extensions-source/tree/main/lib/cryptoaes) | AES-CBC decryption compatible with CryptoJS; JSFuck deobfuscation |
+| [`lib-randomua`](https://github.com/keiyoushi/extensions-source/tree/main/lib/randomua) | Fetches and rotates real-world User-Agent strings |
+| [`lib-synchrony`](https://github.com/keiyoushi/extensions-source/tree/main/lib/synchrony) | JavaScript deobfuscation via the Synchrony engine (QuickJS sandbox) |
+| [`lib-textinterceptor`](https://github.com/keiyoushi/extensions-source/tree/main/lib/textinterceptor) | Renders plain text or HTML as a PNG image page |
+| [`lib-unpacker`](https://github.com/keiyoushi/extensions-source/tree/main/lib/unpacker) | Unpacks Dean Edwards–packed JavaScript; substring extraction helpers |
```
vetleledaal: Avoid en-dash.
```suggestion
| [`lib-unpacker`](https://github.com/keiyoushi/extensions-source/tree/main/lib/unpacker) | Unpacks Dean Edwards-packed JavaScript; substring extraction helpers |
```

CONTRIBUTING.md
```diff
@@ -264,30 +267,225 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+// From a String:
+val dto = jsonString.parseAs<MyDto>()
+
+// With a transform applied before parsing:
+val dto = response.parseAs<MyDto> { it.substringAfter("callback(").dropLast(1) }
+```
+
+**Do not** create a local `private val json: Json by injectLazy()` unless you specifically need a custom JSON configuration (e.g., `isLenient = true` or custom serializers). For standard parsing, the global instance is already available via `jsonInstance` and the `parseAs` helpers use it automatically.
+
+**JSON serialization — `toJsonString`**
```
vetleledaal: Avoid em-dash.
```suggestion
**JSON serialization - `toJsonString`**
```

CONTRIBUTING.md
```diff
@@ -264,30 +267,225 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+
+**Date parsing — `tryParse`**
+
+Use `keiyoushi.utils.tryParse` on a `SimpleDateFormat` instance to safely parse a date string.
+It returns `0L` on failure or when the input is `null`, which is exactly what the app expects.
+
+```kotlin
+import keiyoushi.utils.tryParse
+
+// Declare dateFormat at class/file level — creating SimpleDateFormat is expensive:
```
vetleledaal: Avoid em-dash.
```suggestion
// Declare dateFormat at class/file level - creating SimpleDateFormat is expensive:
```

CONTRIBUTING.md
```diff
@@ -264,30 +267,225 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+private val dateFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.ROOT).apply {
+    timeZone = TimeZone.getTimeZone("UTC")
+}
+
+chapter.date_upload = dateFormat.tryParse(dateStr)
+```
+
+**Do not** write manual try/catch blocks or null-guards around `SimpleDateFormat.parse()` —
```
vetleledaal: Avoid em-dash.
```suggestion
**Do not** write manual try/catch blocks or null-guards around `SimpleDateFormat.parse()` -
```

CONTRIBUTING.md
```diff
@@ -264,30 +267,225 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+`tryParse` handles both. Also, always declare your `SimpleDateFormat` as a class-level or
+file-level `val` so it is not reconstructed for every chapter.
+
+**Filter helpers — `firstInstance` / `firstInstanceOrNull`**
```
vetleledaal: Avoid em-dash.
```suggestion
**Filter helpers - `firstInstance` / `firstInstanceOrNull`**
```

CONTRIBUTING.md
```diff
@@ -264,30 +267,225 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+Use these instead of `filterIsInstance<T>().first()` / `filterIsInstance<T>().firstOrNull()`.
+
+```kotlin
+import keiyoushi.utils.firstInstance
+import keiyoushi.utils.firstInstanceOrNull
+
+val genreFilter = filters.firstInstanceOrNull<GenreFilter>()
+```
+
+**SharedPreferences — `getPreferences` / `getPreferencesLazy`**
```
vetleledaal: Avoid em-dash.
```suggestion
**SharedPreferences - `getPreferences` / `getPreferencesLazy`**
```

CONTRIBUTING.md
```diff
@@ -264,30 +267,225 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+import keiyoushi.utils.getPreferencesLazy
+
+// Eager:
+private val preferences = getPreferences()
+
+// Lazy (recommended for most cases):
+private val preferences by getPreferencesLazy()
+```
+
+**Next.js data extraction — `extractNextJs` / `extractNextJsRsc`**
```
vetleledaal: Avoid em-dash.
```suggestion
**Next.js data extraction - `extractNextJs` / `extractNextJsRsc`**
```

CONTRIBUTING.md
```diff
@@ -264,30 +267,225 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+    element is JsonObject && "slug" in element
 }
 ```
 
+For client-side navigation responses (`text/x-component` content type), pass the `rsc: 1`
+request header and use `extractNextJsRsc` on the response body string.
+See [#14266](https://github.com/keiyoushi/extensions-source/pull/14266) and
+[#14446](https://github.com/keiyoushi/extensions-source/pull/14446) for real-world usage.
+
+**Extracting URLs — `setUrlWithoutDomain` + `absUrl`**
```
vetleledaal: Avoid em-dash.
```suggestion
**Extracting URLs - `setUrlWithoutDomain` + `absUrl`**
```

CONTRIBUTING.md
```diff
@@ -443,27 +670,39 @@ will be cached.
...
+- **Cache Regex instances:** Define `Regex` instances at the class level or in a `companion object` so they aren't recompiled on every method call.
+- **Do not hardcode `User-Agent`:** Unless absolutely necessary (e.g., to bypass Cloudflare/protection, or to retrieve a specific mobile layout/different selectors), do not hardcode a specific `User-Agent`. Calling `super.headersBuilder()` already provides the app's default User-Agent.
+- **Use `buildString { }`:** When building descriptions or dynamic strings, use Kotlin's `buildString { ... }` instead of manually instantiating a `StringBuilder()`.
+- **Media Types:** `application/json` is intrinsically UTF-8. Avoid using `application/json; charset=utf-8`. Use `"application/json".toMediaType()`.
+- **Use `getUrlWithoutDomain` carefully:** It can be useful when parsing target source URLs, but note a current issue with spaces—replace them with URL-encoded characters (e.g., `%20`).
```
vetleledaal: Avoid em-dash.
```suggestion
- **Use `getUrlWithoutDomain` carefully:** It can be useful when parsing target source URLs, but note a current issue with spaces-replace them with URL-encoded characters (e.g., `%20`).
```

CONTRIBUTING.md
```diff
@@ -264,30 +267,225 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+Place your code in the package `keiyoushi.lib.<mylibname>`. Document public API with KDoc so
+contributors can understand the lib without needing to read `CONTRIBUTING.md`.
+
+#### keiyoushi.utils (core utilities)
+
+The `core/utils` module provides a set of shared extension functions that are available to all extensions
+without any extra Gradle dependency. Prefer using these helpers instead of implementing your own equivalents, as they provide standardized and maintained solutions.
+The utilities live in the `keiyoushi.utils` package and are imported individually.
+
+**JSON parsing - `parseAs`**
```
vetleledaal: Avoid faux headings.
```suggestion
##### JSON parsing - `parseAs`
```

CONTRIBUTING.md
vetleledaal: Avoid faux headings.
```suggestion
##### JSON serialization — `toJsonString`
```

CONTRIBUTING.md
vetleledaal: Avoid faux headings.
```suggestion
###### Filter helpers — `firstInstance` / `firstInstanceOrNull`
```

CONTRIBUTING.md
vetleledaal: Avoid faux headings.
```suggestion
###### Next.js data extraction — `extractNextJs` / `extractNextJsRsc`
```

CONTRIBUTING.md
vetleledaal: Avoid faux headings.
```suggestion
###### Extracting URLs — `setUrlWithoutDomain` + `absUrl`
```

vetleledaal (commented):
Comments mostly from lint issues raised by `markdownlint`, filtered to changed lines.

CONTRIBUTING.md
```diff
@@ -264,30 +267,226 @@ some interfaces and stubs from the [app](https://github.com/mihonapp/mihon) for
...
+) {
+    fun toSManga() = SManga.create().apply {
+        url = mangaId.toString()
+        thumbnail_url = coverImg
+        this.title = title
+    }
+}
+```
+
+
```
vetleledaal:
```suggestion
```

vetleledaal (dismissed):

vetleledaal (approved):

PR #14890, opened 2026-04-19
VortexScans (EN): Switch to HTML scraping

### LINKED ISSUE #14841: Vortex Scans: `HTTP 403` error
### Source information

Vortex Scans 1.4.61

### Source language

English

### Steps to reproduce

1. Open the extension
2. Open an entry
3. Update it

### Expected behavior

It should update it normally

### Actual behavior

Instead it’s like this and on webview it works perfectly

Image

Image

### Mihon/Mihon fork version

Tachimanga v4.14(437)

### Android version

IOS 26.4.1

### Other details

I’m on IOS so the error may not be the same on Android

### Acknowledgements

### 

---

Ploosh:
Closes #14841

## What's happening

The backend API endpoints `/api/post` (manga details + chapter list) and `/api/chapter` (page images) started returning `HTTP 403 Forbidden` for anonymous callers. `/api/query` (used for popular / latest / search) is unaffected.

Reproduced with curl:

```
GET https://api.vortexscans.org/api/post?postId=547       → 403 {"message":"Forbidden"}
GET https://api.vortexscans.org/api/chapter?chapterId=... → 403 {"message":"Forbidden"}
GET https://api.vortexscans.org/api/query?page=1&...       → 200 ✓
```

Response headers expose `X-Better-Auth-Session` and the CORS preflight advertises `X-Captcha-Response` in `access-control-allow-headers`. `/api/auth/get-session` returns `null` and sets no cookie, and no anonymous/guest sign-in endpoint exists. Tried many header/cookie/method permutations and alternate paths — the lockdown is deterministic and auth-required.

## What this PR does

Switches `mangaDetailsRequest`, `chapterListRequest`, and `pageListRequest` to scrape the SSR HTML instead. The Astro frontend embeds everything we need in the initial HTML:

- `/series/{slug}` — manga details and the **full** chapter list (with id, slug, number, createdAt, `isLocked`, `isAccessible`) inside a JSON blob on an `<astro-island props="...">` tag.
- `/series/{slug}/{chapter-slug}` — all page images as standard `` tags.

Astro's prop serialization wraps values as `[0, value]` / `[1, [items]]`, so the mapper uses a string-aware bracket-matching scanner (with targeted regexes for leaf fields) rather than full JSON parsing — safer than replacing `[0, ` blindly across the blob.

`/api/query` is kept for popular / latest / search. `PostResponseDto`, `PostDto`, `PostChapterDto`, `ChapterResponseDto`, `ChapterDto`, `ChapterImageDto`, `API_POST_PATH`, `API_CHAPTER_PATH`, `resolvePostId`, `resolveChapterId`, `postRequest`, `chapterRequest` are removed since nothing calls them anymore.

## Tested

Built with `./gradlew :src:en:arvenscans:assembleDebug` (no warnings) and installed into Mihon v0.19.9 on an Android 14 (arm64) emulator. Verified:

- Popular / Latest / Search load manga
- Opening a manga loads details + chapter list (previously 403)
- Opening an unlocked chapter loads all pages in order (previously 403)
- Locked chapters are filtered out by default; enabling "Show inaccessible chapters" surfaces them with the 🔒 prefix

Also spot-checked the parsing against two series (`solo-ex-rank-summoner` with 21 chapters, `chrono-labyrinth` with 13 chapters) to confirm counts, lock flags, and page ordering.

## Checklist

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions (61 → 62)


Ploosh:
Update: switched chapter list back to the API, thanks to @ArseneLapin on Discord for pointing out that `GET /api/chapters?postId={id}&take=999&skip=0` still works anonymously. I missed it yesterday because I was only testing with a `chapterId` param.

- **Chapter list** — now uses `/api/chapters` (clean JSON with `id`, `slug`, `number`, `title`, `createdAt`, `isLocked`, `isAccessible`, and `mangaPost.slug` for URL building).
- **Manga details** — HTML scraping (still needed for description / alt titles).
- **Page images** — HTML scraping (no public API; `/api/chapter` is 403).

Bumped `extVersionCode` to 63. Verified on the emulator that the chapter list loads via the new endpoint and page images still render.

src/en/arvenscans/src/eu/kanade/tachiyomi/extension/en/arvenscans/Mapper.kt
```diff
@@ -65,8 +52,150 @@ fun PostChapterDto.toSChapterModel(mangaSlug: String, dateFormat: SimpleDateForm
     }
 }
 
-private fun buildDescription(post: PostDto): String? {
-    val synopsis = post.postContent
+fun Document.toSMangaDetailsModel(): SManga {
```
AwkwardPeak7: can the astro prop be cleanly extracted using [this](https://github.com/keiyoushi/extensions-source/blob/d4a611ef42fc4c96f875073bc6ce007bab4e2e5b/src/en/asurascans/src/eu/kanade/tachiyomi/extension/en/asurascans/AsuraScans.kt#L309-L328)?

src/en/arvenscans/src/eu/kanade/tachiyomi/extension/en/arvenscans/VortexScans.kt
```diff
@@ -121,11 +122,11 @@ class VortexScans :
         return "$baseUrl/$SERIES_PATH_SEGMENT/$slug"
     }
 
-    override fun mangaDetailsRequest(manga: SManga): Request = postRequest(resolvePostId(manga))
+    override fun mangaDetailsRequest(manga: SManga): Request = seriesPageRequest(extractMangaSlug(manga.url))
 
     override fun mangaDetailsParse(response: Response): SManga {
-        val payload = response.parseAs<PostResponseDto>()
-        return payload.post.toSMangaDetailsModel()
+        val document = Jsoup.parse(response.body.string(), response.request.url.toString())
```
AwkwardPeak7:
```suggestion
        val document = response.asJsoup()
```

Eradrim:
Weard from what I see this wasn't merged yet but old extension started working today 
Screenshot_20260420_182604_Mihon

Ploosh:
Endpoints are back up. Retested `/api/post` and `/api/chapter` anonymously and they're returning 200 again, and other users on #14841 confirm the old extension started working around 2026-04-20 without any change.

Going to hold off on merging this for now. With the API healthy, the primary benefit (unblocking users) is gone, and committing to a mixed API+HTML approach means maintaining ~150 LOC of Astro prop-parsing through every frontend redeploy for no current user benefit.

Leaving the branch in place as an escape hatch. The `X-Better-Auth-Session` header and `X-Captcha-Response` in CORS preflight suggest backend auth gating is still being actively developed, so if the 403s return I'll pick this back up, possibly reshaped as an API-first-with-scraping-fallback rather than a full swap.

src/en/arvenscans/src/eu/kanade/tachiyomi/extension/en/arvenscans/Mapper.kt
Ploosh: Done in e7957878. Replaced the whole manual parser (findMatchingBracket / extractJsonStringField / extractJsonIntField / decodeJsonString / extractGenreNames) with the `extractAstroProp` + `unwrapAstro` pattern. Much cleaner, thanks for the pointer.

src/en/arvenscans/src/eu/kanade/tachiyomi/extension/en/arvenscans/VortexScans.kt
Ploosh: Applied in e7957878, both here and in `pageListParse`.

AwkwardPeak7:
leaving this open for now as the extension seems to work?

PR #14899, opened 2026-04-19
BiliManga: add deep links

hualiong:
Checklist:


src/zh/bilimanga/src/eu/kanade/tachiyomi/extension/zh/bilimanga/BiliManga.kt
```diff
@@ -30,7 +30,7 @@ class BiliManga :
 
     override val lang = "zh"
 
-    override val name = "Bilimanga.net"
+    override val name = "嗶哩漫畫"
```
stevenyomi: This name was already rejected in the first PR.

stevenyomi (changes requested):

src/zh/bilimanga/src/eu/kanade/tachiyomi/extension/zh/bilimanga/UrlActivity.java
```diff
@@ -0,0 +1,33 @@
+package eu.kanade.tachiyomi.extension.zh.bilimanga;
+
+import android.app.Activity;
+import android.content.Intent;
+import android.os.Bundle;
+import android.util.Log;
+
+/**
+ * Introducing Kotlin-related code (kotlin-stdlib) may cause ClassNotFoundError on some custom ROMs,
```
AwkwardPeak7: you have to avoid using the kotlin stdlib while writing the kotlin code, Ik it is counter intuitive

reference: https://github.com/keiyoushi/extensions-source/blob/main/src/en/reimanga/src/eu/kanade/tachiyomi/extension/en/reimanga/UrlActivity.kt

also consider passing the whole url as is through the activity, and do the parsing in the source class, this way users get the ability to paste raw urls in the app

AwkwardPeak7 (approved):

PR #14969, opened 2026-04-21
utils: Add Protobuf helper

manti-X:


core/src/main/kotlin/keiyoushi/utils/Protobuf.kt
```diff
@@ -0,0 +1,68 @@
...
+ * @param transform A function to transform the raw [ByteArray] before it is decoded.
+ */
+inline fun <reified T> Response.parseAsProto(transform: (ByteArray) -> ByteArray): T = use { transform(it.body.bytes()).decodeProto() }
+
+/**
+ * Parses this [ResponseBody] into an object of type [T] using Protobuf deserialization.
+ *
+ * The body is automatically closed after reading.
+ */
+inline fun <reified T> ResponseBody.parseAsProto(): T = use { protoInstance.decodeFromByteArray(it.bytes()) }
```
Secozzi: No need for `use {}`, calling `.bytes()` already closes the response body https://square.github.io/okhttp/5.x/okhttp/okhttp3/-response-body/index.html

Ploosh (commented):
Nice addition — mirrors the `Json.kt` utilities well. Two small points on patterns/conventions.

core/src/main/kotlin/keiyoushi/utils/Protobuf.kt
```diff
@@ -0,0 +1,83 @@
...
+import kotlinx.serialization.protobuf.ProtoBuf
+import okhttp3.MediaType.Companion.toMediaType
+import okhttp3.RequestBody
+import okhttp3.RequestBody.Companion.toRequestBody
+import okhttp3.Response
+import okhttp3.ResponseBody
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
+
+val protoInstance: ProtoBuf = Injekt.get()
```
Ploosh, line 16: Would it be worth matching `Json.kt:11`'s pattern here?

```kotlin
val protoInstance: ProtoBuf by injectLazy()
```

That version is lazy (resolved on first use rather than at class-load time), matches the convention the rest of `keiyoushi.utils` follows, and avoids any eager-init / circular-dep edge cases with Injekt.

core/src/main/kotlin/keiyoushi/utils/Protobuf.kt
```diff
@@ -0,0 +1,83 @@
...
+val PROTOBUF_MEDIA_TYPE = "application/protobuf".toMediaType()
```
Ploosh, line 17: Small note on media-type choice: `application/x-protobuf` is still the variant many older servers expect, while `application/protobuf` is the newer form. If a specific source only accepts the `x-` variant, a consumer would need to override — so consider either exposing this as an `open val` at the source level, or a short kdoc noting the choice and that it may need overriding. Not a blocker.

core/src/main/kotlin/keiyoushi/utils/Protobuf.kt
manti-X: basically this: https://github.com/keiyoushi/extensions-source/pull/10039/changes#r2260579561
`injectLazy()` just adds a lazy without solving a problem. unless there’s an actual init order issue, i dom’t see a reason to make it lazy

core/src/main/kotlin/keiyoushi/utils/Protobuf.kt
```diff
@@ -0,0 +1,85 @@
...
+inline fun <reified T : Any> T.encodeProto(proto: ProtoBuf = protoInstance): ByteArray = proto.encodeToByteArray(this)
+
+/**
+ * Parses the response body into an object of type [T] using Protobuf deserialization.
+ *
+ * The response is automatically closed after reading.
+ *
+ * @param proto The [ProtoBuf] instance to use for deserialization.
+ */
+inline fun <reified T> Response.parseAsProto(proto: ProtoBuf = protoInstance): T = use { it.body.bytes().decodeProto(proto) }
```
AwkwardPeak7: use the stream variant to consume the bytes

manti-X: actually there is no difference, both `bytes()` and `byteStream().readBytes()` read the entire body into a `ByteArray`. or what did you imagine?

AwkwardPeak7: There should be a `ProtoBuf.decodeFromStream` or am I miss remembering

manti-X: this is not json xD
it's an open issue tho at [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization/issues?q=is%3Aissue%20state%3Aopen%20OutputStream%2FInputStream)

AwkwardPeak7: 😔

AwkwardPeak7:
please revert `byteStream().readBytes()` back to `bytes()`

AwkwardPeak7 (approved):

PR #14993, opened 2026-04-22
Kmanga fix chapter name parsing

### LINKED ISSUE #14302: K Manga: Split chapters use wrong chapter numbers
### Source information

K Manga 1.4.3

### Source language

English

### Steps to reproduce

1. Install the K Manga extension (version 1.4.3)
2. Add a manga that contains split chapters, such as:
   - `I'm a Curse Crafter, and I Don't Need an S-Rank Party!`
   - `The Remarkable Revenge of a Vindictive Villainess`
3. Observe the chapter list. For example, the manga `I'm a Curse Crafter...` shows chapters like:
   ```
   CHAPTER 56(1) That Hero, This Curse Crafter
   CHAPTER 55(3) Locking Swords
   CHAPTER 55(2) Locking Swords
   CHAPTER 55(1) Locking Swords
   ...
   ```
   but the assigned chapter numbers are not the expected `55.3`, `55.2`, `55.1`, etc. after reading (wrong number in history and tracking)

### Expected behavior

Chapters that have a part suffix (e.g., `(1)`, `(2)`, `(3)`) should be numbered as `X.Y` (where X is the main chapter number and Y is the part). For example, `CHAPTER 55(3)` should be `55.3`.

### Actual behavior

The extension currently uses the `index` field from the API as the chapter number, which corresponds to the order in the episode list rather than the actual chapter numbering. This leads to arbitrary numbers like `133` for `CHAPTER 55(3)` instead of the expected `55.3`. This occurs because in `Dto.kt` the `toSChapter` function sets `chapter_number = index.toFloat()`.

### Mihon/Mihon fork version

TachiyomiSY 1.12.0

### Android version

Android 16

### Other details

The issue appears to be in the mapping of episode data to `SChapter`. The `Episode` class has an `index` property which is used as the chapter number. However, for split chapters, the actual chapter number is embedded in the `episode_name`. The current code:

```kotlin
// Dto.kt
fun toSChapter(dateFormat: SimpleDateFormat): SChapter = SChapter.create().apply {
    url = "/title/$titleId/episode/$episodeId"
    val lock = if (isLocked) "🔒 " else ""
    name = lock + episodeName
    chapter_number = index.toFloat()   // <-- problem
    date_upload = dateFormat.tryParse(startTime)
}
```

The chapter number should be parsed from `episodeName` (e.g., extract `55` and `(3)` to form `55.3`) and fallback to index if that fails.

### Acknowledgements

### 

---

darwin-256:
Closes #14302

Checklist:

### DIFF
--- a/src/en/kmanga/src/eu/kanade/tachiyomi/extension/en/kmanga/Dto.kt
+++ b/src/en/kmanga/src/eu/kanade/tachiyomi/extension/en/kmanga/Dto.kt
@@ -7,9 +7,10 @@ import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import java.text.SimpleDateFormat
 
-private val chapterNameRegex = Regex("""(?i)(?:chapter|ch|episode|ep|第)\.?\s*(\d+(?:\.\d+)?)(?:\s*[(（](\d+)[)）])?""")
+private val chapterNameRegex = Regex("""(?i)(?:chapter|ch|episode|ep|第).?\s*(\d+(?:\.\d+)?)(?:\s*[(（](\d+)[)）])?""")
 private val splitChapterRegex = Regex("""(\d+(?:\.\d+)?)\s*[(（](\d+)[)）]""")
 private val fallbackChapterRegex = Regex("""^(\d+(?:\.\d+)?)(?:\s*[(（](\d+)[)）])?""")
+private val partSuffixRegex = Regex("""\s*[(（]\d+[)）]""")
 
 // Popular
 @Serializable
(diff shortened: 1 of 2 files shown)


src/en/kmanga/src/eu/kanade/tachiyomi/extension/en/kmanga/Dto.kt
```diff
@@ -92,8 +96,34 @@ class Episode(
...
+            ?: splitChapterRegex.find(episodeName)
+            ?: fallbackChapterRegex.find(episodeName)
+
+        chapter_number = if (match != null) {
+            val main = match.groupValues[1]
+            val part = match.groupValues[2]
+
+            if (part.isNotEmpty()) {
+                // Modify the display name visually so "CHAPTER 55(3)" becomes "CHAPTER 55.3"
+                val replacement = match.value.replaceFirst(Regex("""\s*[(（]$part[)）]"""), ".$part")
```
MuhamadSyabitHidayattulloh: Seperate this regex to

src/en/kmanga/src/eu/kanade/tachiyomi/extension/en/kmanga/Dto.kt
```diff
@@ -92,8 +96,34 @@ class Episode(
...
+                parsedName = parsedName.replaceRange(match.range, replacement)
+
```
MuhamadSyabitHidayattulloh: Source have "10.5(2)"?

AwkwardPeak7 (approved):

PR #15074, opened 2026-04-25
add HentaiLoop

### LINKED ISSUE #13755: Add Hentailoop source
### Source name

Hentailoop

### Source link

https://hentailoop.com/

### Source language

English

### Other details

18+/NSFW = yes

### Acknowledgements

- [x] I have checked that the extension does not already exist by searching the [extension listing](https://keiyoushi.github.io/extensions/) and verified it does not appear there.
- [x] I have written a meaningful title with the source name.

### 

---

AwkwardPeak7:
closes #14999
Closes #13755

depends on #15072

Checklist:

There are other multiple languages on the site, but majority of titles are in english, so decided to not add source factory
<details>
<summary>Counts from the site</summary>
image
</details>


src/all/hentailoop/build.gradle
darwin-256, line 1: ### 1. `build.gradle` needs to be migrated to `build.gradle.kts`
The PR still uses the legacy `.gradle` format and the legacy plugin (`kei.plugins.extension.legacy`). To use the new standard, you must rename the file to `build.gradle.kts`, use the modern plugin alias, and declare your `source {}` and `deeplink {}` blocks.

Your `build.gradle.kts` should look like this:
```kotlin
plugins {
    alias(kei.plugins.extension)
}

keiyoushi {
    name = "HentaiLoop"
    versionCode = 1
    contentWarning = ContentWarning.NSFW
    libVersion = "1.4"

    source {
        lang = "all"
        baseUrl = "https://hentailoop.com"
    }

    deeplink {
        host("hentailoop.com")
        path("/manga/..*")
    }
}
```

src/all/hentailoop/AndroidManifest.xml
darwin-256, line 1: ### 2. Delete `UrlActivity.kt` and `AndroidManifest.xml`

As noted in the URL intent filter section of `CONTRIBUTING.md`, manual
`AndroidManifest.xml` and `UrlActivity.kt` files are deprecated and no longer
needed. The build system automatically generates these for you based on the
`deeplink {}` block in `build.gradle.kts`. Please delete both of these files from
the PR.

src/all/hentailoop/src/eu/kanade/tachiyomi/extension/all/hentailoop/HentaiLoop.kt
darwin-256, line 1: ### 3. Main Class (`HentaiLoop.kt`) Annotations and Overrides

Because we are switching to the generated DSL, KSP automatically injects the
extension properties into your class.

  - Add the `@Source` annotation to the `HentaiLoop` class.
  - Remove the manual overrides for `name`, `lang`, and `baseUrl`.
  - Dynamic `baseUrl` conflict: Your code currently overrides `baseUrl` dynamically
    using the `captchaRequired` variable. With the new KSP approach, `baseUrl` is
    automatically injected as a static string based on what you put in
    `build.gradle.kts`, so attempting to override it will cause a compiler error.
    You should remove the `captchaRequired` logic that changes the `baseUrl`. If the
    user hits a captcha on search, throwing the Exception `"Captcha Required!
    Open WebView..."` is sufficient. The app will open the WebView to the main
    baseUrl by default where they can solve the challenge.
```kotlin
import keiyoushi.annotation.Source

@Source
class HentaiLoop : HttpSource() {
    // name, lang, and baseUrl are handled by KSP. Remove them here.
    ...
}
```
### 4. Hardcoded host in Deep Link check

In `deepLink(url: String)` you have:
```kotlin
if (httpUrl.host == DOMAIN && ...)
```
`CONTRIBUTING.md` explicitly states: Avoid checking for hardcoded host strings
(e.g., `queryUrl.host == "site.com"`). Prefer dynamically checking against the
source's baseUrl.

Please update this to check against `baseUrl.toHttpUrl().host` so that the
extension won't break if mirrors or a new domain are configured in the future.

### 5. Unnecessary `HttpUrl.Builder` usage

In `getMangaUrl()` and `getChapterUrl()`, you are using `HttpUrl.Builder` for simple
path combinations:
```kotlin
val url = HttpUrl.Builder().apply {
    scheme("https")
    host(DOMAIN)
    addPathSegment("manga")
    addPathSegment(manga.url)
    addPathSegment("")
}.build().toString()
```
`CONTRIBUTING.md` notes: Static URLs don't need `HttpUrl.Builder`: Use string
interpolation directly for URLs with no dynamic query parameters.

You can simplify these methods dramatically:
```kotlin
override fun getMangaUrl(manga: SManga): String = "$baseUrl/manga/${manga.url}/"
```

PR #15097, opened 2026-04-26
HoneyManga: added preferences to hide Novels and Mangas with chosen tags

### LINKED ISSUE #15024: HoneyManga: Option to/Hide novels
### Source name

HoneyManga

### Source language

Ukrainian

I'd like the source settings to include an option to hide novels in types (uk: Новели) (since the app can't open them) and yaoi in genres (uk: Яой / Шьонен-ай) (since I really don't like seeing it in my feed).

### Other details

_No response_

### Acknowledgements

### 

---

Exinaus:
Checklist:
Closes #15024 

### DIFF
--- a/src/uk/honeymanga/src/eu/kanade/tachiyomi/extension/uk/honeymanga/HoneyManga.kt
+++ b/src/uk/honeymanga/src/eu/kanade/tachiyomi/extension/uk/honeymanga/HoneyManga.kt
@@ -1,5 +1,7 @@
 package eu.kanade.tachiyomi.extension.uk.honeymanga
 
+import androidx.preference.MultiSelectListPreference
+import androidx.preference.PreferenceScreen
 import eu.kanade.tachiyomi.extension.uk.honeymanga.dtos.CompleteHoneyMangaDto
 import eu.kanade.tachiyomi.extension.uk.honeymanga.dtos.HoneyMangaChapterPagesDto
 import eu.kanade.tachiyomi.extension.uk.honeymanga.dtos.HoneyMangaChapterResponseDto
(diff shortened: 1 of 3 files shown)


src/uk/honeymanga/src/eu/kanade/tachiyomi/extension/uk/honeymanga/HoneyManga.kt
```diff
@@ -161,6 +197,14 @@ class HoneyManga : HttpSource() {
         mangaList.size == DEFAULT_PAGE_SIZE,
     )
 
+    private fun makeMangasPageSearch(mangaList: List<HoneyMangaDto>): MangasPage = MangasPage(
+        mangaList
+            .filter { !blockTypes().contains(it.type) }
+            .filter { blockGenres().intersect((it.genres ?: emptyList()).toSet()).isEmpty() }
+            .map(::makeSManga),
```
AwkwardPeak7, line 204: app will not go to next page if current pages ends up being empty due to client side filtering

Exinaus: That function used only while searching manga. For popular/latest makeMangasPage() are used. I specifically made copy of makeMangasPage, renamed it to makeMangasPageSearch to add filters to search only. And applied to a function that parses results from the search.
Popular/Latest tab are working fine.

I can see the issue if search result returns more than maximum items per page, but how likely is that?
Maybe it should be a user job to narrow the search it it returns that many results?

AwkwardPeak7: so text search usually doesn't have more than one page?

Exinaus: That's the search from Popular/Latest screen


That's the Global search


Maybe, some manga with dozens of spin-offs that can handle it, but i just didn't think of it

That's probably a 5th page of a Popular tab, still loading ... i was able to scroll to it without any issue

AwkwardPeak7: popular/latest won't be an issue since you send the filter to the source directly. problem may be with search. but ig it is good enough ¯\\\_(ツ)\_\/¯

AwkwardPeak7 (approved):

AinsOoalGon86:
You're my hero. Could you make the same filter for Zenko and MangaInUA(when will it be fixed)?

PR #15270, opened 2026-04-29
MimiHentai [VI]: Support related

cuong-tran:
Checklist:


src/vi/mimihentai/src/eu/kanade/tachiyomi/extension/vi/mimihentai/MiMiHentai.kt
```diff
@@ -196,7 +197,6 @@ class MiMiHentai : HttpSource() {
     // =============================== Related ================================
     // disable suggested mangas on Komikku due to heavy rate limit
     override val disableRelatedMangasBySearch = true
-    override val supportsRelatedMangas = false
```
AwkwardPeak7, line 199: the actual methods not overridden?

cuong-tran: App defaults to call `popularMangaParse()` on a response returned from `getMangaDetails()` and that works because they use same selector and element.

AwkwardPeak7 (approved):

PR #15323, opened 2026-04-30
HDoujin: fix language code

Hiirbaf:
Checklist:


src/all/hdoujin/src/eu/kanade/tachiyomi/extension/all/hdoujin/HDoujinFactory.kt
AwkwardPeak7, line 1: you would need to override the id for that particular one so old users don't have to migrate. 

get current id for kr from the repo index file and override `id` in class

AwkwardPeak7 (approved):

PR #15620, opened 2026-05-07
refactor UrlActivities

AwkwardPeak7:
- Moved url parsing / slug extraction logic to `fetchSearchMang`
- Removed all parsing from Activities
- bonus: allows users to paste manga url in browse
- Ai assisted

Doing this now to eventually have the urlactivity and manifest deeplinks be generated via gradle

Don't know if I should bump all of these sources together...

Checklist:

### DIFF
--- a/lib-multisrc/comicaso/build.gradle.kts
+++ b/lib-multisrc/comicaso/build.gradle.kts
@@ -2,4 +2,4 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 5
+baseVersionCode = 6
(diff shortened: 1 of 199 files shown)


FourTOne5 (changes requested):
People can complain if it doesn't work. I'm here for the gradle configuration for extensions

FourTOne5 (approved):

vetleledaal (approved):
Might as well bump the version, so existing users will update to a post-Proguard version. It's a change in behavior either way.

PR #15640, opened 2026-05-07
utils: add crypto utils

manti-X:
Checklist:

### DIFF
--- a/core/src/main/kotlin/keiyoushi/utils/Crypto.kt
+++ b/core/src/main/kotlin/keiyoushi/utils/Crypto.kt
@@ -1,0 +1,29 @@
+package keiyoushi.utils
+
+/**
+ * Decodes this hexadecimal string into a [ByteArray].
+ *
+ * The input string must contain only valid hexadecimal characters (`0-9`, `a-f`, `A-F`)
+ * and must have an even length.
+ *
+ * @return the decoded bytes represented by this hex string
+ * @throws IllegalArgumentException if the string has an odd length or contains invalid hex characters
+ */
+fun String.decodeHex(): ByteArray {
+    require(length % 2 == 0) { "Unexpected hex string: $this" }
+
+    val result = ByteArray(length / 2)
+    for (i in result.indices) {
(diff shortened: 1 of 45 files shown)


core/src/main/kotlin/keiyoushi/utils/Crypto.kt
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Decodes this hexadecimal string into a [ByteArray].
+ *
+ * The input string must contain only valid hexadecimal characters (`0-9`, `a-f`, `A-F`)
+ * and must have an even length.
+ *
+ * @return the decoded bytes represented by this hex string
+ * @throws IllegalArgumentException if the string has an odd length or contains invalid hex characters
+ */
+fun String.decodeHex(): ByteArray {
```
manti-X, line 12: This is taken from the okio [`ByteStringNonJs.kt`](https://github.com/square/okio/blob/master/okio/src/nonJsMain/kotlin/okio/internal/ByteStringNonJs.kt) with the only difference being that it returns a `ByteArray` instead of a `ByteString`, avoiding an extra conversion.

It's also faster and creates less overhead than the previous `decodeHex`.

[`kotlin.text.hexToByteArray()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/hex-to-byte-array.html) (since kotlin 2.2) can also be used when a custom `HexFormat` is needed. While it's slower than the new implementation, it still performs better than the previous `decodeHex`.

https://pl.kotl.in/HOAq1Et1s

lib-multisrc/madara/build.gradle.kts
```diff
@@ -2,7 +2,7 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 49
+baseVersionCode = 50
```
AwkwardPeak7, line 5: lets not bump madara 3rd time in the last 24 hours

darwin-256:
comix changed the hash #15644 check if this still applies

AwkwardPeak7 (approved):

PR #15657, opened 2026-05-08
Rework Gradle build logic

FourTOne5:
Checklist:


gradle/build-logic/src/main/kotlin/PluginExtensionLegacy.kt
```diff
@@ -0,0 +1,188 @@
...
+    override fun apply(target: Project): Unit = with(target) {
+        plugins {
+            alias(libs.plugins.android.application)
+            alias(libs.plugins.kotlin.serialization)
+            alias(kei.plugins.spotless)
+        }
+
+        configureKotlin()
+
+        assert(!extra.has("pkgNameSuffix"))
```
AwkwardPeak7: these `assert` statements may be no-op. Need to pass `-ea` flag to `java` for asserts, which the gradle process doesn't do.

https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/assert.html
https://togithub.com/gradle/gradle/issues/29103

gradle/build-logic/src/main/kotlin/PluginSpotless.kt
```diff
@@ -0,0 +1,82 @@
...
+import keiyoushi.gradle.extensions.alias
+import keiyoushi.gradle.extensions.libs
+import keiyoushi.gradle.extensions.plugins
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.configure
+import java.io.Serializable
+
+@Suppress("UNUSED")
+class PluginSpotless : Plugin<Project> {
```
AwkwardPeak7, line 14: I may be missing something, but this isn't hooked into build? `spotlessCheck` or `spotlessApply` was dependency of prebuild previously.

AwkwardPeak7:
also update the regex in `generate-build-matrices.py` to include any new directories in the core files regex
Also include them in sparse checkout section of contributing guide

AwkwardPeak7:
nitpick: use `keiyoushi` instead of `kei`

FourTOne5:
> nitpick: use `keiyoushi` instead of `kei`

it's a bit hard to use `keiyoushi` for version catalog name when the package inside `build-logic` is `keiyoushi.gradle.` due to resolution issues

gradle/build-logic/gradle.properties
```diff
@@ -0,0 +1,13 @@
+# Configuration should be synced with [/gradle.properties]
+#
+# Gradle properties are not passed to included builds 
+# https://github.com/gradle/gradle/issues/2534
+org.gradle.caching=true
+org.gradle.configuration-cache=true
+org.gradle.configureondemand=true
+org.gradle.jvmargs=-Xmx6144m
+org.gradle.parallel=true
+org.gradle.workers.max=5
```
AwkwardPeak7: do we need max workers here? it was removed from the one in project root

gradle/build-logic/src/main/kotlin/PluginExtensionLegacy.kt
```diff
@@ -0,0 +1,179 @@
...
+import org.gradle.api.Project
+import org.gradle.api.plugins.BasePluginExtension
+import org.gradle.api.plugins.ExtraPropertiesExtension
+import org.gradle.kotlin.dsl.configure
+import org.gradle.kotlin.dsl.dependencies
+import org.gradle.kotlin.dsl.extra
+import org.gradle.kotlin.dsl.register
+
+@Suppress("UNUSED")
+class PluginExtensionLegacy : Plugin<Project> {
```
AwkwardPeak7, line 22: why name it legacy? there is no replacement yet (nitpick ig)

FourTOne5: I would rather not move everything later

AwkwardPeak7 (approved):

PR #15591, opened 2026-05-06
Add DGManga

Exinaus:
Checklist:

Closes #13067

- [ ] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [ ] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [ ] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [x] Have removed `web_hi_res_512.png` when adding a new extension

PR #15690, opened 2026-05-09
DGManga update

Exinaus:
Checklist:

Changing the way url stored in `SChapter`, in case there might be a need for a change in `pageListParse` for another method to get/download images (instead of current regex from javascript in the body).

### DIFF
--- a/src/uk/dgmanga/src/eu/kanade/tachiyomi/extension/uk/dgmanga/DGManga.kt
+++ b/src/uk/dgmanga/src/eu/kanade/tachiyomi/extension/uk/dgmanga/DGManga.kt
@@ -34,6 +34,7 @@ class DGManga :
     private val apiUrl = "https://dgmanga.app/api"
     override val lang = "uk"
     override val supportsLatest = true
+    override val versionId = 2
     private val preferences by getPreferencesLazy()
     override val client = network.cloudflareClient.newBuilder()
         .build()
--- a/src/uk/dgmanga/src/eu/kanade/tachiyomi/extension/uk/dgmanga/DGManga.kt
+++ b/src/uk/dgmanga/src/eu/kanade/tachiyomi/extension/uk/dgmanga/DGManga.kt
@@ -154,6 +154,12 @@ class DGManga :
     }
 
     // ============================== Manga/Chapters ===============================
+    // WebView url
+    override fun getChapterUrl(chapter: SChapter): String {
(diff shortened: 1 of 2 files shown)


src/uk/dgmanga/src/eu/kanade/tachiyomi/extension/uk/dgmanga/DGManga.kt
```diff
@@ -154,15 +154,23 @@ class DGManga :
...
     override fun chapterListParse(response: Response): List<SChapter> {
         val dto = response.parseAs<List<ChapterResponseDto>>()
         return dto.map { dto ->
             SChapter.create().apply {
-                name = "Том ${dto.volumeNumber} Розділ ${dto.chapterNumber.toString().removeSuffix(".0")} ${dto.chapterName ?: ""}"
-                url = "/read/${dto.title}/${dto.chapterNumber.toString().removeSuffix(".0")}?chapterId=${dto.id}"
+                val vol = dto.volumeNumber.toString().removeSuffix(".0")
+                val num = dto.chapterNumber.toString().removeSuffix(".0")
+                name = "Том $vol Розділ $num ${dto.chapterName ?: ""}"
+                url = "${dto.id}/$num/${dto.title}"
```
vetleledaal, line 174: This introduces a new URL format, You should provide a migration path or bump the `versionId` if they aren't compatible. Changing the URL will in any case cause issues for existing entries in your library- popular/latest/search won't be able to indicate duplicates and downloaded chapters might get unlinked.

Exinaus: I've bumped versionId. Extension were released just 2 days ago, in #15591 so i kinda hoped to do the change before many people were affected.

vetleledaal (approved):

src/uk/dgmanga/src/eu/kanade/tachiyomi/extension/uk/dgmanga/DGManga.kt
AwkwardPeak7: I would say bumping versionId is not needed, simply add a check in `pageListRequest` to refresh chapter list if old format is detected. 
versionId bump is mostly relevant when manga url changes and since manga url can't be updated with a refresh, it needs migration

PR #15696, opened 2026-05-09
Comix: filters/details rework + WebView-based signer for protected endpoints

### LINKED ISSUE #15680: Comix demographic filtering not working
### Source information

Comix 1.4.14

### Source language

English 

### Steps to reproduce

I am trying to exclude josei demographic but it's not working 

### Expected behavior

It should exclude josei demographic 

### Actual behavior

It's showing josei titles 

### Mihon/Mihon fork version

Komikku app.komikku.beta r10472

### Android version

13

### Other details

_No response_

### Acknowledgements

### 

---

### LINKED ISSUE #15693: Comix [EN]: Add New Filters, Preferences Settings and Manga Details
### Source name

Comix 1.4.14

### Source language

English

**Add new filters to Comix, such as:**
- Suggestive rating;
- Author search;
- Artist search;
- Update release year to match new website changes.

**Add new extension settings, such as:**
- Content preferences as extension settings (automatic search filtering? Note: Issue #14450 asks for the same feature, but it was before new Comix layout so the images are outdated);
- Content filter as extension settings to replace default hide NSFW content (Safe, Suggestive, etc...).

**And in manga details:** 
- Since author and artist search are added, it would be nice to search for them when the author/artist fields were pressed, just like genres are doing it in current version (using new filter author/artist search);
- Add new information where applicable (year of publication/licensing, language (e.g. ZH), content filter (Safe, etc...), rank (e.g. 84), amount of users that rated an entry and number of followers in description.

### Other details

**Comix filters:**

- New Author/Artist fields, suggestive rating (can override preferences if filtered manually and all search uses preference filters first?) and new release year format (oldest currently can go to 1928 and newest can go to 2027):
Image

**Comix settings/preferences:**

- For "Content filter" above:
Image

- For Content preferences above:
Image

**Comix manga details:**

- Add extra information in description:
Image

### Acknowledgements

### 

---

### LINKED ISSUE #15703: Comix: `HTTP 403` when loading chapter list
### Source information

Comix 1.4.14

### Source language

English

### Steps to reproduce

Open any entry
Get 403 error

### Expected behavior

Chapter list should display normally

### Actual behavior

403 error

### Mihon/Mihon fork version

Mihon 0.19.9 and latest Komikku beta

### Android version

Android 11

### Other details

Image
Webview loads everything fine

### Acknowledgements

### 

---

nobottomline:
Closes #15693.
Closes #15680.
Closes #15703.

Two independent things landed here, in the order they happened:

## Signer / response decryption (closes #15703)

Comix's `/manga/{hid}/chapters` and `/chapters/{cid}` endpoints require a `_=<token>` signature whose algorithm and keys live inside a Jscrambler-style obfuscated VM bundle. Both rotate on every site deploy. The previous fix (#15644) replaced the hard-coded keys; ~24h later the site rotated again and the source has been broken since.

Re-rolling the same hard-coded constants is no longer an option: the live `secure-*.js` no longer contains any `RC4`-shaped keys in plaintext — they are decrypted at runtime by the VM, which refuses to run outside a real browser (`VM decryption key not available` from QuickJS / Node with shims). On top of that, the API now also encrypts response bodies (`x-enc: 1` header → `{"e":"<blob>"}`), so even the right token returns something that fails `parseAs` with `MissingFieldException: result missing` — which is exactly what users see in the issue.

The new `Signer.kt` (renamed from `Hash.kt`) hosts a hidden off-screen `WebView` that loads `comix.to/` and lets the site's own bundle bootstrap. It then auto-detects two `vmf_<id>.<fn>` endpoints by behaviour:

- **signer** — the function whose output for a known path matches a base64url token shape;
- **installer** — the function that, when handed a fake `axios` instance, registers a response interceptor (the site's response-side decryptor).

Requests tagged with `X-Comix-WebView-Proxy: 1` are intercepted in `WebViewProxyInterceptor`, signed via the WebView, fetched, fed through the captured response interceptor, re-wrapped as `{result: <decoded>}` to match the existing DTOs, and handed back to OkHttp as a synthetic 200 `Response`. JS↔Kotlin uses `addJavascriptInterface` (avoids `evaluateJavascript` JSON round-trip truncation on large bodies). `chapterListRequest` and `pageListRequest` set the header; everything else flows through plain OkHttp.

Because the work is delegated to the site's own primitives, this survives:

- key rotation (keys live inside the VM, never seen by the extension),
- signer / installer name rotation (probe is behaviour-based, not name-based),
- namespace name rotation (every `vmf_*` is walked),
- algorithm rewrites — as long as `comix.to` itself can sign and decrypt in a browser, so can the extension.

Inspired by `lxhentai/TokenResolver`'s WebView pattern, but extended with a fake-axios shim to capture the response decryptor.

### DTO change

`/chapters/{id}` response shape changed: `pages` is now `{baseUrl, items: [{url, width, height}]}` (relative urls + a base) instead of a `List<{url}>` of absolute URLs. `Dto.kt` and `pageListParse` are updated; the join also tolerates absolute URLs in case the API switches back.

## Filters / content rating / details rework (closes #15693, #15680)

Reworks Comix's filter panel, adds matching source-level settings, and surfaces a few new manga-detail fields. The site's API was probed via Playwright to confirm the actual parameters used by `comix.to` (some had drifted: `release_year[from]` is now `year_from`, `?author=` is ignored unless replaced by `authors[]=ID`, etc.).

### Filters

- **Tags / Genres / Formats** are now three separate sections that mirror the site's browse panel. Tags accept comma-separated free text; Genres and Formats are the curated 30 + 9 entries the site shows. Narrative tags that used to live under Genres (Aliens, Animals, etc.) are now searched by name through Tags.
- **Match (AND/OR)** drives the site's `genres_mode` — defaults to AND like the site does.
- **Author / Artist** are now first-class text filters with multi-value (comma-separated) support. Names are resolved to numeric IDs via `/api/v1/tags/search` and submitted as `authors[]` / `artists[]`.
- **Content rating** filter ("Use preference" / Safe / Suggestive / Erotica / Pornographic). When set, it overrides the source-level preference for that search.
- **Year range** uses the site's current `year_from`/`year_to` parameters and now spans `1928..currentYear+1`.
- Filter order matches the site's UI.

### Settings

- The "Hide NSFW" switch is replaced by a **Content rating** list preference (Show all / Safe / Suggestive / Erotica / Pornographic). Default is **Suggestive** — same effective behaviour as the old switch's default. Users who had explicitly disabled "Hide NSFW" are migrated to "Show all".
- Added a **Show extra info in description** switch (default on) controlling the new fields below.

### Bonus fix (#15680)

The Demographic filter was using TriState (include/exclude/off), but the site's API silently ignores both `-id` and `demographics_ex[]` for this field. Switched to a checkbox-only filter so what users select is what actually gets filtered.

### Manga details

- Description appends year, original language, content rating, rank, ratings count, and follower count when available.
- Genre chips are now trimmed to type + curated genres + demographics (+ NSFW) instead of dumping the long narrative tag list — that matches what `comix.to` shows on the page.

### Quality

- Resolved tag IDs go through the same shared `/tags/search` helper for Author, Artist, and Tags.
- `firstInstanceOrNull` from `keiyoushi.utils` used per CONTRIBUTING.md guidance.
- Bumped `extVersionCode` to 15.

## Notes

- iOS forks that run extensions through a Swift-side runtime (Tachimanga and similar) need a `WKWebView` + `WKScriptMessageHandler` port of the signer to pick up this fix, since Kotlin's `WebView` has no equivalent on their side. Komu already has this in place.

## Checklist


MuhamadSyabitHidayattulloh:
Can you check this issues #15703?

Alpha-782:
Any way to also add a genre preference inclusion/exclusion or is it not planned?

nobottomline:
Out of scope for this PR — please open a feature request so it can be tracked properly. The pattern would mirror the new Content rating preference, so happy to pick it up if it lands as an issue.

Alpha-782:
It was mentioned in issue #15693, but it's not a problem. I can make another issue about it.

vetleledaal (approved):
A bit slow on my test device, but code can be improved later.

nobottomline:
Thanks for the merge! Yeah, the WebView cold-start is the trade-off for the runtime delegation — first signed request per session pays for the page load and bundle bootstrap. Plenty of low-hanging fruit (persisting the detected signer expression + token cache across sessions, eager warmup on source init) — happy to send a follow-up PR with the perf pass.

dajotim937:
> A bit slow on my test device, but code can be improved later.

Not sure "a bit" is describing it correctly. I got 2 timeouts out 8 titles on library update and titles with many chapters (as ORV) can't even load chapter list after 1-2 min of waiting.

nobottomline:
Yeah, that's much worse than 'a bit slow' — thanks for the actual numbers. The serialised-through-one-WKWebView design pays the bundle load + signer probe per cold-start session, and every signed call goes through that same queue, so library updates serialise badly and bigger chapter lists (ORV-tier) cascade into the per-call timeout.

Concrete fixes I'll send as a separate PR shortly:

- Persist the per-path signed token in SharedPreferences (tokens are deterministic for a given path until the site rotates, and the existing 401/403 retry already invalidates) — repeat library updates skip the WebView entirely
- Persist the detected `signer` / `installer` expression so cold start can skip the probe step and only re-probe on 401/403
- Eager WebView warmup at source init instead of lazy on first signed request

Should make cold start ~5× faster and repeat reads effectively instant. Will tag this PR / the relevant issues when it's up.

nobottomline:
Quick update on the perf pass — turned out trickier than I expected, so flagging where things stand instead of going dark.

**Where the time actually goes** (per chapter-list page, 100 chapters):
- Sign URL: ~1ms (cheap)
- `fetch()` over the network: 200–500ms
- **Decrypt the response via the captured axios interceptor: 1–3s** — this is the dominant cost; it walks the obfuscated VM bytecode for every item in the page
- JSON parse + re-stringify wrap: 100–500ms
- `addJavascriptInterface` bridge back to Kotlin (body grows large): 200–500ms

So ~2–5s per page on a quick device, regardless of network.

**Why parallelism plateaus around 3 in-flight**: the WebView's JS event loop is single-threaded. `await fetch()` yields cleanly so multiple network calls overlap, but `await __cmxRes(resp)` runs the decrypt synchronously and blocks the loop until it returns. So page 2 can be fetching while page 1 is decrypting, but page 3 ends up waiting on page 2's decrypt to finish before it can start its own. For a 6000-chapter title (~60 pages) that's roughly `(decrypt time × pages) / 3` floor — i.e. ~minute-and-change no matter how many threads we pile on Kotlin-side.

**What I can't change cheaply**:
- Move decrypt out of the WebView — the algorithm is anti-tamper-guarded VM bytecode and won't run in QuickJS / Node / etc., so we're locked into the page's own runtime
- Parallelise decrypt in JS — it's a single-threaded loop and the per-call work is CPU-bound

**Modest wins I tried** (persistent signer/installer cache, eager warmup, reusing the captured response interceptor across calls instead of re-installing per request, bounded-fan-out pagination) — they help cold start and shave per-call overhead, but the decrypt step dominates the wall time and they ran into their own iOS-fork shaped edge cases I haven't fully isolated yet. Pulling that work back rather than ship it half-cooked.

**The architectural shift that would actually move the needle**: split sign / fetch / decrypt instead of doing all three in the WebView per call. Sign once per manga via the WebView (cached in SharedPreferences — tokens are deterministic until the site rotates, which the existing 401/403 retry already invalidates), fetch the encrypted body via OkHttp directly (parallel native HTTP, no WebView serialisation), then bounce only the encrypted blob through the WebView for the decrypt. That removes the network roundtrip from the JS event loop entirely, lets Mihon's native client parallelise pulls properly, and leaves only the unavoidable decrypt step inside the JS thread. Expect roughly the 3× from there. It's a meaningful refactor though, so I want to take it as a proper PR with on-device tests rather than rush it. Tracking it for follow-up.

nobottomline:
Heads up that today's deploy rotated the namespace prefix from `vmf_` to `vmC_`, which the probe here was still hard-coded against — sorry for the regression. Fix is up in #15748 (drops the prefix constraint entirely so future rotations don't need a code change), closes #15729.

PR #15748, opened 2026-05-10
Comix: drop hard-coded namespace name from signer probe

### LINKED ISSUE #15729: Comix: `signer not detected after 25s` error
### Source information

Comic 1.4.15

### Source language

English

### Steps to reproduce

Image

I keep getting this error on Comix after latest version update.

"Comix: signer not detected after 25s. Try opening the source in WebView once to clear any anti-bot challenge, then refresh"

### Expected behavior

Should get chapters when refreshed.

### Actual behavior

Instead is giving an error.

### Mihon/Mihon fork version

App ID: app.komikku App version: 1.13.5

### Android version

Android 16

### Other details

_No response_

### Acknowledgements

### 

---

nobottomline:
Closes #15729.

Today's `comix.to` deploy rotated the obfuscated VM namespace from `vmf_<id>` to `vmC_<id>`. The probe in `Signer.kt` shipped in #15696 hard-coded `topName.indexOf('vmf_') !== 0`, so it no longer matched anything, the WebView poll never found the signer / installer, and every protected request died with `signer not detected after 25s`.

The KDoc on #15696 advertised this as covered by "namespace name rotation"; that was wrong — only the function names inside the namespace were behaviour-detected, the namespace name itself was still pinned.

## Probe is now two-tier

So the common case stays cheap:

- **Fast path** matches `^vm[A-Za-z]_` — every observed deploy so far. Same speed as a single-regex check.
- **Fallback** walks remaining top-level objects on `window` and keeps those whose own keys look like obfuscator output (≥5 own keys, ≥3 of them 1-3 char alphabetic identifiers — `Qi`, `Xi`, `Gi`, …). This is a near-unique fingerprint of the bundle's mangled exports and excludes every nominally-named browser global, so we never accidentally invoke `alert`, `confirm`, `open`, etc. as a side effect of probing. If the site ever rotates beyond `vm[A-Za-z]_`, no extension update is needed.

Within either tier the signer/installer detection is unchanged: signer = function returning a base64url token for a known path; installer = function that registers a response interceptor on a fake axios.

## A few concurrency fixes alongside

While the file was open I also addressed a handful of edge cases that mostly hide on a single-threaded use but show up on parallel library updates:

- `WebViewProxyInterceptor` is prepended to the chain via `interceptors().add(0, ...)` instead of `addInterceptor(...)`. Forks that ship an upstream short-circuit interceptor for `comix.to` traffic (e.g. iOS clients routing through native HTTP) would otherwise dispatch the request before our proxy runs and receive `Missing token`. There's a code comment explaining why.
- `invalidate()` calls `ready.countDown()` before swapping in a fresh `CountDownLatch`, releasing any thread already waiting on the old one.
- `ensureReady()` no longer spawns a second WebView while a first one is still loading; it falls through to the existing latch.
- The poll `Runnable` bails out if its captured WebView is no longer the active one — a race with `invalidate()`.
- Per-call correlation token uses `UUID.randomUUID()` instead of `nanoTime()`. No realistic collision risk either way; UUID is just the conventional choice for this kind of token.

Bumped `extVersionCode` to 16.

Cross-fork credit: @D-Brox ran into the same regression in nekomangaorg/Neko#3070 and shipped an equivalent regex-based fix; the concurrency cleanups above were sanity-checked against suggestions Gemini left on that PR.

## Checklist

### DIFF
--- a/src/en/comix/build.gradle
+++ b/src/en/comix/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Comix'
     extClass = '.Comix'
-    extVersionCode = 15
+    extVersionCode = 16
     isNsfw = true
 }
 
--- a/src/en/comix/src/eu/kanade/tachiyomi/extension/en/comix/Comix.kt
+++ b/src/en/comix/src/eu/kanade/tachiyomi/extension/en/comix/Comix.kt
@@ -34,8 +34,16 @@ class Comix :
 
     private val preferences: SharedPreferences by getPreferencesLazy()
 
+    // `WebViewProxyInterceptor` must run *first* in the chain so it
+    // short-circuits the request to the WebView before any other
+    // interceptor (rate-limit, forks' native-HTTP routers, etc.) can
+    // touch it — otherwise signed requests reach `comix.to` without the
+    // `_=token` parameter and the server replies "Missing token".
+    // Please keep `interceptors().add(0, …)` here even if it looks
+    // unusual; switching to `.addInterceptor(...)` reintroduces that
+    // failure mode on downstream forks that share this code.
     override val client = network.cloudflareClient.newBuilder()
-        .addInterceptor(WebViewProxyInterceptor)
+        .apply { interceptors().add(0, WebViewProxyInterceptor) }
         .rateLimit(5)
         .build()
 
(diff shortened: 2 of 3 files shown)


vetleledaal:
Closing in favor of #15763. If that doesn't cover every change you wanted to do then please open another PR.

dajotim937:
@vetleledaal 
You should have merge this one instead of previous, because current version is slow af.
More info: https://github.com/keiyoushi/extensions-source/pull/15696#issuecomment-4413735064

vetleledaal:
I tested both PRs before merging and they were comparable performance wise. The other PR also addresses issues with Suwayomi, which is why that one was used instead of this PR.

I ran some additional tests, which confirms they are pretty much the same:

### Title 1
- 29.45s #15748
- 28.20s #15763

### Title 2
- 1m27.11s #15748
- 1m27.36s #15763

### Title 3
- 06.77s #15748
- 05.03s #15763

PR #15765, opened 2026-05-10
WitchScans, Drake Scans, Elf Toon (EN): bypass locked chapters via WP REST API

nobottomline:
Three sites use a WordPress lock-chapter plugin that hides chapter images from anonymous users — witchscans/elftoon strip `` tags from the public chapter HTML, drake 302s the chapter URL to `/login/`. None of them close the standard WP REST API though, so `/wp-json/wp/v2/posts/<id>` still returns the canonical content with the original image URLs.

WitchScans and Elf Toon run the same custom \`wp-manga-auth-profiles\` plugin (same operator). Drake runs a different plugin (\`lock-chapters\`) but leaves the same hole.

### How it works

New helper `MangaThemesiaRestPageFallback` in `lib-multisrc/mangathemesia` (with a matching switch preference). All three extensions extract the WP post id from the chapter list (witch/elftoon from `a[data-id]`, drake from `[data-post-id]`), encode it on `chapter.url` as a fragment, and short-circuit `pageListRequest` to the REST endpoint when the toggle is on. `pageListParse` dispatches by content-type so free chapters keep going through the normal HTML parser.

### Defaults

- **WitchScans** — bypass default **ON**. The lock has no working payment flow on the chapter (the buy button currently just redirects to login).
- **Elf Toon** — bypass default **ON**. Same plugin / same operator as WitchScans, same rationale.
- **Drake Scans** — bypass default **OFF** (opt-in). Drake actually monetizes via paid early-access, so don't bypass without the user explicitly asking.

When the bypass is on, the chapter list selector also stops applying the existing \`pref_hide_paid_chapters\` filter so users don't have to flip two toggles to read locked chapters.

### Scope

A quick scan of other paywalled MangaThemesia extensions didn't turn up more sites with the same hole — most have \`wp/v2/posts\` disabled at the WP layer. There's almost certainly room for more targeted testing, this PR just covers the three I could verify end-to-end.

### Test

- WitchScans / Elf Toon: opened locked chapter, gets the full page list via REST. Free chapters still parse via HTML (no extra request, no regression). Toggling bypass off restores the original \"no pages\" / hidden-from-list behavior.
- Drake Scans: same flow. Verified the locked chapter \`martial-evolution-…-chapter-128\` returns 6 pages with bypass on; with bypass off the chapter URL still 302s to login as before.

Checklist:

- [ ] Updated \`extVersionCode\` value in \`build.gradle\` for individual extensions
- [x] Updated \`overrideVersionCode\` or \`baseVersionCode\` as needed for all multisrc extensions
- [ ] Referenced all related issues in the PR body (e.g. \"Closes #xyz\")
- [ ] Added the \`isNsfw = true\` flag in \`build.gradle\` when appropriate
- [ ] Have explicitly kept the \`id\` if a source's name or language were changed
- [ ] Have removed \`web_hi_res_512.png\` when adding a new extension

### DIFF
--- a/lib-multisrc/mangathemesia/assets/i18n/messages_en.properties
+++ b/lib-multisrc/mangathemesia/assets/i18n/messages_en.properties
@@ -32,3 +32,5 @@ pref_dynamic_url_title=Automatically update dynamic URLs
 pref_dynamic_url_summary=Automatically update random numbers in manga URLs.\nHelps mitigating HTTP 404 errors during update and "in library" marks when browsing.\nNote: This setting may require clearing database in advanced settings and migrating all manga to the same source.
 pref_hide_paid_chapters_title=Hide chapters which require a purchase
 pref_hide_paid_chapters_summary=Hide chapters which must be purchased using coins.\nYou might want to disable this if you want to be notified of paid chapters so that you can go purchase them.
+pref_unlock_paid_chapters_title=Unlock locked chapters via REST API
+pref_unlock_paid_chapters_summary=When a chapter renders without images (locked by the site), fall back to the WordPress REST API to extract the page list.\nDisable this to respect the site's lock.
--- a/lib-multisrc/mangathemesia/build.gradle.kts
+++ b/lib-multisrc/mangathemesia/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
     id("lib-multisrc")
 }
 
-baseVersionCode = 32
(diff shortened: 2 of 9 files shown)


prem-k-r:
Hi @nobottomline, Sorry for the ping.
If you don't mind, could you please resolve the conflicts and merge it into the main branch? I'd like to download the APKs.
Thank you for your time! 🙂

PR #15768, opened 2026-05-10
Reset Scans, Utoon (EN): bypass locked chapters via icmadara REST API

nobottomline:
Both sites run the icmadara WP plugin alongside `wp-manga-chapter-coin` / `mycred`. The lock plugin filters the chapter HTML, but icmadara exposes `/wp-json/icmadara/v1/mangas/slug/<slug>` (full chapter list with post ids) and `/wp-json/icmadara/v1/capitulo/<id>` (image list) — both unauthenticated, so the canonical chapter content survives the lock.

### How it works

New helper `MadaraIcmadaraPageFallback` in `lib-multisrc/madara`. When the bypass is on, `chapterListParse` uses the icmadara mangas-by-slug endpoint as the source of truth (instead of the HTML selector that hides locked entries) — that returns all chapters with proper URLs and stamps the capitulo id into each `chapter.url` as a fragment. `pageListRequest` then short-circuits to `/wp-json/icmadara/v1/capitulo/<id>`; `pageListParse` dispatches by content-type so free chapters keep going through the normal Madara pipeline. Disabling the toggle restores the parent's HTML behaviour exactly.

### Defaults

- **Reset Scans** — bypass default **ON**. The lock has no working purchase flow attached, it just gates visibility on the public HTML.
- **Utoon** — bypass default **OFF** (opt-in). Utoon actually monetizes via paid coins (`monedas` field on the icmadara payload), leave the bypass behind a deliberate toggle.

### Scope

A quick scan of other Madara extensions didn't turn up more sites with the same leaky icmadara setup — most either don't ship the plugin or block the `capitulo` route. There's almost certainly room for more targeted testing, this PR just covers the two I could verify end-to-end.

### Test

- Reset Scans: chapter list now includes the locked chapters at the top of every series (instead of hiding them); opening one returns the full page list via `/icmadara/v1/capitulo/<id>` (~36-page chapter verified). Free chapters still parse via the standard Madara HTML reader (no extra request).
- Utoon: with bypass off, no behavioural change — the existing `:not(.premium-block)` filter still hides locked chapters. With bypass on, the locked chapter list shows up and chapter-46 (`monedas=7`) returns 17 pages. Free chapters unaffected.

Checklist:

- [ ] Updated \`extVersionCode\` value in \`build.gradle\` for individual extensions
- [x] Updated \`overrideVersionCode\` or \`baseVersionCode\` as needed for all multisrc extensions
- [ ] Referenced all related issues in the PR body (e.g. \"Closes #xyz\")
- [ ] Added the \`isNsfw = true\` flag in \`build.gradle\` when appropriate
- [ ] Have explicitly kept the \`id\` if a source's name or language were changed
- [ ] Have removed \`web_hi_res_512.png\` when adding a new extension

### DIFF
--- a/src/en/resetscans/build.gradle
+++ b/src/en/resetscans/build.gradle
@@ -3,7 +3,7 @@ ext {
     extClass = '.ResetScans'
     themePkg = 'madara'
     baseUrl = 'https://reset-scans.org'
-    overrideVersionCode = 11
+    overrideVersionCode = 10
     isNsfw = false
 }
 
--- a/src/en/resetscans/build.gradle
+++ b/src/en/resetscans/build.gradle
@@ -3,7 +3,7 @@ ext {
     extClass = '.ResetScans'
     themePkg = 'madara'
     baseUrl = 'https://reset-scans.org'
-    overrideVersionCode = 10
+    overrideVersionCode = 11
     isNsfw = false
 }
 
--- a/src/en/resetscans/src/eu/kanade/tachiyomi/extension/en/resetscans/ResetScans.kt
+++ b/src/en/resetscans/src/eu/kanade/tachiyomi/extension/en/resetscans/ResetScans.kt
@@ -1,6 +1,15 @@
 package eu.kanade.tachiyomi.extension.en.resetscans
 
+import androidx.preference.PreferenceScreen
 import eu.kanade.tachiyomi.multisrc.madara.Madara
+import eu.kanade.tachiyomi.multisrc.madara.MadaraIcmadaraPageFallback
+import eu.kanade.tachiyomi.source.ConfigurableSource
+import eu.kanade.tachiyomi.source.model.Page
+import eu.kanade.tachiyomi.source.model.SChapter
+import eu.kanade.tachiyomi.util.asJsoup
+import keiyoushi.utils.getPreferencesLazy
+import okhttp3.Request
+import okhttp3.Response
 import java.text.SimpleDateFormat
 import java.util.Locale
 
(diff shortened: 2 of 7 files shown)


src/en/utoon/build.gradle
```diff
@@ -3,7 +3,7 @@ ext {
     extClass = '.Utoon'
     themePkg = 'madara'
     baseUrl = 'https://utoon.net'
-    overrideVersionCode = 3
+    overrideVersionCode = 4
```
MuhamadSyabitHidayattulloh: No need to bump override version because it has been done in the base version

src/en/resetscans/build.gradle
```diff
@@ -3,7 +3,7 @@ ext {
     extClass = '.ResetScans'
     themePkg = 'madara'
     baseUrl = 'https://reset-scans.org'
-    overrideVersionCode = 10
+    overrideVersionCode = 11
```
MuhamadSyabitHidayattulloh: Same here

nobottomline:
fixed in ac96215 — reverted both. also did the same on #15765 since it had the same redundant bump on three extensions.

darwin-256:
No longer works

nobottomline:
@darwin-256 what specifically broke for you — the icmadara fallback (default on) or the html selector path? just pushed c89c75a40 with the same selector + date fix as #15958 so the html path matches the new site markup, but want to make sure that's actually what you hit and not something else

darwin-256:
> @darwin-256 what specifically broke for you — the icmadara fallback (default on) or the html selector path? just pushed [c89c75a](https://github.com/keiyoushi/extensions-source/commit/c89c75a400550df090e29e604f5a50fbc383b042) with the same selector + date fix as #15958 so the html path matches the new site markup, but want to make sure that's actually what you hit and not something else

Never mind apparently loaded wrong script when testing still works. Will close my PR for the fix since you already have this

DrPropheto:
After this pr was opened the current extensions for reset scans / utoon have stopped working as reported in #15913 
Something similar happened with Asura scans when a feature to bypass their monetized content was introduced, they retaliated by breaking the logic behind how their site worked, essentially making the extension unusable. 
I am not saying that this pr is what caused the change but I wonder how helpful it is to propose these bypasses when they will almost certainly lead to the websites taking measures against them; not only stopping the bypassers but also everyone who is willing to wait for the chapters to be unlocked or wants to read older chapters.

TeTKiK15:
I think if they wanted to approve these two rest api prs, they would have approved it by now. Either it cannot be done or they deemed it more appropriate not to do it. And only the resetscans are corrupted. If you haven't checked, you shouldn't say both are broken.

Xyndra:
I'm not sure how legal this is, considering this is clearly unintended behavior used to circumvent payment. Wouldn't want keiyoushi to get DMCA'd

Ali-Nayef:
How long does it usually take for things to get reviewed to be approved?

Xyndra:
> How long does it usually take for things to get reviewed to be approved?

It would have been approved a long time ago normally

Smol-Ame:
There is no guarantee or ETA for any request, including reviewing & approving proposed changes. No one is assigned any work or getting paid to do anything. All the work is done by contributors & maintainers volunteering their own time & efforts.

src/en/resetscans/src/eu/kanade/tachiyomi/extension/en/resetscans/ResetScans.kt
```diff
@@ -9,12 +18,69 @@ class ResetScans :
...
+    // live (~36 page chapter survives the lock). Default ON because the
+    // lock is purely a paywall on public HTML; the API itself doesn't
+    // honor any access check.
+    private val restFallback = MadaraIcmadaraPageFallback(defaultEnabled = true)
+
+    // Hide locked chapters from the HTML chapter list (anchors that point
+    // at "#"). When the bypass is on, chapterListParse swaps in the
+    // icmadara-derived list anyway, so this only affects the disabled-
+    // bypass path.
+    override fun chapterListSelector() = "li.wp-manga-chapter:not(:has(> a[href*=\"#\"]))"
```
darwin-256, line 43:
```suggestion
    override fun chapterListSelector() = "li.wp-manga-chapter:not(:has(a[href*='#']))"
```
Changed again

AwkwardPeak7:
image
image
the icmadara endpoints don't seem to exist

PR #15763, opened 2026-05-10
fix(Comix): support for suwayomi

D-Brox:
Some Suwayomi CEF method stubs where being called. Those fixes don't affect the extension for Mihon.

Refactors the eval probe result parsing to an interface function.
Also fixes #15729 with the `vm*_*` namespace detection.

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [ ] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [ ] Have removed `web_hi_res_512.png` when adding a new extension
- [ ] This PR is AI-assisted, I have reviewed the changes manually and confirmed they are not slop

PR #15792, opened 2026-05-11
Comix: cache captured response interceptor + persist signer refs

nobottomline:
Small follow-up on #15748 / #15763 — current main works, this just trims a couple of edges I ran into while looking at the slow chapter-list reports on big titles. Not a perf revolution; mostly a cleanup pass that picks up a modest win on huge lists and nudges robustness around site rotations.

## What changes

- **Decryptor is captured once at WebView boot** into `window.__cmxRes`. `proxyFetch` uses it directly instead of running `installer(fakeAxios)` every call. The installer goes through the VM dispatcher and isn't free, so on multi-page chapter lists those saved per-call ms add up.
- **Signer / installer expressions persist** to a dedicated `SharedPreferences` file. Warm starts go straight to installing the decryptor from the cached refs on `onPageFinished` instead of re-probing. If the cached refs are stale (site rotated since last run), the install JS signals `bridge.installerMissing()` and the poll falls back to a fresh probe in the same boot cycle — no more 30 s silent waits after a rotation.
- **A handful of small concurrency fixes**: `invalidate()` sets an error on pending slots before swapping the latch (otherwise a concurrent fetch can see `\"\"` and treat it as a successful empty body); the probe poll bails out if its captured WebView is no longer the active one; `ensureReady()` doesn't spawn a second WebView while a first one is still loading.
- **Hygiene**: explicit-key removal in `invalidate()` instead of `prefs.edit().clear()` (leaves room for unrelated keys in the same prefs file); bracket notation for both namespace and function name in generated refs (`window[\"vmx_…\"][\"Qi\"]`); `response.use { … }` around the synthetic response build; `headersAsJson` switched to `buildString { }` per CONTRIBUTING.md; `LOAD_TIMEOUT_S` 30 → 12 so silent waits surface as actionable errors instead of dragging on.

## Numbers

Same device, same Wi-Fi session, cold start each row, no chapter dedupe. Whatever the site does between deploys, I tried to stick to apples-to-apples: each timing is `main` vs this branch back-to-back.

| Title          | Chapters | main       | this branch | Δ           |
|----------------|---------:|------------|-------------|-------------|
| ORV            |      937 | 0:57.32    | 0:58.14     | +0.82 s     |
| Yuan Zun       |    3 469 | 3:20.48    | 3:23.03     | +2.55 s     |
| MP             |   12 716 | 12:50.79   | 12:24.12    | **−26.67 s** |

ORV and Yuan Zun are within the per-session noise floor on the same network. The MP win lines up with the per-call install-once theory: 127 pages × ~210 ms ≈ 27 s, which matches the delta within rounding. So not a step change — measurable but small, and only on really big lists.

(Side note for anyone reading this who might think the chapter-dedupe pref makes things faster: it doesn't. It runs the same fetch + decrypt loop and then adds LinkedHashMap-based comparison work on top. The pref's own description warns about it being slow on big lists. I tested with it off.)

## Things I tried that didn't pay off

Two attempts I had to revert before this PR, mentioning them in case anyone else considers them:

- **Sign / fetch / decrypt split** routing the actual fetch through OkHttp (so Mihon's rate limiter and parallel library updates could batch network I/O properly). On sequential single-manga chapter lists this turned out a small net regression — the response body crossing the Kotlin↔JS boundary for decrypt and the extra interceptor-chain work per page costs more than the parallel-fetch potential, which isn't even exercised on this workload.
- **Pre-signing all chapter paths after `chapterListParse`**. The single JS pass over thousands of paths held the WebView main thread long enough that the next manga's request and any in-flight decrypt queued behind it. Net negative on big lists.

The real step-change for big chapter lists is a WebView pool — multiple WebViews each holding their own captured decryptor, so the obfuscated VM's per-item decrypt work can actually run in parallel instead of serialising through one JS event loop. That's a bigger piece of work (+150-200 MB RAM at peak, coordination, anti-tamper unknowns), so I'd want to do it as its own PR with its own benchmarks rather than slip it in here.

## Checklist

- [x] Referenced all related issues in the PR body

### DIFF
--- a/src/en/comix/src/eu/kanade/tachiyomi/extension/en/comix/Comix.kt
+++ b/src/en/comix/src/eu/kanade/tachiyomi/extension/en/comix/Comix.kt
@@ -55,8 +55,24 @@ class Comix :
 
     override fun imageUrlParse(response: Response) = throw UnsupportedOperationException()
 
+    /**
+     * The first time the user opens the source we know they intend to
+     * browse, so warm up the signer in the background. Doing it here
+     * (and not in `init {}`) keeps idle sources that Mihon eagerly
+     * instantiates on the source list from paying the WebView cost.
+     */
+    private fun maybeWarmUpSigner() {
+        if (!signerWarmedUp) {
+            signerWarmedUp = true
+            Signer.warmUp()
+        }
+    }
+
+    @Volatile private var signerWarmedUp = false
+
     // ============================== Popular ==============================
     override fun popularMangaRequest(page: Int): Request {
+        maybeWarmUpSigner()
         val url = apiUrl.toHttpUrl().newBuilder().apply {
             addPathSegment("manga")
             addQueryParameter("order[score]", "desc")
--- a/src/en/comix/src/eu/kanade/tachiyomi/extension/en/comix/Comix.kt
+++ b/src/en/comix/src/eu/kanade/tachiyomi/extension/en/comix/Comix.kt
@@ -37,15 +37,12 @@ class Comix :
 
     override val client = network.cloudflareClient.newBuilder()
         .apply {
-            // Insert at position 0 so requests carrying PROXY_HEADER reach
-            // our interceptor BEFORE any other application interceptor.
-            // Critical for Komu (iOS): its `CallNativeHttpInterceptor`
-            // re-routes comix.to traffic through Foundation NSURLSession
-            // to match Safari's TLS fingerprint and short-circuits the
-            // chain — adding our interceptor at the END would mean it
-            // never sees the request. Placing it first is harmless on
-            // Mihon Android too (no interceptor before us cares about
-            // PROXY_HEADER, and we always pass-through unsigned reqs).
+            // Prepended so requests carrying PROXY_HEADER reach our
+            // interceptor before any other application interceptor. Harmless
+            // on Mihon (no interceptor before us cares about PROXY_HEADER,
+            // and we pass through unsigned requests unchanged), but matters
+            // for downstream forks that wrap the chain with their own
+            // short-circuit on `comix.to`.
             interceptors().add(0, WebViewProxyInterceptor)
         }
         .rateLimit(5)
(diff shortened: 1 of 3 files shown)


src/en/comix/src/eu/kanade/tachiyomi/extension/en/comix/Signer.kt
```diff
@@ -281,7 +343,10 @@ object Signer {
                 )
                 wv.layout(0, 0, WEBVIEW_WIDTH, WEBVIEW_HEIGHT)
                 wv.clearCache(true)
-            } catch (_: RuntimeException) { /* Unsupported by Suwayomi */ }
+            } catch (_: RuntimeException) {
+                // Suwayomi's `WebView` shim throws "Stub!" on these. Mihon's
+                // real WebView and Komu's WKWebView shim handle them fine.
```
FourTOne5: What is this "Komu"

Alpha-782: iOS thing?

FourTOne5: That's not an answer

nobottomline: Yeah — it's a personal iOS project I work on elsewhere, the note had no business being in the public code. Dropped both refs in 8f7601a9.

src/en/comix/src/eu/kanade/tachiyomi/extension/en/comix/Comix.kt
```diff
@@ -37,15 +37,12 @@ class Comix :
...
-            // chain — adding our interceptor at the END would mean it
-            // never sees the request. Placing it first is harmless on
-            // Mihon Android too (no interceptor before us cares about
-            // PROXY_HEADER, and we always pass-through unsigned reqs).
+            // Prepended so requests carrying PROXY_HEADER reach our
+            // interceptor before any other application interceptor. Harmless
+            // on Mihon (no interceptor before us cares about PROXY_HEADER,
+            // and we pass through unsigned requests unchanged), but matters
+            // for downstream forks that wrap the chain with their own
+            // short-circuit on `comix.to`.
```
FourTOne5, line 45: something you should keep in mind is that we don't really care about any other downstream fork. if your app doesn't work like vanilla Mihon then your app should be fixed. small stuff like this is ok but we won't add anything big workaround

PR #15838, opened 2026-05-13
fix(comix): Suwayomi looper deadlock

D-Brox:
For some reason, on Suwayomi's LooperThread, `handler.post { ... }` + `latch.await()` deadlocks.

The LooperThread posts work to itself then blocks, so the work never runs. The guard runs the WebView creation inline when already on the LooperThread, avoiding the deadlock. On Android this branch is never taken (background thread calls `captureToken` instead), so it's harmless.

Checklist:

### DIFF
--- a/src/en/comix/build.gradle
+++ b/src/en/comix/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Comix'
     extClass = '.Comix'
-    extVersionCode = 18
+    extVersionCode = 19
     isNsfw = true
 }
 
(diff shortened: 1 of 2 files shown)


AwkwardPeak7 (approved):
This should ideally be fixed in suwayomi as there are multiple extensions using this `handler.post` pattern

D-Brox:
I agree, tho the deadlock seems to only affect the page list function. I tested with a few others that also use the WebView, but it is only happening to Comix. I suspect it is only affecting it because of the runBlocking call tho.

PR #15850, opened 2026-05-13
[TR] Fix MangaTR extension

dpentx:
Checklist:

@mkkurisu got this extension working again. I cleaned up some rough edges I noticed, so I’m opening this PR. The extension is currently functional.

Since images are split using Canvas, it needs to be used in vertical scrolling mode for panels to display properly. Horizontal scrolling also does not work because of Mihon’s page merging behavior.


darwin-256:
# PR Review: Manga-TR Extension

Thanks for the contribution! After reviewing the code against the contributing guidelines, here are the issues that need to be addressed:

---

## Critical Issues

### 1. Uses deprecated `ParsedHttpSource`
`ParsedHttpSource` is deprecated. The class should extend `HttpSource` instead.
```kotlin
// Current
class MangaTR : ParsedHttpSource() {
// Should be
class MangaTR : HttpSource() {
```

### 2. `themePkg` mismatch in `build.gradle`
`themePkg = 'fmreader'` is declared, but the class extends `ParsedHttpSource` instead of the fmreader theme class. Either use the theme properly or remove `themePkg` entirely.

### 3. `imageUrlParse` returns empty string instead of throwing
Per the guidelines, unused inherited methods should throw `UnsupportedOperationException`:
```kotlin
// Current
override fun imageUrlParse(document: Document): String = ""
// Should be
override fun imageUrlParse(response: Response): String = throw UnsupportedOperationException()
```

---

## Code Quality Issues

### 4. `Page` instantiation doesn't use named parameters
```kotlin
// Current
pages.add(Page(pages.size, "", url))
// Should be
pages.add(Page(pages.size, imageUrl = url))
```

### 5. Redundant `.trim()` on Jsoup `.text()` calls
Jsoup's `.text()` already trims whitespace. Remove all `.trim()` after `.text()`:
- `popularMangaFromElement`: `element.text().trim()` → `element.text()`
- `searchMangaFromElement`: `element.text()?.trim()` → `element.text()`
- `mangaDetailsParse`: multiple instances
- `chapterFromElement`: `link.text().trim()` → `link.text()`

### 6. Comments in Turkish instead of English
All code comments should be in English for a shared repository:
```kotlin
// Current: "// Base64 şifresini çöz ve 0x5A XOR anahtarı ile kilidi aç"
// Current: "// Çıkan sonucu JSON olarak oku ve sırayı listeye kaydet"
// Current: "// Şifreli sıralama haritası"
```

### 7. Empty catch block in `pageListParse`
Silently swallowing exceptions hides bugs. At minimum, log the error or consider removing the try/catch:
```kotlin
} catch (e: Exception) {
    // Silently ignored - bad practice
}
```

### 8. `e.printStackTrace()` in `decodeOrder`
This is not proper logging in Android. Use the extension's logging utilities or remove it.

---

## Architecture / Best Practice Issues

### 9. Custom `cleanUrl` instead of `setUrlWithoutDomain` + `absUrl`
The guidelines recommend using `element.absUrl("href")` combined with `setUrlWithoutDomain()`. The custom `cleanUrl` method is fragile and should be replaced.

### 10. Inline header building in `imageRequest`
Custom headers should be assigned to a separate variable or class-level property, not built inline:
```kotlin
// Current
override fun imageRequest(page: Page): Request = GET(page.imageUrl!!, headersBuilder().add("Accept", "image/avif,image/webp,*/*").build())

// Should be
private val imageHeaders by lazy {
    headersBuilder().add("Accept", "image/avif,image/webp,*/*").build()
}
override fun imageRequest(page: Page): Request = GET(page.imageUrl!!, imageHeaders)
```

### 11. `supportsLatest = true` but latest is identical to popular
Per the guidelines: *"If a source only has a latest listing, use the latest listing in place for the popular listing and set `supportsLatest = false`."* Since both listings are identical, set `supportsLatest = false` and remove the `latestUpdates*` overrides.

### 12. Uses `org.json.JSONArray`/`JSONObject` instead of `keiyoushi.utils.parseAs`
The guidelines recommend using `parseAs` with `@Serializable` DTOs for JSON parsing instead of the manual `org.json` API. Both in `pageListParse` and `decodeOrder`, the JSON should be parsed using the recommended approach with proper DTO classes (using `class` not `data class`).

### 13. Synchronous blocking calls in `chapterListParse` loop missing `.use {}`
The `while` loop with `client.newCall(...).execute()` makes blocking network calls. You should ensure the response bodies are properly closed by wrapping them in `.use { }`:
```kotlin
// Current
val postResponse = client.newCall(POST(requestUrl, chapterHeaders, body)).execute()
document = postResponse.asJsoup()

// Should be
client.newCall(POST(requestUrl, chapterHeaders, body)).execute().use { postResponse ->
    document = postResponse.asJsoup()
}
```

dpentx:
@darwin-256,

Thanks for your suggestions. I couldn’t apply some of them because they caused errors during the build process. After going through your suggestions, I took another quick look at the extension and the pages started loading a bit faster. Hopefully, that will be enough for the extension to get accepted 😄

Starmania:
You added back ParsedHttpSource...

Smol-Ame:
Marking this as a draft, especially since this seems like half-tested AI slop. Double-check the code & apply the suggestions first before any considerations is considered.

dpentx:
@Starmania, @Smol-Ame can you check this now?

src/tr/mangatr/src/eu/kanade/tachiyomi/extension/tr/mangatr/MangaTR.kt
```diff
@@ -1,490 +1,244 @@
...
-    override val requestPath = "manga-list-sayfala.html"
-
-    // Popular
-
-    override fun popularMangaNextPageSelector() = "div.pagination-wrap a.pagination-link"
+class MangaTR : HttpSource() {
+    override val name = "Manga-TR"
+    override val baseUrl = "https://manga-tr.com"
+    override val lang = "tr"
+    override val supportsLatest = false // Guideline Issue 11
```
MuhamadSyabitHidayattulloh, line 32:
```suggestion
    override val supportsLatest = false
```

MuhamadSyabitHidayattulloh: Remove such comment tags

src/tr/mangatr/src/eu/kanade/tachiyomi/extension/tr/mangatr/MangaTR.kt
```diff
@@ -1,490 +1,244 @@
...
-            .build()
-        return GET(url, headers)
-    }
+    override val client: OkHttpClient = network.cloudflareClient.newBuilder()
+        .addInterceptor(DDoSGuardInterceptor(network.cloudflareClient))
+        .rateLimit(2)
+        .build()
 
-    // Latest
+    private val json: Json by injectLazy()
```
MuhamadSyabitHidayattulloh, line 39: You can use parseAs utils for json parser

src/tr/mangatr/src/eu/kanade/tachiyomi/extension/tr/mangatr/MangaTR.kt
```diff
@@ -1,490 +1,244 @@
...
-            includedGenres.isNotEmpty() -> {
-                includedGenres.forEach { url.addQueryParameter("tur", it.id) }
-            }
-            specialTur != null -> {
-                url.addQueryParameter("tur", specialTur)
+    override fun popularMangaParse(response: Response): MangasPage {
+        val document = response.asJsoup()
+        val mangas = document.select("li.sidebar-list__item:has(a.sidebar-list__link), div.media-card").map { element ->
+            SManga.create().apply {
+                val link = element.selectFirst("a.sidebar-list__link, a.media-card__link, a")
```
MuhamadSyabitHidayattulloh, line 59: Link and title must be mandatory

src/tr/mangatr/src/eu/kanade/tachiyomi/extension/tr/mangatr/MangaTR.kt
```diff
@@ -1,490 +1,244 @@
...
-            .firstOrNull { it.text().trim() == "›" }
-            ?.attr("href")
-            ?.let { href -> Regex("""[?&]page=(\d+)""").find(href)?.groupValues?.get(1)?.toIntOrNull() }
-        val hasNextPage = nextPageFromArrow != null && nextPageFromArrow > currentPage
+    override fun imageRequest(page: Page): Request = GET(page.imageUrl!!, imageHeaders) // Guideline Issue 10
 
-        return MangasPage(mangas, hasNextPage)
+    private fun parseRelativeDate(date: String): Long {
+        val calendar = Calendar.getInstance()
+        val number = Regex("""(\d+)""").find(date)?.value?.toIntOrNull() ?: return 0
```
MuhamadSyabitHidayattulloh, line 233: Separate regex

Starmania: (Move regex inside the class or as a companion. This way, it is not recreated for every date parse)

src/tr/mangatr/build.gradle
```diff
@@ -3,7 +3,7 @@ ext {
     extClass = '.MangaTR'
     themePkg = 'fmreader'
```
Starmania, line 4:
```suggestion
```

src/tr/mangatr/src/eu/kanade/tachiyomi/extension/tr/mangatr/MangaTR.kt
```diff
@@ -1,490 +1,244 @@
...
-                .takeUnless { it.isBlank() }
-
-            genre = document.detailMetaLinks("Tür", "tur=").joinToString { it.text() }
-
-            val translateLabel = document.detailMetaLink("Çeviri", "ceviri").orEmpty()
-            status = parseTranslateStatusBadge(translateLabel)
+            title = document.selectFirst("h1")?.text() ?: "" // Guideline Issue 5
+            thumbnail_url = document.selectFirst("img.poster-card__image, img[src*='covers']")?.absUrl("src")
+            description = document.selectFirst("div.detail-copy")?.text() // Guideline Issue 5
+            author = document.select("span.detail-meta-row__value:contains(Yazar), a[href*='author']").text() // Guideline Issue 5
```
Starmania, line 100: Could you double check that author is right in app ?

src/tr/mangatr/src/eu/kanade/tachiyomi/extension/tr/mangatr/MangaTR.kt
```diff
@@ -1,490 +1,244 @@
...
-    // Search
-
-    private var captchaUrl: String? = null
-
-    private var cachedGenres: List<Genre> = emptyList()
+    override fun headersBuilder() = super.headersBuilder()
+        .add("Referer", "$baseUrl/")
+        .add("Accept", "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8")
+        .add("Accept-Language", "tr-TR,tr;q=0.9,en-US;q=0.8,en;q=0.7")
+    // Intentionally removed User-Agent to let Mihon use the default WebView User-Agent.
```
Starmania, line 45:
```suggestion
```

src/tr/mangatr/src/eu/kanade/tachiyomi/extension/tr/mangatr/MangaTR.kt
```diff
@@ -1,490 +1,244 @@
...
-            AgeRestrictionFilter(),
-            ContentTypeFilter(),
-            SpecialTypeFilter(),
-        )
-        if (cachedGenres.isNotEmpty()) {
-            baseFilters += GenreList(cachedGenres)
-        }
-        return FilterList(baseFilters)
+    private val imageHeaders by lazy {
+        // Guideline Issue 10
```
Starmania, line 48:
```suggestion
```

src/tr/mangatr/src/eu/kanade/tachiyomi/extension/tr/mangatr/MangaTR.kt
Starmania, line 39:
```suggestion
```

src/tr/mangatr/src/eu/kanade/tachiyomi/extension/tr/mangatr/MangaTR.kt
```diff
@@ -1,490 +1,244 @@
...
-import keiyoushi.utils.firstInstanceOrNull
-import keiyoushi.utils.parseAs
-import kotlinx.serialization.json.JsonArray
-import kotlinx.serialization.json.JsonElement
-import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.intOrNull
+import kotlinx.serialization.json.jsonArray
+import kotlinx.serialization.json.jsonObject
 import kotlinx.serialization.json.jsonPrimitive
```
Starmania, line 18:
```suggestion
import keiyoushi.utils.parseAs
```
Use parseAs (part 1)

src/tr/mangatr/src/eu/kanade/tachiyomi/extension/tr/mangatr/MangaTR.kt
```diff
@@ -1,490 +1,244 @@
...
+import okhttp3.OkHttpClient
 import okhttp3.Request
 import okhttp3.Response
-import org.jsoup.nodes.Document
-import org.jsoup.nodes.Element
 import rx.Observable
-import java.io.IOException
-import java.nio.charset.StandardCharsets
-import java.util.Locale
+import uy.kohesive.injekt.injectLazy
```
Starmania, line 25:
```suggestion
```

src/tr/mangatr/src/eu/kanade/tachiyomi/extension/tr/mangatr/MangaTR.kt
```diff
@@ -1,490 +1,244 @@
...
+                val partIdx = key.toIntOrNull()
+                val pos = element.jsonPrimitive.intOrNull
+                if (partIdx != null && pos != null) list.add(Pair(partIdx, pos))
+            }
         }
-        date_upload = parseRelativeDate(
-            element.selectFirst("div.chapter-card__meta span")?.text()?.trim() ?: "",
-        )
+        return list
     }
```
Starmania, line 180:
```suggestion
    // Base64 + XOR 0x5A decode, then JSON: either [pos0, pos1, …] (index=partIdx) or {"partIdx": pos, …}
    private fun decodeOrder(encoded: String): List<Pair<Int, Int>> {
        val raw = Base64.decode(encoded, Base64.DEFAULT)
        val jsonStr = String(
            raw.map { ((it.toInt() and 0xFF) xor 0x5A).toByte() }.toByteArray(),
            Charsets.UTF_8,
        ).trimEnd('\u0000').trim()

        return if (jsonStr.startsWith("[")) {
            jsonStr.parseAs<List<Int?>>().mapIndexedNotNull { index, pos ->
                pos?.let { index to it }
            }
        } else {
            jsonStr.parseAs<Map<String, Int?>>().mapNotNull { (key, pos) ->
                if (pos != null) key.toIntOrNull()?.let { it to pos } else null
            }
        }
    }
```

src/tr/mangatr/src/eu/kanade/tachiyomi/extension/tr/mangatr/MangaTR.kt
```diff
@@ -1,490 +1,244 @@
...
-        if (parsed.isNotEmpty()) {
-            cachedGenres = parsed
-        }
-    }
 
-    override fun popularMangaParse(response: Response): MangasPage {
-        val document = response.asJsoup()
-        cacheGenresFromListPage(document)
+        return pages
+    }
```
Starmania, line 225:
```suggestion
    override fun pageListParse(response: Response): List<Page> {
        val document = response.asJsoup()
        val pages = mutableListOf<Page>()

        document.select("div.chapter-page").forEach { chapterPage ->
            val rawParts = chapterPage.attr("data-parts")
            val rawOrder = chapterPage.attr("data-order") // Encrypted map

            if (rawParts.isNotEmpty()) {
                val parts = rawParts.replace("&quot;", "\"")
                    .parseAs<List<String>>()
                    .map { it.replace("\\/", "/").replace("&amp;", "&") }

                if (rawOrder.isNotEmpty()) {
                    decodeOrder(rawOrder)
                        .sortedBy { it.second }
                        .forEach { (partIdx, _) ->
                            parts.getOrNull(partIdx)?.let { url ->
                                pages.add(Page(pages.size, imageUrl = url))
                            }
                        }
                } else {
                    parts.forEach { url -> pages.add(Page(pages.size, imageUrl = url)) }
                }
            }
        }

        if (pages.isEmpty()) {
            throw Exception("Page list is empty. Please open in WebView to bypass the security check.")
        }

        return pages
    }
```

Starmania (commented):
Use parseAs (part 2)

PR #15860, opened 2026-05-13
Enhance CONTRIBUTING.md with new guidelines

darwin-256:
Add guidelines for configurable sources and preferences in CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,6 +53,7 @@ or fixing it directly by submitting a Pull Request.
     - [Misc notes](#misc-notes)
     - [Advanced Extension features](#advanced-extension-features)
       - [Extension logic and app features](#extension-logic-and-app-features)
+      - [Configurable Sources and Preferences](#configurable-sources-and-preferences)
       - [URL intent filter](#url-intent-filter)
       - [Update strategy](#update-strategy)
       - [Renaming existing sources](#renaming-existing-sources)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -466,6 +467,12 @@ class MyDto(
 }
 ```
 
+- **Use `@Serializable` classes instead of `JsonObject`:** Do not manually traverse `JsonObject` or `JsonArray`. Define `@Serializable` classes and use `parseAs<T>()`.
+- **Map only used fields:** Do not map all fields from the JSON response in your DTOs if they are not used. Omit unused fields to keep the class clean and reduce bytecode.
+- **Mandatory fields should not have defaults:** Do not provide default empty/null values to mandatory fields (like a manga's ID or title) in DTOs just to avoid parsing exceptions. Let the parser fail early so broken entries are detected.
+- **Avoid `buildJsonObject` for requests:** Instead of manually building `JsonObject` with `buildJsonObject { put(...) }`, define a `@Serializable` request DTO class and use `toJsonRequestBody()`.
+- **Do not use `peekBody` for JSON:** Avoid using `response.peekBody(Long.MAX_VALUE).string()` to read JSON. Use `response.parseAs<T>()` directly.
```
AwkwardPeak7: `peekBody` has it's use in interceptors. did you mean `response.body.string.parseAs()`?

CONTRIBUTING.md
```diff
@@ -804,6 +822,9 @@ empty, so the app will skip the `fetchImageUrl` step and directly call `fetchIma
 - **Do not override default `HttpSource` methods:** Avoid overriding methods like `mangaDetailsRequest` or `chapterListRequest` if they only replicate the default behavior (`GET(baseUrl + manga.url, headers`). Only override them if the source requires a different URL structure or custom headers for those specific requests.
 - **Configurable sources:** By implementing `ConfigurableSource`, you can add settings backed by `SharedPreferences`.
 - **Code organization:** For readability, group related methods together in your extension class (e.g., all popular manga methods, then all latest manga methods, then search methods, and so on). A logical ordering like Popular → Latest → Search → Details → Chapters → Pages → Filters → Utilities makes the class easier to navigate and maintain without needing explicit section header comments.
+- **DTO extensions:** Move mapping extensions for DTOs (like `fun MyDto.toSManga()`) into the DTO file itself to keep the main source class clean.
+- **Libraries for single sources:** Do not create a `lib/` module for code that is only used by a single multisrc theme. Keep it within the multisrc module.
```
AwkwardPeak7: this very much depends imo

CONTRIBUTING.md
```diff
@@ -804,6 +822,9 @@ empty, so the app will skip the `fetchImageUrl` step and directly call `fetchIma
...
+- **`initialized` flag:** You do not need to manually set `initialized = true` inside `getMangaDetails`. The app's base `fetchMangaDetails` method automatically sets it to `true` after successfully calling `getMangaDetails`.
```
AwkwardPeak7: remove this too, imo better to be explicit where needed instead of relying on app behavior. since overriding `fetchMangaDetails` means app no longer sets it.

AwkwardPeak7 (approved):

PR #15988, opened 2026-05-17
Refactor: Replace deprecated network.cloudflareClient globally

darwin-256:
Checklist:


lib-multisrc/comicaso/src/eu/kanade/tachiyomi/multisrc/comicaso/Comicaso.kt
```diff
@@ -25,7 +25,7 @@ abstract class Comicaso(
 
     override val supportsLatest = true
 
-    override val client = network.cloudflareClient
+    override val client = network.client
```
vetleledaal: Is this needed? It's now effectively [the same as the default](https://github.com/mihonapp/mihon/blob/e8c6e3364dded0cb21df572be175cbb01adc243f/source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/online/HttpSource.kt#L67-L68).
```suggestion
```

bapeey: You can use this regex to delete redundant client overrides

```regex
# Python
(?m)^\s*\n?override\s+val\s+client(?:\s*:\s*OkHttpClient)?\s*=\s*network\.client\s*$(?!\n\s*\.)
```

AwkwardPeak7: this applies to all files

darwin-256: solved

AwkwardPeak7 (approved):

PR #15993, opened 2026-05-17
feat(YellowNote): Add video page support with mp4 external playback

GGggb112:
## Summary

Add video support to YellowNote (XChina source) extension. Works with the reader's built-in "Open externally" menu for one-tap video playback.

## How It Works

1. Parses the embedded `var videos = [...]` JS array from photo detail pages
2. Extracts video CDN domain (`img.xchina.io`)
3. Creates `Page` entries with:
   - `url` → full mp4 address (for external playback)
   - `imageUrl` → first image as cover thumbnail
4. Video pages appear before image pages in the reader

**User flow**: Swipe to a video page → see cover → tap menu "Open externally" → system player opens mp4

### YellowNote.kt
- Added `videosRegex`, `videoUrlRegex`, `domainRegex` to extract video data from embedded JS
- Rewrote `pageListParse()`:
  - Now reads raw HTML body for JS extraction
  - Parses video array and creates `Page` entries with mp4 URLs
  - Image pages are offset by video count for correct indexing
- Added `import org.jsoup.Jsoup` for parsing HTML from string

## Testing

- Verified HTML structure on live XChina pages with video content (e.g., 371P+9V works)
- Video URL extraction tested against real page source
- Non-video pages (no `videos` array) fall back to original image-only behavior

## Notes

- This does NOT embed a video player in the reader — it uses the existing "Open externally" menu
- Requires Mihon reader (Tachiyomi forks) with external-open support
- Clean separation: video and image pages don't interfere with each other

### DIFF
--- a/src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNote.kt
+++ b/src/all/yellownote/src/eu/kanade/tachiyomi/extension/all/yellownote/YellowNote.kt
@@ -22,6 +22,7 @@ import keiyoushi.utils.tryParse
 import okhttp3.HttpUrl.Companion.toHttpUrl
 import okhttp3.Request
 import okhttp3.Response
+import org.jsoup.Jsoup
 import org.jsoup.nodes.Document
 import org.jsoup.nodes.Element
 import java.text.SimpleDateFormat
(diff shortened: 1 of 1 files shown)


Smol-Ame:
Keiyoushi is a manga extensions repo, we don't do/support anime/video extensions.

PR #16012, opened 2026-05-18
Asmodeus Scans add Wayback Machine support

tigeryu8900:
Asmodeus Scans has apparently been IP-banning those who use third-party apps instead of a web browser.

Recently, I tried to open an asmotoon source only to get a 404 response. I tried a VPN, and it worked until I got a 404 response a few minutes later. I also tried other VPN locations, but they all got IP-banned within minutes.

This change adds support for redirecting requests to the Wayback Machine (web.archive.org) as a way to circumvent this.

Checklist:

### DIFF
--- a/src/en/asmotoon/build.gradle
+++ b/src/en/asmotoon/build.gradle
@@ -3,7 +3,7 @@ ext {
     extClass = '.Asmotoon'
     themePkg = 'keyoapp'
     baseUrl = 'https://asmotoon.com'
-    overrideVersionCode = 0
+    overrideVersionCode = 1
     isNsfw = false
 }
 
--- a/src/en/asmotoon/src/eu/kanade/tachiyomi/extension/en/asmotoon/Asmotoon.kt
+++ b/src/en/asmotoon/src/eu/kanade/tachiyomi/extension/en/asmotoon/Asmotoon.kt
@@ -42,7 +42,7 @@ class Asmotoon :
     val waybackMachineClient: OkHttpClient = super
         .client
         .newBuilder()
-        .addInterceptor(WaybackMachineInterceptor())
+        .addInterceptor(WaybackMachineInterceptor("""\w+://asmotoon\.com/.*""".toRegex()))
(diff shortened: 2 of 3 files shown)


Riztard:
does spoofing header as normal web browser not working?

Smol-Ame:
Extension works just fine on my end currently, connected to a VPN as well. Have you considered that maybe the source just sucks & was having transient server/website issues?

tigeryu8900:
I had the user agent spoofed as my default browser.

I don't think they look at the headers since the Wayback Machine is still able to archive the site despite having a user agent that's obviously a bot.

I think they probably check if resources on the page are loaded. That could explain why the Wayback Machine isn't IP-banned since it archives all resources on the page.

tigeryu8900:
> Extension works just fine on my end currently, connected to a VPN as well. Have you considered that maybe the source just sucks & was having transient server/website issues?

I did consider that. I never encountered this issue while using a browser. Also, once I receive a 404 response, I would consistently get a 404 response for months while changing the IP address to somewhere near would fix it.

I had this problem for a long time now, but back then it rarely happened. Now, it happens within minutes.

I'll try clearing cookies to see if that fixes it.

tigeryu8900:
Update: I think I found the culprit. It's the update mechanism.

When you have a few entries from Asmodeus Scans in your library, and the app updates these entries, there will be a QPS spike that I believe is what causes the IP ban.

I found out that redirecting traffic to the Wayback Machine actually prevents this since those queries are distributed across multiple Wayback Machine crawlers, and the Wayback Machine itself does caching and rate limiting.

Smol-Ame:
Instead of scraping/maintaining a completely separate thing, which relies on both that site to be up & for them to have said scraped whatever chapter/chapter pages, add/increase the rate-limit on the extension to something like `1,4`

tigeryu8900:
Got it, created a new PR.

Just a side note, it's possible to scrape on demand via API, which is what this PR does when a snapshot is too old or missing.

PR #16028, opened 2026-05-19
utils: update preference util

manti-X:
it was part of #10020

### DIFF
--- a/core/src/main/kotlin/keiyoushi/utils/Context.kt
+++ b/core/src/main/kotlin/keiyoushi/utils/Context.kt
@@ -4,5 +4,4 @@ import android.app.Application
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
 
-// This ensures the type token isn't duplicated at each call site
 val applicationContext: Application get() = Injekt.get()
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/keiyoushi/utils/Preferences.kt
```diff
@@ -23,4 +23,7 @@ inline fun HttpSource.getPreferencesLazy(
 /**
  * Returns the [SharedPreferences] associated with passed source id
  */
-fun getPreferences(sourceId: Long): SharedPreferences = Injekt.get<Application>().getSharedPreferences("source_$sourceId", 0x0000)
+fun getPreferences(sourceId: Long): SharedPreferences = application.getSharedPreferences("source_$sourceId", 0x0000)
+
+// This ensures the type token isn't duplicated at each call site
+val application: Application get() = Injekt.get()
```
AwkwardPeak7: maybe move to a new file named Context, and make this applicationContext. we can reuse it for webview calls.

core/src/main/kotlin/keiyoushi/utils/Context.kt
```diff
@@ -0,0 +1,8 @@
+package keiyoushi.utils
+
+import android.app.Application
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
+
+// This ensures the type token isn't duplicated at each call site
```
AwkwardPeak7:
```suggestion
```

AwkwardPeak7 (approved):

PR #16029, opened 2026-05-19
MoeTruyen: Add option use API, decrypt method

FiorenMas:
Checklist:


AwkwardPeak7 (commented):
why a preference? just use one or the other whichever is better. probably api

FiorenMas:
> why a preference? just use one or the other whichever is better. probably api

Because this is not an official API, it's a third-party API created by a member of the team that built the site. The website is temporarily closed, so only this third-party API is working right now. I think when the website opens again, using web scraping will be better.

FiorenMas:
Website is back online. Added image encryption and made minor code changes. Switched to HTML scraping as the default method (reasons above), but kept the API as a fallback in case the website goes down.

AwkwardPeak7 (approved):

PR #16033, opened 2026-05-19
Happymh (ZH): Fix page ordering

### LINKED ISSUE #15915: HappyMH extension: Images are loading out of order / incorrectly spliced
### Source information

Happymh 1.4.21

### Source language

Chinese

### Steps to reproduce

1. Open the HappyMH extension.
2. Select a manga/manhwa 
3. Open a chapter 
4. Scroll through the chapter.
5. Observe that the panels and text bubbles do not line up because the image slices are out of order.

### Expected behavior

The image slices should load and stitch together in the correct sequential order to display a continuous, readable webtoon strip.

### Actual behavior

The images load out of sequence, resulting in cut-off text bubbles and misaligned artwork (see attached screenshot).  This issue is especially obvious and disruptive with webtoon-style (long-strip) series.

### Mihon/Mihon fork version

Mihon 0.19.9

### Android version

Android 16

### Other details

Image

### Acknowledgements

### 

---

disk-iq8:
Closes #15915 

Checklist:

### DIFF
--- a/src/zh/happymh/src/eu/kanade/tachiyomi/extension/zh/happymh/Happymh.kt
+++ b/src/zh/happymh/src/eu/kanade/tachiyomi/extension/zh/happymh/Happymh.kt
@@ -300,8 +300,6 @@ class Happymh :
         val comicId = chapterUrl.pathSegments[0]
         val chapterId = chapterUrl.pathSegments[2]
 
-        val startTime = System.nanoTime()
-
         refreshGACookies("$baseUrl/mangaread/$comicId/$chapterId")
 
         val url = "$baseUrl/v2.0/apis/manga/reading".toHttpUrl().newBuilder()
--- a/src/zh/happymh/src/eu/kanade/tachiyomi/extension/zh/happymh/Happymh.kt
+++ b/src/zh/happymh/src/eu/kanade/tachiyomi/extension/zh/happymh/Happymh.kt
@@ -1,5 +1,11 @@
 package eu.kanade.tachiyomi.extension.zh.happymh
 
+import android.app.Application
+import android.os.Handler
+import android.os.Looper
+import android.webkit.CookieManager
+import android.webkit.WebView
+import android.webkit.WebViewClient
 import android.widget.Toast
 import androidx.preference.EditTextPreference
 import androidx.preference.PreferenceScreen
@@ -18,6 +24,7 @@ import eu.kanade.tachiyomi.source.model.SManga
 import eu.kanade.tachiyomi.source.online.HttpSource
 import eu.kanade.tachiyomi.util.asJsoup
 import keiyoushi.utils.getPreferences
+import keiyoushi.utils.parseAs
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.decodeFromStream
 import okhttp3.FormBody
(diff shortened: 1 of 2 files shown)


src/zh/happymh/src/eu/kanade/tachiyomi/extension/zh/happymh/Happymh.kt
```diff
@@ -243,6 +299,8 @@ class Happymh :
         val comicId = chapterUrl.pathSegments[0]
         val chapterId = chapterUrl.pathSegments[2]
 
+        refreshGACookies("$baseUrl/mangaread/$comicId/$chapterId")
```
AwkwardPeak7, line 302: won't the cookie be set by just calling with okhttp?

disk-iq8: webview is a must because of google crap, it's locally generated cookie, see https://github.com/keiyoushi/extensions-source/issues/15915#issuecomment-4486872389 for the details, the function call needs fresh cookies so it's called on each time, the cookie updates implicitly so it's not the issue here. unless you're referring to something else

AwkwardPeak7: Ah I see, I would say prefer using a non webview solution if possible. If it breaks again/often then we can use webview

disk-iq8: I think it's not worth it

according to https://www.googletagmanager.com/gtag/js?id=G-HVJMXGJXFJ

we would need to forge a local cookie, but that alone won't do, so submit it to google servers too as 

```
POST https://analytics.google.com/g/collect 

with some params party, would need trial and error rule out unneeded ones

| v | 2 |
| tid | G-E2G2LX2SKZ |
| gtm | 45je65f1v879443488za200zd879443488 |
| _p | 1779173840925 |
| gcd | 13l3l3l3l1l1 |
| npa | 0 |
| dma | 0 |
| _eu | EAAAAGQ |
| cid | [random].[random] |
| frm | 0 |
| ir | 1 |
| pscdl | noapi |
| rcb | 9 |
| sr | 1216x684 |
| ul | en-us |
| gaf | 2 |
| _s | 1 |
| tag_exp | 0~115616986~115938466~115938469 |
| sid | 1779173435 |
| sct | 1 |
| seg | 1 |
| dl | https://m.happymh.com/user/login |
| dr | https://m.happymh.com/mangaread/chaxianguan/5287383 |
| dt | 登录页————嗨皮漫画，HAPPY漫画 |
| en | page_view |
| _ee | 1 |
| ep.page_group | user |
| tfd | 13365 |

```

then set the custom cookie header without killing off the cf_clearance from main webview 

seems a little redundant to me as webview takes a 500ms or a second since gtag is called early on

lamaxama: You inspired me. This cookie isn’t actually generated by GA—it’s generated by the site dev using some kind of encryption (with super deceptive name). 
Then it gets validated on the chapter detail api: if it matches the encryption rules, it returns the correct page order; otherwise, it returns a scrambled order.

Using a webview can trigger the generation of this cookie. This explains why opening the page once in a webview gives you the correct order of pages.

I'll submit a PR to generate this cookie directly. But the cookie generation algorithm might change soon — if it does, I'd suggest just using a webview to solve it next time.

lamaxama: https://github.com/keiyoushi/extensions-source/pull/16068

disk-iq8: heh, devious naming, I did attempt forging one back then but got ignored, so I didn't delve deep into why server refused it, natural assumption pointed to google side and their backend interaction, but it was just piggy backing on something else

@AwkwardPeak7 you may close this pull, thank you

PR #16096, opened 2026-05-22
Add list-returning extraction methods for Next.js parsing

bapeey:
Checklist:


core/src/main/kotlin/keiyoushi/utils/NextJs.kt
Starmania, line 32:
```suggestion
/** 
 * @return true if caller should stop recursing (only when firstOnly && match found)
 */
private fun <T> extractValueNextJs(
```
This would make things clearer for when `true` is returned.

Starmania (commented):
Good idea

core/src/main/kotlin/keiyoushi/utils/NextJs.kt
```diff
@@ -18,24 +18,38 @@ import kotlin.reflect.typeOf
...
     }
     val children: Iterable<JsonElement> = when (payload) {
         is JsonObject -> payload.values
         is JsonArray -> payload
     }
     for (child in children) {
-        val result = extractValueNextJs(child, predicate, deserializer)
-        if (result != null) return result
+        if (extractValueNextJs(child, predicate, deserializer, results, firstOnly)) {
+            return true
```
AwkwardPeak7, line 50: wouldn't this return true on first child?

bapeey: `true` can only be returned when `firstOnly` is true, so it shouldn't be a problem

PR #16153, opened 2026-05-24
Update base URL for Team Lanh Lung extension

minhnh10x:
Change the base URL from nhalung.com to lunghihi.icu


src/vi/cmanga/src/eu/kanade/tachiyomi/extension/vi/cmanga/CManga.kt
FiorenMas, line 1: Why deleted CManga?

minhnh10x: Because the website has stopped working

FiorenMas: Make another PR

PR #16227, opened 2026-05-26
Comix: Add fallback handling for image and chapter failures

Chronosmage-alt:
Checklist:

Hey, my first PR, so there are no doubt things I could improve.

This PR allows duplicate chapters to be used as fallbacks if an image URL returns a 404 while reading or downloading.

If all duplicate chapter variants contain dead links for the same image, the fallback will still fail and return a 404 as normal.

I cleaned it up so that the fallback only triggers after an actual 404 response, rather than running on every request to reduce the load on the source

This is probably more of a resilience feature than a direct fix for the 404 errors, but it has been working reliably for me during testing last night and this morning

Let me know what you think. And if you can do the same thing with less code, please lmk :)

Yes, I've tested on Mihon, and it works well and works with Suwa aswell

### DIFF
--- a/src/en/comix/build.gradle
+++ b/src/en/comix/build.gradle
@@ -1,7 +1,7 @@
 ext {
     extName = 'Comix'
     extClass = '.Comix'
-    extVersionCode = 24
+    extVersionCode = 25
     isNsfw = true
 }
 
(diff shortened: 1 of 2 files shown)


Chronosmage-alt:
Been using this for a few days on both mihon and suwa. It's working well. While the site is still having some issues it's fixing about 70% of the issue. But of course it can't solve multiple same dead image links

Chronosmage-alt:
Updated to match the current build, tested and is working for me

Chronosmage-alt:
@AwkwardPeak7 Would you mind taking a look at this when you have some time? 

Thanks

Chronosmage-alt:
Gonna close this PR As i think its too heavy, looks like comix are also aware of crawlers so i dont want to make it too heavy that they end up trying to block us again

PR #16285, opened 2026-05-28
Yorai (EN): Fix api changes

### LINKED ISSUE #15867: Yorai: `HTTP 404` error
### Source information

Yorai

### Source language

English

### Steps to reproduce

1.open extension
2.shows error 404
3.checked webview 
4.shows and reading on webview is normal

Image

Image

### Expected behavior

To be able to read normally

### Actual behavior

Http 404

### Mihon/Mihon fork version

0.19.9

### Android version

Android 14

### Other details

_No response_

### Acknowledgements

### 

---

disk-iq8:
Closes #15867

Checklist:

### DIFF
--- a/src/en/yorai/src/eu/kanade/tachiyomi/extension/en/yorai/Yorai.kt
+++ b/src/en/yorai/src/eu/kanade/tachiyomi/extension/en/yorai/Yorai.kt
@@ -76,6 +76,13 @@ class Yorai : HttpSource() {
         return GET(url, headers)
     }
 
+    override fun getMangaUrl(manga: SManga): String = "$baseUrl/comic/${manga.url}"
+
+    override fun getChapterUrl(chapter: SChapter): String {
+        val (slug, number) = chapter.url.split("|")
+        return "$baseUrl/comic/$slug/chapter/$number"
+    }
+
     override fun searchMangaParse(response: Response): MangasPage {
         val data = response.parseAs<Browse>()
 
@@ -84,7 +91,7 @@ class Yorai : HttpSource() {
             return MangasPage(
                 comics.map {
                     SManga.create().apply {
-                        url = "/comic/${it.slug}"
+                        url = it.slug
                         title = it.title
                         thumbnail_url = baseUrl + it.coverUrl
                     }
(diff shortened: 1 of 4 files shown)


src/en/yorai/src/eu/kanade/tachiyomi/extension/en/yorai/Yorai.kt
```diff
@@ -87,13 +80,13 @@ class Yorai :
...
 
         with(data) {
-            val hasNextPage = pagination.page <= pagination.totalPages
+            val hasNextPage = page <= totalPages
             return MangasPage(
-                series.map {
+                comics.map {
                     SManga.create().apply {
-                        url = "/manga/${it.id}"
+                        url = "/comic/${it.slug}"
```
bapeey: Since users will already have to migrate, save only the slug so we can just change the path in request methods next time; same for chapters

disk-iq8: I wanted to avoid migration but couldn't find any hacky way to map comicId -> slug, a bunch of those sites seems to switch to `rsc: 1` GET on most endpoints, which makes the mangaDetails messy to parse too

tho about the slug-only saving, I think it's not that needed right, since we can make a small helper method to sanitize it if it ever changes again, besides the webview will be broken until you prepend the path back to slug with that get manga url function thingy which I'm too lazy to remember

bapeey: It would still be nice to avoid having to sanitize it later though. Also, if the `id` and the `slug` are the same, you could just do it now to avoid the migration, up to you ig

disk-iq8: Had no idea as I never used the extension so went with the safe choice, I see both wayback machine and google were useless 

but Bing takes some rare W on this one, I could find this url yorai.io/manga/3b130157-66e4-4239-95d6-52cdc9106ecc, which is some incompatible uuid thing

Anyway, thanks for your review I'll do the slug storage a little later

AwkwardPeak7 (approved):

PR #16471, opened 2026-06-05
Update CONTRIBUTING.md with URL manipulation guidelines

darwin-256:
Clarified guidelines for URL manipulation and deprecated practices.


Exinaus:
I’ve seen some use of `HttpUrl` that probably should be avoided (or is it my preference?), where people take url (saved like: `/path/$id/`), convert it to `HttpUrl` just to get slug with `pathSegments()`. And convert it back to string for request.

That's real example from code, i didn't wrote it just to illustrate things:
```
    override fun pageListRequest(chapter: SChapter): Request {
        val chapterId = "$baseUrl${chapter.url}".toHttpUrl().pathSegments.last()
        val url = "$API_URL/chapters/$chapterId"
        return GET(url, headers)
    }
```
Is it wrong to prefer split("/") or substringAfterLast there?

Or another real example:

```
    private fun buildImageUrl(imageId: String): String {
        val url = "$IMAGE_STORAGE_URL/$imageId".toHttpUrl().newBuilder()
            .addQueryParameter("optimizer", "image")
            .addQueryParameter("width", "560")
            .addQueryParameter("quality", "70")
            .addQueryParameter("height", "auto")
            .build()
        return url.toString()
    }
```

Maybe some specifics needed where to use `HttpUrl`, or it will be used everywhere. Or those examples are fine?

AwkwardPeak7 (approved):

AwkwardPeak7:
@Exinaus the idea is that if url contains other things e.g as part of fragment, blindly splitting string can include unwanted stuff in the final result. (especially true when using `setUrlWithoutDomain` with source provided urls, we can't know if they had some sort of fragment or query parms)

PR #16513, opened 2026-06-07
RateLimit implementation

AwkwardPeak7:
- Sliding-window rate limiter as an OkHttp interceptor
- Each RateLimit defines a permit budget over a time window, with an optional minimum dispatch interval to smooth bursts
- Rules are evaluated in order, first match wins — more specific rules should come before broader ones
- `rateLimit` call can be chained, adding multiple rules to same interceptor
- `NetworkInterceptor` so cached response is not waited upon

Why our own implementation?
 - [Upstream is deprecating](https://github.com/mihonapp/tachiyomix/blob/d4936afa6481f8edea10cd4902a97d4ea5090202/library/src/main/java/eu/kanade/tachiyomi/network/interceptor/RateLimitInterceptor.kt#L20-L25)
 - Gives us more control over the ratelimit implementation

### DIFF
--- a/core/src/main/kotlin/keiyoushi/network/RateLimit.kt
+++ b/core/src/main/kotlin/keiyoushi/network/RateLimit.kt
@@ -78,16 +78,16 @@ private class RateLimitInterceptor : Interceptor {
 
         if (call.isCanceled()) throw IOException("Canceled")
 
-        val matched = rateLimits.firstOrNull { it.shouldLimit(request.url) }
+        val rateLimit = rateLimits.firstOrNull { it.shouldLimit(request.url) }
             ?: return chain.proceed(request)
 
-        val id = matched.acquireSlot(call)
+        val id = rateLimit.acquireSlot(call)
         try {
             val response = chain.proceed(request)
-            if (response.networkResponse == null) matched.releaseSlot(id)
+            if (response.networkResponse == null) rateLimit.releaseSlot(id)
             return response
         } catch (e: Exception) {
-            matched.releaseSlot(id)
+            rateLimit.releaseSlot(id)
             throw e
         }
     }
--- a/core/src/main/kotlin/keiyoushi/network/RateLimit.kt
+++ b/core/src/main/kotlin/keiyoushi/network/RateLimit.kt
@@ -77,7 +77,7 @@ fun OkHttpClient.Builder.rateLimit(
     period: Duration = 1.seconds,
     interval: Duration = Duration.ZERO,
     shouldLimit: (HttpUrl) -> Boolean = { true },
-): RateLimitBuilder = RateLimitBuilder(this).rateLimit(permits, period, interval, shouldLimit)
+): RateLimitBuilder = RateLimitBuilder(this, listOf(RateLimitRule(permits, period, interval, shouldLimit)))
 
 internal class RateLimitRule(
     val permits: Int,
--- a/core/src/main/kotlin/keiyoushi/network/RateLimit.kt
+++ b/core/src/main/kotlin/keiyoushi/network/RateLimit.kt
@@ -8,6 +8,7 @@ import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.Response
 import java.io.IOException
+import java.util.ArrayDeque
 import java.util.concurrent.locks.Condition
 import java.util.concurrent.locks.ReentrantLock
 import kotlin.concurrent.withLock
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/keiyoushi/network/RateLimit.kt
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        val ts = SystemClock.elapsedRealtime()
+        queue.addLast(ts)
+        lastDispatchTime = ts
+        ts
+    }
+
+    private fun RateLimit.releaseSlot(timestamp: Long): Unit = lock.withLock {
+        if (queue.isEmpty() || timestamp < queue.first()) return
```
vetleledaal: Is this guard correct? Check if threading issues could crop up.

core/src/main/kotlin/keiyoushi/network/RateLimit.kt
```diff
@@ -0,0 +1,122 @@
...
+        queue.removeFirstOccurrence(timestamp)
```
vetleledaal: Will collisions cause issues?

core/src/main/kotlin/keiyoushi/network/RateLimit.kt
```diff
@@ -0,0 +1,122 @@
...
+
+            val now = SystemClock.elapsedRealtime()
+            while (queue.isNotEmpty() && queue.first() <= now - periodMillis) queue.removeFirst()
+
+            val windowWait = if (queue.size < permits) 0L else queue.first() - (now - periodMillis)
+            val intervalWait = lastDispatchTime + intervalMillis - now
+            val waitTime = maxOf(windowWait, intervalWait, 0L)
+
+            if (waitTime == 0L) break
+            retryCondition.awaitNanos(waitTime * 1_000_000L)
```
vetleledaal: ?
```suggestion
            retryCondition.awaitNanos(waitTime.inWholeNanoseconds)
```

vetleledaal: Consider using `Duration` here too.

Comment is on the wrong line- meant for `windowWait`, `intervalWait`, `waitTime`.

core/src/main/kotlin/keiyoushi/network/RateLimit.kt
AwkwardPeak7: `releaseSlot` runs everything in a `lock` (ReentrantLock) so no threading issue

core/src/main/kotlin/keiyoushi/network/RateLimit.kt
AwkwardPeak7: `SystemClock.elapsedRealtime()` returns `Long`

core/src/main/kotlin/keiyoushi/network/RateLimit.kt
AwkwardPeak7: it could yes, tho two requests sharing same timestamp down to milliseconds is rare. A solution can be to store an id/index instead of timestamp 🤔

core/src/main/kotlin/keiyoushi/network/RateLimit.kt
vetleledaal: I mean, you could do the calculation with `Duration` and convert to `Long` afterwards.

AwkwardPeak7: did that

vetleledaal (dismissed):
Not 100% sure on splitting the array into timestamps/ids. Only suggestion would be using `Pair`, but not sure.

vetleledaal (dismissed):

core/src/main/kotlin/keiyoushi/network/RateLimit.kt
```diff
@@ -0,0 +1,171 @@
...
+ * @param interval    Minimum gap between consecutive dispatches. Smooths bursts —
+ *                    `permits=10, interval=100.ms` allows 10/s but spaced ≥100ms apart.
+ * @param shouldLimit Whether this rule applies to a given URL.
+ */
+fun OkHttpClient.Builder.rateLimit(
+    permits: Int,
+    period: Duration = 1.seconds,
+    interval: Duration = Duration.ZERO,
+    shouldLimit: (HttpUrl) -> Boolean = { true },
+): RateLimitBuilder = RateLimitBuilder(this).rateLimit(permits, period, interval, shouldLimit)
```
FourTOne5:
```suggestion
): RateLimitBuilder = RateLimitBuilder(this, listOf(RateLimitRule(permits, period, interval, shouldLimit)))
```

FourTOne5 (dismissed):

PR #16556, opened 2026-06-08
fix(japscan): potential fix for honeypot chapters showing wrong number/URL

### LINKED ISSUE #1: Convention on minor variable changes
### Describe why this should be added

In my implementation of the Bato.To fix, I have a minor change to the extension that changes the name of two variables.
In this case, it's changing "imgHttpsLis" to "imgHttps", and "imgHttpsLisString" to "imgHttpsString".
This is minor, but when the source changes its tag's name, the variable describing that name should also be changed.

### Other details

_No response_

### Acknowledgements

WardenPro:
## Problem observed

When browsing a manga chapter list (tested on *Dragon Ball Kakusei*), some chapters display the wrong number and link to a wrong URL that returns 404.

Example: chapter 1 is displayed as **「Chapitre 000005」** and its URL is `/manga/dragon-ball-kakumei/000005/` instead of `/1/`.

## Root cause (best guess — not 100% certain)

Japscan injects honeypot elements into each chapter row. Looking at the DOM, those honeypots use this inline style to stay invisible to readers:

```html
<a style="position:absolute;clip-path:circle(0);overflow:hidden" fyex="/manga/.../000005/">Chapitre 000005</a>
```

`isHidden()` currently catches `clip-path:inset(100%` but **not** `clip-path:circle(0)`, so these elements slip past the hidden-element filter and get picked up instead of the real chapter `<div>`.

Additionally, the number comparison (`chapterNum == urlNum`) passes for these honeypots because both the text **and** the URL share the same zero-padded number (`"000005" == "000005"`), so the existing slug+number heuristic doesn't reject them.

## Changes

1. **`clip-path:circle(0)` added to `HIDDEN_STYLE_TOKENS`** — the most direct fix; this style is the actual hiding mechanism used on the observed honeypots.

2. **Reject zero-padded chapter URL numbers** — real chapter URLs use plain integers (`/1/`, `/1181/`); zero-padded segments like `/000005/` only appear in honeypots, so we explicitly reject them.

## Caveats

- I'm not fully certain this covers all honeypot variants Japscan uses — the site may rotate its obfuscation techniques.
- I haven't been able to build and run the extension locally to confirm; the analysis is based on inspecting the live DOM and reading the source.
- Fix #2 (zero-padding check) is a secondary safety net — fix #1 should be sufficient on its own for the observed pattern.

Happy to adjust or revert either change based on feedback from maintainers who have better visibility into Japscan's anti-scraping patterns.

### DIFF
--- a/src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
+++ b/src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
@@ -59,7 +59,7 @@ class Japscan :
     // Sometimes an adblock blocker will pop up, preventing the user from opening
     // a cloudflare protected page
     private val internalBaseUrl = "https://www.japscan.foo"
-    override val baseUrl = internalBaseUrl
+    override val baseUrl = "$internalBaseUrl/mangas/?sort=popular&p=1"
 
     override val lang = "fr"
 
--- a/src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
+++ b/src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
@@ -642,6 +642,10 @@ class Japscan :
         return Observable.just(images)
     }
 
+    override fun pageListParse(response: Response): List<Page> = throw UnsupportedOperationException("Not used")
+
+    override fun imageUrlParse(response: Response): String = ""
+
     // Filters
     private class TextField(name: String) : Filter.Text(name)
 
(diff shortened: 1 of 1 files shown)


darwin-256:
Could you possibly convert from ParsedHttpSource to HttpSource?

WardenPro:
Sure, no problem! Converting to `HttpSource` now.

src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -60,7 +59,7 @@ class Japscan :
     // Sometimes an adblock blocker will pop up, preventing the user from opening
     // a cloudflare protected page
     private val internalBaseUrl = "https://www.japscan.foo"
-    override val baseUrl = "$internalBaseUrl/mangas/?sort=popular&p=1"
+    override val baseUrl = internalBaseUrl
```
AwkwardPeak7: why remove?

WardenPro: No strong reason — it was an unnecessary cleanup on my part. Reverting it to keep the original value.

AwkwardPeak7 (approved):

PR #16748, opened 2026-06-16
MangaDotNet: Use the "Relevance" sort order by default when searching

KenzKD:
Previously it would sort the search results by Latest Update

Checklist:


Smol-Ame:
Why did you create another branch/PR instead of just pulling upstream on the existing branch/PR?

Alpha-782:
I don't know if you're aware... but "Relevance" and "Latest Updates" are the same. There is no "Relevance" on the site, so it uses default, which is "Latest Updates".

KenzKD:
> I don't know if you're aware... but "Relevance" and "Latest Updates" are the same. There is no "Relevance" on the site, so it uses default, which is "Latest Updates".

Maybe it didn't have one initially but it is there now:

image

If I switch to "Latest" I get these results which are clear not as good as the "Relevance" sort order:

image

KenzKD:
> Why did you create another branch/PR instead of just pulling upstream on the existing branch/PR?

I am not that use to dealing with PRs since the work I do, usually just has me working on repos that only I use.

I had rebased the branch not realizing the mess it would lead to in the PR. So I deleted the old branch and made another one to get rid of the mess.

Alpha-782:
You're right, that filter exists when there is a query. It doesn't appear by default.

KenzKD:
> You're right, that filter exists when there is a query. It doesn't appear by default.

I think I get the confusion. I was directly typing into the search bar in the top of the website, which already prevent you from just leaving it blank.
image

Only if you open the "Advanced Search" directly does it open with "Latest" as the default. 

If we are searching anything we will have a query anyways at least on Komikku, so it would be better to have "Relevance" as the default sort order when searching.

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
```diff
@@ -254,12 +254,10 @@ class Mangadotnet :
             addQueryParameter("page", page.toString())
 
             filters.firstInstanceOrNull<SortFilter>()?.also { filter ->
-                if (filter.sort == "relevance" && query.isBlank()) {
-                    addQueryParameter("sortBy", "latest")
-                } else {
-                    addQueryParameter("sortBy", filter.sort)
+                if (filter.sort.isNotBlank()) {
+                    addQueryParameter("sort", filter.sort)
+                    addQueryParameter("sortOrder", if (filter.ascending) "asc" else "desc")
```
AwkwardPeak7: keep logic of if query is blank, and relevance is selected, then set latest; since site don't show relevance filter for no query.

KenzKD: I have now reverted back to the previous logic and also fixed checking for when relevance was selected.

It seems that `filter.sort == "relevance"` was always returning false and had to be changed to `filter.sort == "" `. By luck, it seemed that MangaDotNet would just send the latest chapters to the extension if the query was blank.

Now, it should properly check for when relevance is selected and the query is blank.

KenzKD:
Closed in favour of #16896

PR #16874, opened 2026-06-21
Add support for ext-lib 1.6 based extensions

AwkwardPeak7:
Pr open for comments.

1.6 extensions are expected to implement `KeiSource` instead of `HttpSource` directly.


core/build.gradle.kts
```diff
@@ -20,6 +20,10 @@ android {
 
 dependencies {
     compileOnly(libs.bundles.common)
+    compileOnly(libs.tachiyomi.lib.v16)
+    compileOnly(libs.okhttp.brotli)
+    compileOnly(libs.okhttp.logging)
+    compileOnly(libs.okhttp.dnsOverHttps)
```
lamaxama: The dnsOverHttps is unused and can be removed.

core/src/main/kotlin/keiyoushi/source/KeiSource.kt
```diff
@@ -0,0 +1,416 @@
...
+     * Customizes the base Headers builder. Subclasses can override this to append source-specific
+     * headers.
+     *
+     * Note: "User-Agent", "Referer", and "Origin" are already added by default.
+     */
+    protected open fun Headers.Builder.configureHeaders(): Headers.Builder = this
+
+    final override fun headersBuilder() = super.headersBuilder()
+        .set("Referer", "$baseUrl/")
+        .set("Origin", baseUrl)
```
hualiong, line 144: Perhaps `HttpSource#getHomeUrl` should be used instead of `baseUrl` to handle websites with frontend-backend separation, where the API host may differ from the `baseUrl` ?

I'm not sure if the return value of `getHomeUrl` is the same as `baseUrl`, both without the trailing slash, so this is just a quesion.

AwkwardPeak7: to my understanding, `getHomeUrl` is supposed to be full url that should be opened when opening webview from browse screen. This may include path to specific subpage, e.g the listing page with view query parm for example, as many sites don't have cloduflare protection on home page but on hot paths that extensions usually use.

core/src/main/kotlin/keiyoushi/source/KeiSource.kt
```diff
@@ -0,0 +1,416 @@
...
+            triggerBackgroundFilterFetch()
+        }
+
+        if (parsed != null) return parsed
+
+        val hint = if (filterFetchAttemptCount.get() >= maxFilterFetchAttempts) {
+            Filter.Header("Failed to fetch filters, restart the app to retry")
+        } else {
+            Filter.Header("Press 'Reset' to fetch more filters")
+        }
```
hualiong: For users who use other languages, would using English prompts be less user-friendly?

AwkwardPeak7: yes localization needs to be added. I am just not sure what is the best way to go about it

AwkwardPeak7: suggest translations (or improvements)

core/src/main/kotlin/keiyoushi/source/KeiSource.kt
```diff
@@ -0,0 +1,435 @@
...
+
+        return FilterList(
+            buildList {
+                addAll(getFilterList(data = null))
+                if (isNotEmpty()) {
+                    add(Filter.Separator())
+                }
+                Filter.Header(
+                    when (lang) {
+                        else -> "Press 'Reset' to load filters"
```
bapeey:
```suggestion
                        "es", "es-419" -> "Pulse 'Restablecer' para cargar los filtros"
                        else -> "Press 'Reset' to load filters"
```

core/src/main/kotlin/keiyoushi/source/KeiSource.kt
```diff
@@ -0,0 +1,436 @@
...
+        return FilterList(
+            buildList {
+                addAll(getFilterList(data = null))
+                if (isNotEmpty()) {
+                    add(Filter.Separator())
+                }
+                Filter.Header(
+                    when (lang) {
+                        "es", "es-419" -> "Pulse 'Restablecer' para cargar los filtros"
+                        else -> "Press 'Reset' to load filters"
```
manti-X:
```suggestion
                        else -> "Tap 'Reset' to load filters"
```
after looking into it more, I’d prefer "Tap", since most users are on mobile

https://developer.apple.com/design/human-interface-guidelines/writing#:~:text=Write%20for%20how,you%20mean%20%E2%80%9Ctap.%E2%80%9D

core/src/main/kotlin/keiyoushi/source/KeiSource.kt
```diff
@@ -0,0 +1,436 @@
...
+        if (parsed != null) return parsed
+
+        return FilterList(
+            buildList {
+                addAll(getFilterList(data = null))
+                if (isNotEmpty()) {
+                    add(Filter.Separator())
+                }
+                Filter.Header(
+                    when (lang) {
```
manti-X:
```suggestion
                    when (lang) {
                        "de" -> "Tippe auf „Zurücksetzen“, um die Filter zu laden"
                        "ja" -> "「リセット」をタップしてフィルターを読み込む"
```

core/src/main/kotlin/keiyoushi/source/KeiSource.kt
```diff
@@ -0,0 +1,438 @@
...
```
Exinaus:
```suggestion
                Filter.Header(
                    when (lang) {
                        "ru" -> "Нажмите «Сбросить», чтобы загрузить фильтры"
                        "uk" -> "Натисніть «Скинути», щоб завантажити фільтри"
```

core/src/main/kotlin/keiyoushi/source/KeiSource.kt
```diff
@@ -0,0 +1,438 @@
...
+                        "de" -> "Tippe auf „Zurücksetzen“, um die Filter zu laden"
```
xMohnad:
```suggestion
                        "de" -> "Tippe auf „Zurücksetzen“, um die Filter zu laden"
                        "ar" -> "اضغط على 'إعادة تعيين' لتحميل الفلاتر"
```

gradle/kei.versions.toml
```diff
@@ -1,6 +1,6 @@
 [versions]
 # Used in [/gradle/build-config/src/main/kotlin/PluginKotlinMultiplatform.kt]
-android-sdk-min = "21"
+android-sdk-min = "26"
```
FourTOne5: do 26 for 1.6 extension. keep 21 for 1.4

core/src/main/kotlin/keiyoushi/source/KeiSource.kt
```diff
@@ -0,0 +1,456 @@
...
+    ): SMangaUpdate {
+        check(fetchDetails || fetchChapters) { "getMangaUpdate called with nothing to fetch (fetchDetails=false, fetchChapters=false)" }
+
+        check(updatesInFlight.add(manga.url)) { "getMangaUpdate must not be called concurrently for same manga" }
+
+        try {
+            return fetchMangaUpdate(manga, chapters, fetchDetails, fetchChapters)
+        } finally {
+            updatesInFlight.remove(manga.url)
+        }
```
FourTOne5: instead of failing the 2nd+ request we should return the response back to all the request

FourTOne5 (changes requested):

core/src/main/kotlin/keiyoushi/source/KeiSource.kt
AwkwardPeak7: the parallel request could be of either manga or chapters (parallel), and apps should not call it like that. so hard throw

FourTOne5: fair enough

gradle/kei.versions.toml
AwkwardPeak7: should we really complicate it like this?

FourTOne5: in that case no need to wait for 1.6. bump it now

FourTOne5 (dismissed):

FourTOne5 (dismissed):
i'll undraft and merge smh

AwkwardPeak7:
no, I am adjusting it a bit

FourTOne5:
image

FourTOne5 (approved):

PR #16876, opened 2026-06-21
Madokami [EN]: Fix Entries

Alpha-782:
In suggestions tab for apps like Komikku, archives were appearing as entries when they were not, which this PR addresses.

Used `SUPPORTED_ARCHIVE_TYPES` from Mihon source model to filter supported archives out.

Checklist:


src/en/madokami/src/eu/kanade/tachiyomi/extension/en/madokami/Madokami.kt
```diff
@@ -207,4 +212,11 @@ class Madokami :
         screen.addPreference(username)
         screen.addPreference(password)
     }
+
+    private fun isArchiveUrl(url: String): Boolean {
+        val path = url.substringBefore("?").substringBefore("#").lowercase(Locale.ROOT)
+        return path.endsWith(".zip") || path.endsWith(".cbz") || path.endsWith(".rar") ||
```
Rafou376: Why not do something like that?

```kt
val exts = listOf(".zip", ".cbz", ".rar", ".cbr", ".7z", ".cb7", ".tar", ".cbt")
return exts.any { path.endsWith(it) }
```

Alpha-782: I guess that works, too.

AwkwardPeak7 (approved):

PR #16891, opened 2026-06-22
Mangadotnet [EN]: Update Search

Alpha-782:
Site changed how search shows results, and blur 18+ entries. Logic is:
- Not blurworthy -> No 18+
- Blurworthy -> 18+ Only

The preference for both (No 18+ & 18+ Only) joins the results to match current website behavior now.

No bump as this should be merged with #16748.

Checklist:

- [ ] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [ ] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [ ] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [ ] Have removed `web_hi_res_512.png` when adding a new extension
- [x] This PR is AI-assisted, I have reviewed the changes manually and confirmed they are not slop

PR #16896, opened 2026-06-22
Mangadotnet (EN): update filters and general cleanup

Rafou376:
Supersedes #16891
Supersedes #16748
Supersedes #16906

Checklist:


src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
```diff
@@ -333,9 +349,18 @@ class Mangadotnet :
 
     override fun searchMangaParse(response: Response): MangasPage {
         val data = response.decodeRscAs<Data<MangaList>>().data
-        updateGenres(data.allGenres)
+        updateGenres(data.allGenres, true)
 
-        return MangasPage(data.mangaList.orEmpty().map { it.toSManga(baseUrl) }, data.hasNextPage())
+        val mode = adultModePref()
+        val filteredManga = data.mangaList.orEmpty().filter { manga ->
```
AwkwardPeak7: remove this preference if it requires filtering client side. client side filtering can lead to all results being filtered out, and app will not try next pages.

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
```diff
@@ -112,6 +113,14 @@ class Mangadotnet :
         if (getString(POPULAR_MODE_PREF, null) == "trending") {
             edit().putString(POPULAR_MODE_PREF, "most-tracked").apply()
         }
+
+        val cacheVersionKey = "pref_genre_cache_v2"
+        if (!contains(cacheVersionKey)) {
+            val cacheDir = Injekt.get<Application>().cacheDir.resolve("source_$id")
```
AwkwardPeak7:
```suggestion
            val cacheDir = applicationContext.cacheDir.resolve("source_$id")
```

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
```diff
@@ -112,6 +113,14 @@ class Mangadotnet :
         if (getString(POPULAR_MODE_PREF, null) == "trending") {
             edit().putString(POPULAR_MODE_PREF, "most-tracked").apply()
         }
+
+        val cacheVersionKey = "pref_genre_cache_v2"
+        if (!contains(cacheVersionKey)) {
```
AwkwardPeak7: why?

AwkwardPeak7 (changes requested):

Alpha-782:
18+ entries, blurred when using search, now appear regardless of the 18+ preference. It's why the client side filtering existed. 

If there's a better way to filter them out, that'd be great, as search shows all results on site, but blurs 18+, which Mihon apps cannot do, hence client-side filtering.

I talked a bit about this on my closed PR, but this website behavior change only happens for search. Popular/Latest still respect the 18+ preference.

Rafou376:
> 18+ entries, blurred when using search, now appear regardless of the 18+ preference. It's why the client side filtering existed.
> 
> If there's a better way to filter them out, that'd be great, as search shows all results on site, but blurs 18+, which Mihon apps cannot do, hence client-side filtering.
> 
> I talked a bit about this on my closed PR, but this website behavior change only happens for search. Popular/Latest still respect the 18+ preference.

Apart from using a null value in `toSManga` for 18+ covers, I don't really see what solutions could be applied

Alpha-782:
An `if` condition?

If `is_blurworthy` value is `0`, show them for `No 18+`.
elif `is_blurworthy` value is `true` (this value might be reversed, I forgot), show it for `18+ Only`.
Else, no filtering (for `No 18+ & 18+ Only`).

My train of thought: 
- When using "No 18+" preference, only show is_blurworthy (value of 0) in the search results,
- When using "18+ Only" preference, only show is_blurworthy (value is true) in the search results,
- When using  "No 18+ & 18+ Only" preference, no filtering, as it would match website behavior (and current changes as well). 

Would this not turn out well in practice? Something similar to below:

**Note:** To further reduce chances of pagination issues, we can use the new "cards per page" the website uses currently (we have the choice between 28, 56 and 84, which would immensely help with the `depth` approach). A normal 84 card page is around 65KB, which should be completely fine.

Adding `perPage` in searchMangaRequest:
```
            addAdultParam()
            if (query.isNotBlank()) {
                addQueryParameter("search", query)
            }
            addQueryParameter("page", page.toString())
            addQueryParameter("perPage", "84")
```

image

```
    // =============================== Search ===============================
    override fun fetchSearchManga(page: Int, query: String, filters: FilterList): Observable<MangasPage> {
        if (query.startsWith("\u200B\u200B")) {
            val name = query.removePrefix("\u200B\u200B")
            val newFilters = filters.apply { firstInstance<ArtistFilter>().state = name }
            return fetchSearchMangaRecursive(page, "", FilterList(newFilters))
        }
        if (query.startsWith("\u200B")) {
            val name = query.removePrefix("\u200B")
            val newFilters = filters.apply { firstInstance<AuthorFilter>().state = name }
            return fetchSearchMangaRecursive(page, "", FilterList(newFilters))
        }
        if (query.startsWith("https://")) {
            val url = query.toHttpUrlOrNull()
            if (
                url != null &&
                (url.host == baseUrl.toHttpUrl().host) &&
                url.pathSize > 1 &&
                (url.pathSegments[0] in listOf("manga", "chapter"))
            ) {
                return if (url.pathSegments[0] == "manga") {
                    val manga = SManga.create().apply {
                        this.url = url.pathSegments[1]
                    }

                    fetchMangaDetails(manga)
                } else {
                    val chapter = SChapter.create().apply {
                        this.url = ChapterUrl(
                            id = url.pathSegments[1],
                            source = url.queryParameter("source") ?: "user",
                            isVolume = url.queryParameter("mode") == "volume",
                        ).toJsonString()
                    }

                    client.newCall(pageListRequest(chapter))
                        .asObservableSuccess()
                        .switchMap {
                            val manga = SManga.create().apply {
                                this.url = it.parseAs<Images>().manga.id.toString()
                            }

                            fetchMangaDetails(manga)
                        }
                }.map {
                    MangasPage(listOf(it), false)
                }
            }

            throw Exception("Unsupported url")
        }

        return fetchSearchMangaRecursive(page, query, filters)
    }

    private fun fetchSearchMangaRecursive(page: Int, query: String, filters: FilterList, depth: Int = 0): Observable<MangasPage> {
        if (depth > 5) {
            return Observable.just(MangasPage(emptyList(), false))
        }

        return client.newCall(searchMangaRequest(page, query, filters))
            .asObservableSuccess()
            .map { response ->
                val data = response.decodeRscAs<Data<MangaList>>().data
                updateGenres(data.allGenres, adultModePref() != "none")

                val mode = adultModePref()
                val filteredManga = data.mangaList.orEmpty().filter { manga ->
                    when (mode) {
                        "none" -> !manga.isAdult
                        "1" -> manga.isAdult
                        else -> true // "both"
                    }
                }.map { it.toSManga(baseUrl) }

                val hasNextPage = data.hasNextPage()

                MangasPage(filteredManga, hasNextPage)
            }
            .concatMap { mangasPage ->
                if (mangasPage.mangas.isEmpty() && mangasPage.hasNextPage) {
                    fetchSearchMangaRecursive(page + 1, query, filters, depth + 1)
                } else {
                    Observable.just(mangasPage)
                }
            }
    }

    override fun searchMangaParse(response: Response): MangasPage = throw UnsupportedOperationException()

**Comment by Seyden** [comment]
Could you add that the default filter reflects the default filter on the page? Right now it uses latest as default in Kei, on the site it uses relevance

**Comment by Alpha-782** [comment]
Already done in [8b3ca74](https://github.com/keiyoushi/extensions-source/pull/16896/commits/8b3ca74c254b843317be9ff27bc7551549d11e67).

File: src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Dto.kt [hunk 4]
```diff
@@ -53,17 +55,30 @@ class BrowseManga(
...
         url = id.toString()
-        title = this@BrowseManga.title
-        thumbnail_url = photo?.let {
-            if (it.startsWith("/")) {
-                baseUrl + it
-            } else if (it.startsWith("http")) {
-                it
-            } else {
-                null
+        title = if (isAdult && hideAdultCovers) "[18+] ${this@BrowseManga.title}" else this@BrowseManga.title
```
AwkwardPeak7: don't change the title

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
```diff
@@ -61,7 +64,22 @@ class Mangadotnet :
     override val baseUrl = "https://mangadot.net"
     override val supportsLatest = true
 
-    override val client = network.client
+    override val client = network.client.newBuilder()
+        .addInterceptor { chain ->
+            val request = chain.request()
+
+            if (request.url.encodedPath == "/nsfw-cover.jpg") {
```
AwkwardPeak7: either leave out the cover (set it to "") or set it to fakeimg.pl with text saying "NSFW"

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
```diff
@@ -112,6 +130,14 @@ class Mangadotnet :
         if (getString(POPULAR_MODE_PREF, null) == "trending") {
             edit().putString(POPULAR_MODE_PREF, "most-tracked").apply()
         }
+
+        val cacheVersionKey = "pref_genre_cache"
+        if (!contains(cacheVersionKey)) {
```
AwkwardPeak7: I meant, why remove genres from cache?

Rafou376: I initially thought it was needed to clear out the demographic tags from the old cache. But looking closer at the code, it filters them on the fly anyway, so in reality, I'm not even sure that keeping the old cache would cause an issue
I'll remove it for now, and @Alpha-782 can confirm if it was actually necessary

Alpha-782: I initially did that change as some users reported that genres were changing position in the filters list, leading to mismatched genre blacklist. You can decide from here, I only did it as precaution.

Alpha-782:
@Rafou376 use `https://fakeimg.ryd.tools/` instead of `https://fakeimg.pl/`. Domain can't be accessed. That domain works fine, on my end.

Rafou376:
> @Rafou376 use `https://fakeimg.ryd.tools/` instead of `https://fakeimg.pl/`. Domain can't be accessed. That domains works fine, on my end.

https://github.com/Rydgel/Fake-images-please/issues/63#issuecomment-2972389633
They haven't updated their doc 😶

Alpha-782:
I was also confused until I saw this comment: https://github.com/Rydgel/Fake-images-please/issues/62#issuecomment-2910739372

AwkwardPeak7 (approved):

PR #16903, opened 2026-06-22
Liliana: filter placeholder SVG pages

ismailoezcan:
## What changed

- Filter Liliana page URLs before creating `Page` entries.
- Ignore placeholder SVG/comment loader URLs such as `loading_comments.svg` so they are not downloaded as manga pages.
- Bump the Liliana `baseVersionCode` so affected extensions receive an update.

## Why

ManhuaPlusOrg can include a non-page placeholder in the chapter image AJAX response:

```text
https://cdn.mangageko.com/mg1/loading_comments.svg
```

That URL is emitted as the final `div.separator[data-index]` entry for some chapters. The downloader then treats it as a manga page and fails while fetching the image because the remote host terminates the TLS handshake.

Filtering the placeholder keeps the page list to actual manga images and avoids failing an otherwise valid chapter download.

## Validation

Ran locally in an Android SDK container:

```text
./gradlew src:en:manhuaplusorg:assembleRelease
./gradlew lib-multisrc:liliana:spotlessCheck src:en:manhuaplusorg:spotlessCheck
```

### DIFF
--- a/lib-multisrc/liliana/build.gradle.kts
+++ b/lib-multisrc/liliana/build.gradle.kts
@@ -4,4 +4,4 @@ plugins {
     alias(kei.plugins.multisrc)
 }
 
-baseVersionCode = 4
+baseVersionCode = 5
--- a/lib-multisrc/liliana/src/eu/kanade/tachiyomi/multisrc/liliana/Liliana.kt
+++ b/lib-multisrc/liliana/src/eu/kanade/tachiyomi/multisrc/liliana/Liliana.kt
@@ -288,16 +288,24 @@ abstract class Liliana(
     }
 
     protected open fun pageListParse(document: Document): List<Page> = if (document.selectFirst("div.separator[data-index]") == null) {
-        document.select("div.separator").mapIndexed { i, page ->
-            val url = page.selectFirst("a")!!.attr("abs:href")
-            Page(i, imageUrl = url)
-        }
+        document.select("div.separator")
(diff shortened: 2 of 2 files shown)


ismailoezcan:
Additional real-world repro example from ManhuaPlusOrg:

Series: **The SS-Rank Freshman at Dawn Academy**

Example chapters:

- https://manhuaplus.org/manga/super-magic-academy-s-ss-rank-freshman/chapter-1
- https://manhuaplus.org/manga/super-magic-academy-s-ss-rank-freshman/chapter-2
- https://manhuaplus.org/manga/super-magic-academy-s-ss-rank-freshman/chapter-3

For these chapters, the AJAX page list endpoint includes the placeholder URL as the last `div.separator[data-index]` page:

```text
https://cdn.mangageko.com/mg1/loading_comments.svg
```

Observed result before this patch:

- chapter 1: real image pages 0-16 load, page 17 is `loading_comments.svg` and fails
- chapter 2: real image pages 0-12 load, page 13 is `loading_comments.svg` and fails
- chapter 3: real image pages 0-17 load, page 18 is `loading_comments.svg` and fails

The failure appears as a download error because the client attempts to fetch the SVG placeholder as a manga page and the remote host terminates the TLS handshake.

A newer chapter from the same series, e.g. chapter 4, uses real `cdn.manhuaplus.cc/*.webp` image URLs only and downloads normally:

- https://manhuaplus.org/manga/super-magic-academy-s-ss-rank-freshman/chapter-4

AwkwardPeak7 (approved):

PR #15946, opened 2026-05-16
add modern gradle config for extensions with build time id generation…

AwkwardPeak7:
… and configurable baseurl preferences

Extension DSL (`build.gradle.kts`)

```kotlin
plugins {
    alias(kei.plugins.extension)
}

keiyoushi {
    // ── Extension metadata ──────────────────────────────────────────
    name = "Example Source"          // required, romanized
    className = "ExampleSource"      // required, PascalCase
    versionCode = 42                 // required, int
    nsfw = false                     // optional, default false
    theme = "madara"                 // optional, multisrc theme name


    // ── Sources ─────────────────────────────────────────────────────
    // At least one `source` block required.
    // Multiple blocks → generated class is a SourceFactory.

    source {
        // ── Identity ────────────────────────────────────────────────
        name = "Example Source"         // required
        lang = "en"                     // required
        id = 1234567890123456789        // optional, auto-generated if omitted
        versionId = 1                   // optional, default 1 (collision disambiguator)


        // ── Preferences ─────────────────────────────────────────────
        configurable = true             // optional, default false
        // Set to true when your source class implements ConfigurableSource.
        // Ensures super.setupPreferenceScreen() is chained after auto-generated prefs.


        // ── Base URL ────────────────────────────────────────────────
        // Choose ONE form:

        // 1. Static URL
        baseUrl("https://example.com")

        // 2. Mirror URLs (adds a ListPreference to settings)
        // baseUrl {
        //     value = "https://example.com"
        //     prefKey = "pref_my_mirror_url"               // optional
        //     mirror("https://mirror1.example.com")
        // }

        // 3. Custom URL (adds an EditTextPreference with URL validation)
        // baseUrl {
        //     value = "https://example.com"
        //     withCustomUrl = true
        // }


        // ── Constructor overrides ───────────────────────────────────
        // Generated as constructor arguments on the leaf class.
        overrides {
            "siteLang" to "english"
            "maxRetries" to 3
            "timeoutMs" to 5000L
            "featureEnabled" to true
        }


        // ── Deeplinks ───────────────────────────────────────────────
        // Auto-infers host from baseUrl if host is omitted.
        // Can be called multiple times.

        deeplink {
            // host = "example.com"       // optional, auto-inferred
            path("/manga/.*")
            path("/title/.*")
        }

        deeplink {
            host = "custom.example.com"
            path("/chapter/.*")
        }
    }


    // ── Additional sources (SourceFactory mode) ─────────────────────
    source {
        name = "Second Source"
[...]

PR #16992, opened 2026-06-25
Add a new build logic DSL

AwkwardPeak7:
split off from https://github.com/keiyoushi/extensions-source/pull/15946

- Contains only kotlin dsl based configuration for extensions and multisrc
  - codegen/other things in a follow up pr
- also includes new metadata fields
  - see https://github.com/mihonapp/mihon/commit/41265cce325e649d7f59c2fceabb954a174247ca

Review the first and last commit individually, second commit is only bulk migration. you can verify it only modified/replaces build.gradle(.kts) files


.github/scripts/create-repo.py
```diff
@@ -79,5 +90,5 @@
 
     index_data.append(extension_data)
 
-with REPO_DIR.joinpath("index.min.json").open("w", encoding="utf-8") as index_file:
+with REPO_DIR.joinpath("index.json").open("w", encoding="utf-8") as index_file:
```
FourTOne5, line 93: so why are we keeping a 3rd format around?

gradle/build-logic/src/main/kotlin/PluginExtension.kt
```diff
@@ -120,6 +95,52 @@ class PluginExtensionLegacy : Plugin<Project> {
...
+        val versionNameProvider = keiyoushi.libVersion.flatMap { libVersion ->
+            assertWithoutFlag(libVersion in VALID_LIB_VERSIONS) {
+                "libVersion $libVersion is not supported. Supported versions: $VALID_LIB_VERSIONS"
+            }
+            versionCodeProvider.map { "$libVersion.$it" }
+        }
+
+        val appNameProvider = keiyoushi.name.map { name ->
+            assertWithoutFlag(name.all { it.code < 0x180 }) { "Extension name should be romanized" }
+            "Tachiyomi: $name"
```
FourTOne5, line 121: after the migration we should switch the prefix to to Keiyoushi

gradle/build-logic/src/main/kotlin/PluginExtension.kt
```diff
@@ -120,6 +95,52 @@ class PluginExtensionLegacy : Plugin<Project> {
...
+            }
+            themeKeiyoushi.baseVersionCode.get() + keiyoushi.versionCode.get()
+        }.orElse(keiyoushi.versionCode)
+
+        val versionNameProvider = keiyoushi.libVersion.flatMap { libVersion ->
+            assertWithoutFlag(libVersion in VALID_LIB_VERSIONS) {
+                "libVersion $libVersion is not supported. Supported versions: $VALID_LIB_VERSIONS"
+            }
+            versionCodeProvider.map { "$libVersion.$it" }
+        }
```
FourTOne5, line 117: Probably something for the future but we can tweak the `versionCode` to be 
```kt
(libVersionMajor.padStart(3, "0") + libVersionMinor.padStart(3, "0") + legacyVersionCode.padStart(3, "0")).toInt()
```

`versionName` can be kept same

.github/scripts/create-repo.py
AwkwardPeak7: Not something that ends up in the final repo, only needed to make sure current ci pipeline works. 
This pipeline will be changed after the `source` block is added to the dsl

gradle/build-logic/src/main/kotlin/PluginExtension.kt
AwkwardPeak7: After all/most extensions are 1.6 yes

gradle/build-logic/src/main/kotlin/PluginExtension.kt
AwkwardPeak7: We can discuss this then, but again probably when all/most extensions are migrated to 1.6

.github/scripts/create-repo.py
FourTOne5: it's not being removed tho? can probably keep it under LOCAL_DIR since we don't push it

AwkwardPeak7: It is overwritten by our index.json generation; but I will move it to  LOCAL_DIR

AwkwardPeak7: ok, looking at it again, it _is_ in the local dir already

FourTOne5: the names are rather confusing across files but ok then

FourTOne5 (approved):

Issue #17025, opened 2026-06-26
Manga  livre Error 403


FinnDd064N:
### Source information

Manga livre 1.4.65

### Source language

Português Brasil

### Steps to reproduce

opening or downloading a chapter present in the extension 

### Expected behavior

open or download the app 

### Actual behavior

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/f0db60eb-1ad0-4d8f-8448-6f407b55ee48" />

### Mihon/Mihon fork version

Mihon 0.19.9

### Android version

Android 16

### Other details

The only problem that persisted after the update. 

New chapters and the extension search function are working normally.

### Acknowledgements

- [x] I have searched [existing issues](https://github.com/keiyoushi/extensions-source/issues) both open & closed, and confirm that this is a new unreported issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to version **[0.19.3](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I have updated the extension to the latest version (as listed on the [extensions listing](https://keiyoushi.github.io/extensions/) page).
- [x] I have tried the [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [x] If this is an issue with the app itself, I should be opening an issue in the app repository.
- [x] I will correctly fill out all of the requested information in this form.

### 

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/keiyoushi/extensions-source/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

mateuss8402:
### **I'm having the same problem. When I open the extension through the webview, this message appears.**
<img width="1014" height="623" alt="Image" src="https://github.com/user-attachments/assets/fa87437d-19fb-49bf-a711-fcc4a319ee75" />

### **And when I open the website directly in the browser, this other message appears; honestly, I don't understand anything anymore.**
<img width="958" height="1508" alt="Image" src="https://github.com/user-attachments/assets/b5432ff8-e1f9-4a42-ab6c-fc0b9006b734" />

The website is completely crazy.

zLuckKing:
A fix has been submitted in PR #17033. It updates the security header used by the extension and restores normal access to the site.

AshbornXS:
[...]

PR #17033, opened 2026-06-27
fix(pt/mangalivre): update security header name and value

zLuckKing:
## What changed

The site `toonlivre.net` changed its internal security header:

- Header name changed from `x-toonlivre-client` to `X-Tly-Sec`
- Default value changed from `web-x` to `web-z99`
- Added `CloudflareResolver` to handle Cloudflare JS challenges automatically via WebView.

## Why

The extension was returning HTTP 403 on all requests because it was sending the old header name `x-toonlivre-client` with the outdated fallback value `web-x`, both of which are no longer recognized by the server.

The new header `X-Tly-Sec` with value `web-z99` was confirmed via browser DevTools by inspecting the actual requests made by the site.

The regex patterns (`ANCHORED_REGEX` and `SHAPE_REGEX`) were also updated to match the new header name and support a wider range of value formats.

## Testing

Tested on Mihon and TachiyomiSY. Popular, latest, search, manga details and chapter pages all loading correctly after the fix.

## Related request

Related to #17025

## Checklist

- [x] Updated `versionCode`
- [x] Tested the changes
- [x] This fixes an existing extension
- [x] Follows the project guidelines


Smol-Ame:
1. Where's the PR checklist?
2. You didn't bump versionCode
3. You didn't reference relevant request

zLuckKing:
Updated the PR description to reference #17025. Thanks!

AwkwardPeak7:
fixed in another pr

PR #17073, opened 2026-06-28
Add source codegen to build logic DSL 

AwkwardPeak7:
```kts
keiyoushi {
    name = "My Extension"
    versionCode = 1
    libVersion = "1.4"

    source {
        lang = "en"
        baseUrl = "https://example.com"
    }
}
```

With mirrors:
```kts
source {
    lang = "en"
    baseUrl("https://example.com") {
        mirror("https://mirror1.com")
        mirror("https://mirror2.com")
    }
}
```

With user-configurable URL:
```kts
source {
    lang = "en"
    baseUrl("https://example.com") {
        withCustomUrl = true
    }
}
```

Multiple sources (replaces SourceFactory):
```kts
source {
    lang = "en"
    baseUrl = "https://en.example.com"
}
source {
    lang = "ja"
    baseUrl = "https://ja.example.com"
}
```

Opting out of codegen (e.g for Komga where name comes from preference):
```kts
source {
    lang = "all"
    baseUrl = "https://example.com"
    skipCodeGen = true
}
```


gradle/build-logic/build.gradle.kts
```diff
@@ -27,6 +28,9 @@ dependencies {
     compileOnly(gradleKotlinDsl())
     compileOnly(libs.android.gradle)
     compileOnly(libs.kotlin.gradle)
+    compileOnly(libs.ksp.gradle)
+    runtimeOnly(libs.ksp.gradle)
```
FourTOne5:
```suggestion
    implementation(libs.ksp.gradle)
```

codegen/src/main/kotlin/keiyoushi/processor/SourceProcessorProvider.kt
FourTOne5, line 1: Can just be class Provider inside `SourceProcessor`

codegen/src/main/resources/META-INF/services/com.google.devtools.ksp.processing.SymbolProcessorProvider
```diff
@@ -0,0 +1 @@
+keiyoushi.processor.SourceProcessorProvider
```
FourTOne5:
```suggestion
keiyoushi.processor.SourceProcessor$Provider
```

FourTOne5 (changes requested):

FourTOne5 (dismissed):

FourTOne5 (approved):

PR #17169, opened 2026-06-30
Nudemoon: move to new dsl

Exinaus:
Checklist:

Changed cookies manipulation to lib:cookieinterceptor, as instructed in contributing guide.
Accessing `baseUrl` (to create an interceptor and creating `domain` var) caused an error, until switched to a lazy method.


src/ru/nudemoon/build.gradle.kts
```diff
@@ -2,10 +2,20 @@ plugins {
     alias(kei.plugins.extension)
 }
 
+dependencies {
```
AwkwardPeak7: place it below the keiyoushi block

src/ru/nudemoon/src/eu/kanade/tachiyomi/extension/ru/nudemoon/Nudemoon.kt
```diff
@@ -26,40 +23,22 @@ import java.net.URLEncoder
...
-    override val name = "Nude-Moon"
-    override val lang = "ru"
+@Source
+abstract class Nudemoon : HttpSource() {
     override val supportsLatest = true
-    private val preferences: SharedPreferences = getPreferences()
-    override val baseUrl by lazy { getPrefBaseUrl() }
-
     private val dateParseRu = SimpleDateFormat("d MMMM yyyy", Locale("ru"))
+    private val domain by lazy { baseUrl.toHttpUrl().host }
```
AwkwardPeak7:
```suggestion
    private val domain get() = baseUrl.toHttpUrl().host
```

AwkwardPeak7 (approved):

src/ru/nudemoon/build.gradle.kts
Exinaus: That helped with the `domain` var. Tried to take `client` out of lazy initiation just out of curiosity, still got:

```
Caused by: java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.Object kotlin.Lazy.getValue()' on a null object reference
	at eu.kanade.tachiyomi.extension.ru.nudemoon.ExtensionGenerated.getCustomUrlPrefs(ExtensionGenerated.kt:21)
```

But it's already merged, before i even wrote about it, so ... works for me.

AwkwardPeak7: yeah this is issue of jvm class initialization order, parent class gets initialized before child class. so accessing baseUrl (which has a backing field of custom url pref helper) during init will throw. note that this will not happen for static baseUrl version

AwkwardPeak7: you can kind of hack together a fix:

```kt
private var clientCache: Pair<String, OkHttpClient>? = null

override val client: OkHttpClient
    get() = if (clientCache != null && clientCache.first == domain) {
            clientCache.second
        } else {
            // build client and save it to clientCache with domain pair
        }
```

PR #17179, opened 2026-06-30
Migrate Senkuro to new dsl

Exinaus:
Checklist:


lib-multisrc/senkuro/src/eu/kanade/tachiyomi/multisrc/senkuro/Senkuro.kt
```diff
@@ -44,12 +41,15 @@ abstract class Senkuro(
 
     private val preferences: SharedPreferences by getPreferencesLazy()
 
-    override val baseUrl: String
-        get() = if (apiUrl.contains("api.senkuro.me")) "https://senkuro.me" else _baseUrl
-
     private val apiUrl: String
         get() = preferences.getString(API_DOMAIN_PREF, API_DOMAIN_DEFAULT)!! + "/graphql"
```
AwkwardPeak7: is the api and site url different? can we not have a mapping of site url to api url? or maybe don't have config for these at all?

Exinaus: Yes, different. API url's are: `https://api.senkuro.me/grpahql` (available only from russia) and `https://api.senkuro.com/graphql` (globally available, maybe banned in russia, because i saw few messages that after changing default api to global in one of the updates ppl couldn't access it, they didn't know that there was a setting to change it). Extension have a preference to choose between one of them, not the `baseUrl`. 
What differentiate betwen those sites are this header (as far as i could see in the code), not the site/api url:

`.add("App-Id", if (name == "Senkuro") "4026531840100" else "5033164800100")`

Same availability rules goes for site url: `senkuro.me` available only from russia, `senkuro.com`/`senkognito.com` - globally.

I could do a reverse, i think. Make a setting to chose from russian/global `baseUrl` and assign API url accordingly.

Exinaus: Double-checked: `senkognito.com` also has api. Basically you can make `apiUrl` from `baseUrl` by replacing `https://` with `https://api.` and adding `/graphql` at the end.

AwkwardPeak7 (approved):

PR #17225, opened 2026-07-01
MangaLivre: reading gate uses x-tly-token (via Headers.append)

JPPNovais:
Follow-up to #17030 / #17082 / #17104. Previous fixes chased the wrong header. Testing the live site revealed the real gate.

**Root cause.** The reading endpoints (`/api/mangas/{id}/chapters/{id}` and `/chapters-paginated`) reject requests with `403 {"error":"Acesso negado. Por favor, use o aplicativo oficial."}` unless a specific client header is present. The open endpoints (search / releases / manga-by-slug) do **not** gate, so browsing worked while reading failed.

In the bundle the request-header builder does:

```js
const S = new Headers(opts.headers || {})
S.append("x-tly-token", "v99-web-z")   // ← the real gate header (.append!)
S.set("app-token", "tok-z99")          // ← decoy the open endpoints ignore
if (method !== "GET"/"HEAD"/"OPTIONS") S.set("x-csrf-token", B)
```

The discovery added in #17104 only scanned `.set(...)`, so it picked the `app-token` **decoy** and the reading gate kept returning 403. The real header is `x-tly-token`, injected via `Headers.append(...)`.

**Fix.** Match both `.set(...)` and `.append(...)` literal pairs when scanning the bundle, and default to `x-tly-token` / `v99-web-z`. The existing multi-candidate `403` retry (which already keys on the "aplicativo oficial" body) then discovers/validates the right header and self-heals on future name/value rotations.

**Verified against the live API** (real browser past Cloudflare, replaying the exact interceptor logic):
- Cold start with the default `x-tly-token: v99-web-z` → `200` (pages load).
- Simulating a stale token (`app-token`) → `403` → the retry scans the bundle, extracts `x-tly-token/v99-web-z`, retries → `200`, and caches it.

### Changes

- Scan `.set(...)` **and** `.append(...)` literal header pairs in the bundle.
- Default client header → `x-tly-token` / `v99-web-z`.
- Bump `versionCode` to 69.

No endpoint URLs or DTOs changed.

### Checklist

- [x] Updated `versionCode` value in `build.gradle.kts`
- [x] Set the `contentWarning` configuration in `build.gradle.kts` appropriately (unchanged: SAFE)
- [x] Have not changed source names
- [ ] Have tested the modifications by compiling and running the extension through Android Studio — **iOS only, no Android.** I verified the header/gate against the live site and replayed the interceptor logic end-to-end (200 on the reading endpoint); checked formatting with ktlint locally; relying on CI to build.
[...]

PR #17272, opened 2026-07-02
MangaLivre: decode base64/atob-obfuscated reading-gate header

JPPNovais:
Follow-up to #17225. The reading gate escalated its obfuscation again, which is why the header discovery kept missing it.

**What changed.** The client header the reading endpoints require is no longer a string literal in the bundle — it is now built with `atob()` (base64):

```js
S.append(atob("YXBwLXNlYy10b2tlbg=="), atob("ejExLXdlYi15"))
//        └ "app-sec-token"             └ "z11-web-y"
```

So `#17225`'s literal `.set/.append("name","value")` scan found only decoys, and reading kept returning `403 {"error":"Acesso negado. Por favor, use o aplicativo oficial."}`.

**Fix.** Also match `.set/.append(atob("…"), atob("…"))` and base64-decode both parts (via okio `decodeBase64().utf8()`), alongside the plain-literal form. Decoded names are sanity-checked against a header-name shape and the standard-header denylist; the existing multi-candidate `403` retry then validates and caches the right one, self-healing across name/value rotations. Default → `app-sec-token` / `z11-web-y`.

**Verified end-to-end against the live API** (real browser, replaying the exact interceptor logic on a reading endpoint):
- Candidates extracted with the new regexes → `[app-sec-token/z11-web-y, lite/1]` (the atob pair is correctly decoded).
- Cold start with the default → `200` (pages load).
- Simulating a stale token → `403` → scans the bundle, decodes `app-sec-token`, retries → `200`, caches it.

### Changes

- Decode base64/`atob(...)` header pairs from the bundle (not just string literals).
- Default client header → `app-sec-token` / `z11-web-y`.
- Bump `versionCode` to 70.

No endpoint URLs or DTOs changed.

### Checklist

- [x] Set the `contentWarning` configuration in `build.gradle.kts` appropriately (unchanged: SAFE)
- [ ] Have tested the modifications by compiling and running the extension through Android Studio — **iOS only, no Android.** Verified the header/gate and replayed the interceptor logic end-to-end against the live reading endpoint (200); checked formatting with ktlint locally; relying on CI to build.

### DIFF
--- a/src/pt/mangalivre/build.gradle.kts
+++ b/src/pt/mangalivre/build.gradle.kts
@@ -4,7 +4,7 @@ plugins {
 
 keiyoushi {
     name = "Manga Livre"
-    versionCode = 69
+    versionCode = 70
     contentWarning = ContentWarning.SAFE
     libVersion = "1.4"
 
--- a/src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
+++ b/src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
@@ -21,6 +21,7 @@ import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import okhttp3.Response
+import okio.ByteString.Companion.decodeBase64
 import java.io.IOException
 import kotlin.time.Duration.Companion.seconds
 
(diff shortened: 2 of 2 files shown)


FelipeGFA:
i think u can avoid all of those chages by just proxying all requests through webview, At least these small header changes won't break the extension until a major change is made

AshbornXS:
> until a major change is made

don't give them ideas 😭

AwkwardPeak7 (approved):

PR #17274, opened 2026-07-02
Move rest of zh Extensions to new dsl

Secozzi:
`skipCodeGen` is used for extensions with dynamic mirror list


src/zh/baozimanhua/src/eu/kanade/tachiyomi/extension/zh/baozimanhua/Baozi.kt
```diff
@@ -56,12 +42,12 @@ class Baozi :
     override val client = network.client.newBuilder()
```
AwkwardPeak7: client by lazy

AwkwardPeak7 (approved):

PR #17299, opened 2026-07-03
Komiic: Revert some changes to fix thumbnail loading error

hualiong:
Checklist:

### DIFF
--- a/core/translations/strings.json
+++ b/core/translations/strings.json
@@ -25,8 +25,18 @@
     "customUrlDialogMessage": "Bỏ trống để sử dụng giá trị mặc định"
   },
   "zh": {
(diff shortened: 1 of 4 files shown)


Rafou376:
In which cases does the error occur?
I can't reproduce it

hualiong:
> In which cases does the error occur? I can't reproduce it

This mirror site (komiic.cc) is just a reverse proxy — it still uses links to public.komiic.com for the thumbnails it returns. For users within China, all domains ending in *.komiic.com have been blocked by the GFW. So, even if you switch to komiic.cc in China, the thumbnails will still fail to load.

That's why I used a interceptor to replace the host in the request if using a mirror site.

AwkwardPeak7 (approved):

PR #17328, opened 2026-07-04
Add custom URL support with mirror selection

dejavui:
Introduce `CustomAndMirrorPreferences`, allowing extensions to combine predefined mirrors with a custom base URL option.

This generalizes the existing MoeTruyen implementation into a reusable framework component, so extensions no longer need to implement their own mirror/custom URL preference logic.

### Changes

* Added `CustomAndMirrorPreferences` to handle mirror selection and custom base URLs in a single preference component.
* Extended `BaseUrlSpec.Mirrors` with optional `entries`, `values`, and `withCustom` support for declarative mirror configuration.
* Updated `SourceProcessor` to generate `CustomAndMirrorPreferences` when custom URLs are enabled, while preserving the existing `MirrorPreferences` path for regular mirror-only sources.
* Migrated the MoeTruyen extension to the new declarative `@Source` configuration.
* Removed the manual mirror/custom URL preference implementation from MoeTruyen, as it is now generated by the framework.

Review needed 👉ﾟヮﾟ👉 from @AwkwardPeak7, @vetleledaal, and @FourTOne5. Thanks!

Checklist:


AwkwardPeak7:
I feel like mirrors and custom urls should remain mutually exclusive.

dejavui:
> I feel like mirrors and custom urls should remain mutually exclusive.

I understand the concern. My intention isn't to merge two unrelated features, but to model different ways of resolving the effective base URL ＼（〇_ｏ）／ 

MoeTruyen has been using this workflow for a time: users choose either the default domain, one of the predefined mirrors, or a custom URL. Only one option is active at a time, so they're still mutually exclusive from the user's perspective.

FourTOne5:
if this were to be considered i would expect the custom and mirrors option to be merged

PR #17418, opened 2026-07-07
MyHentaiGallery: route artist/parody genre clicks to tag listings

patch-and-paste:
Adds support for correctly routing artist/parody genre clicks to tag listings rather than populating the search.

Checklist:


src/en/myhentaigallery/src/eu/kanade/tachiyomi/extension/en/myhentaigallery/MyHentaiGallery.kt
```diff
@@ -224,6 +242,8 @@ abstract class MyHentaiGallery : HttpSource() {
 
     companion object {
         const val PREFIX_ID_SEARCH = "id:"
+        private const val ARTIST_GENRE_PREFIX = "Artist: "
+        private const val PARODY_GENRE_PREFIX = "Parody: "
```
AwkwardPeak7, line 246: perhaps prefix with zero width characters? e.g one zero width character for artist and two zero width characters for genre.

patch-and-paste: > perhaps prefix with zero width characters? e.g one zero width character for artist and two zero width characters for genre.

Visible makes the most sense to me. The main issue with zero-width markers is that they're much harder to debug if something goes wrong. And assuming a client doesn't strip the prefix, it helps tell the user what kind of tag it is, since categories, artists, and parodies all get kinda merged together

AwkwardPeak7 (approved):

PR #17445, opened 2026-07-08
MangaDistrict: Option to not clean title during browsing/search

cuong-tran:
**Reason:** This source often has 2 versions of a same manhwa, the normal one and the "(Uncensored)" one, showing together. For the purpose of normalizing title and duplicate matching, the pref to `Clean title` need to be turned on. But when it's turned on, both versions appear to have the same name while browsing and are indistinguishable.

This PR adds a pref that won't apply cleaning title while browsing. So that:
- When browsing/searching, the 2 versions are distinguishable.
- When opening the detail screen, the title is still cleaned so I can add it to library normalized.

Checklist:

### DIFF
--- a/src/en/mangadistrict/build.gradle.kts
+++ b/src/en/mangadistrict/build.gradle.kts
@@ -4,7 +4,7 @@ plugins {
 
 keiyoushi {
     name = "Manga District"
-    versionCode = 16
+    versionCode = 17
     contentWarning = ContentWarning.NSFW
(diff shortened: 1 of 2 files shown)


AwkwardPeak7 (commented):
do we really need this preference?

cuong-tran:
> do we really need this preference?

I added its use case.

AwkwardPeak7 (approved):

PR #17472, opened 2026-07-09
AllManga: restore page loading via signed API request 

### LINKED ISSUE #17318: AllManga: `Time out waiting for page list` error
### Source information

1.4.19

### Source language

English

### Steps to reproduce

) Access the extension
) Select any comic
) Click on any chapter 

### Expected behavior

Load the chapter

### Actual behavior

Time out waiting for page list

### Mihon/Mihon fork version

Mihon 0.20.0

### Android version

Android 12

### Other details

_No response_

### Acknowledgements

### 

---

PoppingMangoSources:
Closes [#17318](https://github.com/keiyoushi/extensions-source/issues/17318)

Adds support for AllManga's current **mkissa.to** build, which now serves chapter
pages through a signed API. mkissa gates `chapterPages` behind a rotating `aaReq`
signature and returns the page list AES-GCM encrypted in a `tobeparsed` field.
`PageSigner` pulls mkissa's signing key, reproduces the signature and decryption
with built-in `javax.crypto`, and queries the shared `api.allanime.day` so pages
load again. The WebView path is kept as a fallback.

Checklist:

### DIFF
--- a/src/en/allanime/build.gradle.kts
+++ b/src/en/allanime/build.gradle.kts
@@ -4,7 +4,7 @@ plugins {
 
 keiyoushi {
     name = "AllManga"
-    versionCode = 19
+    versionCode = 20
     contentWarning = ContentWarning.NSFW
     libVersion = "1.4"
 
--- a/src/en/allanime/src/eu/kanade/tachiyomi/extension/en/allanime/AllManga.kt
+++ b/src/en/allanime/src/eu/kanade/tachiyomi/extension/en/allanime/AllManga.kt
@@ -54,6 +54,8 @@ abstract class AllManga :
         .rateLimit(1) { it.host == apiUrlHost }
         .build()
 
+    private val pageSigner by lazy { PageSigner(client, headers) }
+
     override fun headersBuilder() = super.headersBuilder()
         .add("Referer", "$baseUrl/")
 
@@ -215,12 +217,24 @@ abstract class AllManga :
 
     /* Pages */
     override fun fetchPageList(chapter: SChapter): Observable<List<Page>> = Observable.fromCallable {
-        pageListFromWebView(chapter)
+        pageListFromApi(chapter).ifEmpty { pageListFromWebView(chapter) }
     }
 
     override fun pageListRequest(chapter: SChapter): Request = throw UnsupportedOperationException()
 
     override fun pageListParse(response: Response): List<Page> = throw UnsupportedOperationException()
+
+    private fun pageListFromApi(chapter: SChapter): List<Page> {
+        val chapterUrlParts = chapter.url.split("/")
+        val mangaId = chapterUrlParts[2]
+        val chapterString = chapterUrlParts[4].removePrefix("chapter-").removeSuffix("-sub")
+
+        val pageList = runCatching {
+            pageSigner.getPageList(mangaId, chapterString, "sub")?.pageList
+        }.getOrNull() ?: return emptyList()
+
+        return pageList.toPages()
+    }
 
     @SuppressLint("SetJavaScriptEnabled")
     private fun pageListFromWebView(chapter: SChapter): List<Page> {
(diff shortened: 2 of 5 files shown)


AwkwardPeak7:
doesn't a simple domain change with the old webview path not work?

PoppingMangoSources:
> doesn't a simple domain change with the old webview path not work?

Yeah, that was my first instinct too, but a domain swap on its own didn’t get me pages. From what I could tell the list is signed now, mkissa signs the chapterPages request with a rotating key it inlines in the reader shell and returns the list AES-GCM encrypted, which is why the old direct query stopped working. Pointing the webview at mkissa just landed on the Cloudflare challenge (the reader route is gated), and allmanga’s old reader wasn’t surfacing chapterPages anymore either, so the capture had nothing to grab. Signing the API call directly was the only thing that loaded reliably for me since that endpoint isn’t behind CF, kept the webview as a fallback for Allmanga because it won't work on Mkissa.

If there’s a lighter way to do it i'm happy to change it, this was just what worked in testing in my app after a lot of possible changes on my end. I tested thoroughly.

Alpha-782:
https://allmanga.to/ is not accessible anymore, just for your info.

I looked a bit more into it before, and chapters on https://mkissa.to/ were using `/manga/` path instead of `/read/`.

PoppingMangoSources:
> https://allmanga.to/ is not accessible anymore, just for your info.
> 
> 
> 
> I looked a bit more into it before, and chapters on https://mkissa.to/ were using `/manga/` path instead of `/read/`. 


Good to know, thanks. The signed API path doesn’t actually depend on allmanga.to being up, it pulls the key from whichever mirror still serves the bootstrap (mkissa right now) and then hits api.allanime.day directly, so pages still load with allmanga down. The only part tied to allmanga is the webview fallback.

On the AllManga Discord they’re saying the site devs are working on fixing allmanga.to, so it should be a temporary outage, I’d lean toward keeping it in the mirror list rather than dropping it, and just making sure mkissa is tried first so it doesn’t hang on the dead host in the meantime.
And yeah, chapter URLs are already built with the /manga/ path (getChapterUrl → $baseUrl/manga/$id/$chapter) in the code, so that lines up with what you saw on mkissa. The reader currently works perfectly on my app with the code I just pushed.

ImonChakraborty:
> Good to know, thanks. The signed API path doesn’t actually depend on allmanga.to being up, it pulls the key from whichever mirror still serves the bootstrap (mkissa right now) and then hits api.allanime.day directly, so pages still load with allmanga down. The only part tied to allmanga is the webview fallback.

How are u getting the part A hex?

simple GET request gives the part B hex via `window.__aaCrypto` along with `epoch` `epochMs` `graceMs` `switchAt` (I suppose it will change over time, fetching dynamically over runtime is better) 

and as much as I understand, `AES key = key_a XOR base64_decode(key_b)`  ?

so what about the part A hex?

PoppingMangoSources:
> How are u getting the part A hex?
> 
> 
> 
> simple GET request gives the part B hex via `window.__aaCrypto` along with `epoch` `epochMs` `graceMs` `switchAt` (I suppose it will change over time, fetching dynamically over runtime is better) 
> 
> 
> 
> and as much as I understand, `AES key = key_a XOR base64_decode(key_b)`  ?
> 
> 
> 
> so what about the part A hex?

Yeah, your understanding is spot on. AES key = key_a XOR base64_decode(key_b).
key_b (part B) is the rotating half. It comes from window.__aaCrypto (that simple GET) along with epoch/epochMs/graceMs/switchAt, and you’re right that it changes over time, so we fetch it live on every request rather than storing it. That part is fully dynamic.

key_a (part A) isn’t served anywhere. It’s a static constant compiled into the site’s own bundle. If you deobfuscate the reader JS you can see it declared right alongside the buildId: 

…231eee4e73":"", cr = mt(362)!=="string" ? "12" : "";

where the string ending in 231eee4e73 is part A (78ebe405…4e73) and cr is the buildId, hardcoded to "12". Neither is in the bootstrap payload, the client sends buildId to /client-crypto/v1/bootstrap?buildId=12, the server never hands it back. So the only way to get part A (and the buildId) is to pull them out of the bundle once, which is what PART_A_HEX / BUILD_ID are. 

On upkeep: since part B rotates dynamically, normal key rotation needs nothing from us. Part A and the buildId are build-pinned literals, they’d only change if the site ships a new front-end build, which is rare and a one-line bump each. And it fails safe in the meantime: if signing ever goes stale it returns empty and falls back to the webview, so it degrades instead of breaking. Stable on buildId 12 so far.

ImonChakraborty:
> Yeah, your understanding is spot on. AES key = key_a XOR base64_decode(key_b). key_b (part B) is the rotating half. It comes from window.__aaCrypto (that simple GET) along with epoch/epochMs/graceMs/switchAt, and you’re right that it changes over time, so we fetch it live on every request rather than storing it. That part is fully dynamic.
> 
> key_a (part A) isn’t served anywhere. It’s a static constant compiled into the site’s own bundle. If you deobfuscate the reader JS you can see it declared right alongside the buildId:
> 
> …231eee4e73":"", cr = mt(362)!=="string" ? "12" : "";
> 
> where the string ending in 231eee4e73 is part A (78ebe405…4e73) and cr is the buildId, hardcoded to "12". Neither is in the bootstrap payload, the client sends buildId to /client-crypto/v1/bootstrap?buildId=12, the server never hands it back. So the only way to get part A (and the buildId) is to pull them out of the bundle once, which is what PART_A_HEX / BUILD_ID are.
> 
> On upkeep: since part B rotates dynamically, normal key rotation needs nothing from us. Part A and the buildId are build-pinned literals, they’d only change if the site ships a new front-end build, which is rare and a one-line bump each. And it fails safe in the meantime: if signing ever goes stale it returns empty and falls back to the webview, so it degrades instead of breaking. Stable on buildId 12 so far.

Thanks for the explanation, appreciate it 👍️

tho,

image

I believe it has changed already, this is for buildId: 13 ig

`const $n = _t(460) !== 'string' ? 'f5dc46e6f42968c5ed0eab602d6ae8f2107991006f02876947e64fcb75d53da6' : '',
zr = '13';`

chunk is: https://cdn.allanime.day/all/mk/_app/immutable/chunks/CY39o1wT.js

this is changing very frequently, doesnt seem like we can hardcode it

PR #6694, opened 2024-12-19
Add DeviantArt

DokterKaj:
Closes #4529 

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [ ] ~Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions~
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [ ] ~Have explicitly kept the `id` if a source's name or language were changed~
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [x] Have removed `web_hi_res_512.png` when adding a new extension

PR #17490, opened 2026-07-10
DeviantArt: cookie login, multi-image fix, username/URL search, Sub-Gallery support

tapdefenser:
The code has already been human checked and tested.

### New Features
1. **Cookie login** — Paste browser-exported cookies.json in Filter to access restricted galleries. Use filter to login, the cookie should use one time only when login, after that no need to use cookie again. Otherwise it will cause sudden logout。
2. **Multi-image deviation support** — Properly extracts and displays all images from multi-image posts.
3. **Username search & URL** — Search by plain username to browse gallery folders, or paste a full DeviantArt gallery URL.
4. **Sub-Gallery support** — The tool now correctly parses sub-folder structures

### Credits
Original DeviantArt extension by @DokterKaj (#6694).

Checklist:


FiorenMas:
Why not login by webview? And then cookie will sync?

tapdefenser:
@FiorenMas Since DeviantArt has a verification step during WebView login that detects and blocks automated software, there's no way to bypass it.

FiorenMas:
I tested it, and the problem is caused by a hardcoded User-Agent. Please remove it (maybe it will lead something broke, because it have a reason for this hardcode UA) and disable your DNS adblocker, if you have one. Force user to input cookie very bad idea.
image

tapdefenser:
@FiorenMas If i use google E-Mail to login it says "This account was originally created with an email address. Try logging in with your username and password instead."
If i turn to use my username and password then it have a forced verification

Then if i try my new google account then it says my Browser is unsafe

FiorenMas:
Case 1: I said you above you don't see it?
Case 2: Google login need some trick remove `wv` in their UA. You can follow code from here
https://github.com/keiyoushi/extensions-source/blob/d41cb5e9ff588e0b2169f8b709a1ad7ae187c590/src/vi/cmanga/src/eu/kanade/tachiyomi/extension/vi/cmanga/CManga.kt#L38-L48

tapdefenser:
@FiorenMas In Case 1, I don't have any DNS adblocker, and I also tried to use mobile data, but it still not Working.
Later I have tested and yes the problem is as you said is the hardcoded User-Agent in the code, which enables the scraper to Windows websites (not on an android site) only.
So i managed to add a switch which can change the UA used by the extension.
https://github.com/keiyoushi/extensions-source/pull/17527

PR #9815, opened 2025-07-27
Jiangzaitoon: update domain, fix blurry thumbnails

vetleledaal:
Madara was also updated to allow overriding `String.getSrcSetImage()`, not bumping the multisrc.

Closes #8771

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [x] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [x] Have removed `web_hi_res_512.png` when adding a new extension

PR #17518, opened 2026-07-10
fix(madara): properly parse srcset with image descriptors (w/x suffixes)

IMXEren:
Refactored getSrcSetImage() to correctly handle comma-separated srcset entries with width (w) and pixel-density (x) descriptors. Previously the method used to compare URLs lexicographically, which breaks on srcset="link-34x68.jpg 34w, link-330x700.jpg 330w" and chooses link-34x68.jpg.

Adds IMAGE_DESCRIPTOR_REGEX companion constant and prefers highest-resolution images by descriptor value, falling back to lexicographical URL comparison when no descriptors are present.

Checklist:

### DIFF
--- a/lib-multisrc/madara/src/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
+++ b/lib-multisrc/madara/src/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
@@ -824,9 +824,27 @@ abstract class Madara : HttpSource() {
     /**
      *  Get the best image quality available from srcset
      */
-    protected open fun String.getSrcSetImage(): String? = this.split(" ")
-        .filter(URL_REGEX::matches)
-        .maxOfOrNull(String::toString)
+    protected open fun String.getSrcSetImage(): String? {
+        val images = this.split(",")
+            .map { it.trim().split("\\s+".toRegex()) }
+            .filter { it.isNotEmpty() && URL_REGEX.matches(it[0]) }
+
+        val imagesWithDescriptor = images
+            .filter { it.size == 2 }
+            .mapNotNull { candidate ->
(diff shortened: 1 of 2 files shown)


IMXEren:
Hey @vetleledaal, just checking in to see if everything looks good here or if anything is missing. 

Thanks!

lib-multisrc/madara/src/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
```diff
@@ -824,9 +824,27 @@ abstract class Madara : HttpSource() {
     /**
      *  Get the best image quality available from srcset
      */
-    protected open fun String.getSrcSetImage(): String? = this.split(" ")
-        .filter(URL_REGEX::matches)
-        .maxOfOrNull(String::toString)
+    protected open fun String.getSrcSetImage(): String? {
+        val images = this.split(",")
+            .map { it.trim().split("\\s+".toRegex()) }
```
vetleledaal: Move the class to avoid recompilation of invariant regex.

vetleledaal (commented):
Need additional consideration when making changes to Madara as it will update a lot of extensions (approx. 300).

I made something similar in the past, but it was specific to that one source (via #9815):

https://github.com/keiyoushi/extensions-source/blob/a4b215a52c9caac41a9825300389711cc1344ddb/src/tr/jiangzaitoon/src/eu/kanade/tachiyomi/extension/tr/jiangzaitoon/Jiangzaitoon.kt#L26-L47

IMXEren:
> Thank you for your review.

So, should I also modify Madara derived sources, if they were using this sort of method to get the srcset image.

AwkwardPeak7 (approved):

PR #17527, opened 2026-07-11
Deviantart Login Assist button

tapdefenser:
Enables Multi-way login in WebView by stripping the ”wv" token from the default User-Agent header. Switch on and login in the Webview
When disabled, a generic windows Firefox User-Agent is used for all requests.
Note: After switching, please restart the App. If this extension stop working, please disable the switch after login.

The code has been human checked and tested

Checklist:


AwkwardPeak7:
wouldn't it be better to make the source work with mobile useragent so hardcoded useragent is not needed

tapdefenser:
Yes i 100％ agree but since i want to fix the Login problem asap> wouldn't it be better to make the source work with mobile useragent so hardcoded useragent is not needed

AwkwardPeak7:
not needed after https://github.com/keiyoushi/extensions-source/pull/17861

PR #17535, opened 2026-07-11
MangaLivre: decode alphabet-index reading-gate token

JPPNovais:
Follow-up to #17521. The site swapped its char-code obfuscation for a **custom-alphabet index** scheme:

```js
$ = "abcdefghijklmnopqrstuvwxyz0123456789-",
me = [0,15,15,4,13,3].map(i => $[i]).join(""),               // "append"
ie = [23,36,19,11,24,36,18,8,6,12,0].map(i => $[i]).join("") // "x-tly-sigma"
```

So each header char is an index into an alphabet literal. The char-code decoder from #17521 no longer matches, and reading fell back to the stale default. Current header: `x-tly-sigma` / `w99-enigma-z` (with a decoy `v00-decoy-y`).

**Approach.** Since the obfuscation *class* keeps changing (literal → base64 → char-codes → char-codes+offset → alphabet-index), this decodes **all** of them into one pool and leans on two stable anchors: the header **name** always looks like `x-…`/`app-…`, and the value is always `<seg>-<word>-<letter>` (e.g. `w99-enigma-z`, `q88-matrix-o`, `c77-block-q`). Candidates are filtered by those strict shapes and validated by the `403 "aplicativo oficial"` oracle, so the decoy and CSS-class noise are skipped.

- `decodeAlphabet`: `[indices].map(i => ALPHA[i])` using alphabet literals from the bundle.
- kept `decodeCharCodes` (with offset) + `decodeAtob` + added `decodeLiterals`.
- strict `NAME_REGEX` / `VALUE_REGEX`; brute-force via the 403 oracle.
- Default → `x-tly-sigma` / `w99-enigma-z`. Bump `versionCode` to 76.

**Verified end-to-end against the live reading endpoint** (real browser, exact logic, real chapter): decodes `["x-tly-sigma", "w99-enigma-z", "v00-decoy-y"]`, and the oracle lands on `x-tly-sigma: w99-enigma-z` → `200` (decoy skipped).

### Checklist

- [x] Set the `contentWarning` configuration appropriately (unchanged: SAFE)
- [ ] Have tested by compiling/running through Android Studio — iOS only; verified decode + gate + oracle end-to-end against the live endpoint, plus ktlint locally; relying on CI to build.

### DIFF
--- a/src/pt/mangalivre/build.gradle.kts
+++ b/src/pt/mangalivre/build.gradle.kts
@@ -6,7 +6,7 @@ plugins {
 
 keiyoushi {
     name = "Manga Livre"
-    versionCode = 75
+    versionCode = 76
     contentWarning = ContentWarning.SAFE
     libVersion = "1.4"
 
--- a/src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
+++ b/src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
@@ -242,12 +242,11 @@ abstract class MangaLivre :
 
     private fun scrapeStaticCandidates(): List<ClientToken> = try {
         val js = fetchBundle()
-        val names = (decodeCharCodes(js) + decodeAtob(js))
-            .filter(::isHeaderCandidate)
-            .distinct()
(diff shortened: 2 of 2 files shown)


AwkwardPeak7 (approved):
this is getting increasingly brittle...

PR #17542, opened 2026-07-11
AllManga (EN): Fix mangaUrl reset

disk-iq8:
Checklist:

### DIFF
--- a/src/en/allanime/src/eu/kanade/tachiyomi/extension/en/allanime/AllManga.kt
+++ b/src/en/allanime/src/eu/kanade/tachiyomi/extension/en/allanime/AllManga.kt
@@ -42,9 +42,10 @@ import java.util.concurrent.TimeUnit
 abstract class AllManga :
     HttpSource(),
     ConfigurableSource {
-    private val apiUrlHost by lazy { apiUrl.toHttpUrl().host }
 
     private val apiUrl = "https://api.allanime.day/api"
+
+    private val apiUrlHost by lazy { apiUrl.toHttpUrl().host }
 
     override val supportsLatest = true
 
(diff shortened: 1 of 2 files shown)


src/en/allanime/src/eu/kanade/tachiyomi/extension/en/allanime/AllManga.kt
```diff
@@ -42,18 +42,45 @@ import java.util.concurrent.TimeUnit
...
 
     override val supportsRelatedMangas = false
 
     private val preferences by getPreferencesLazy()
 
+    private val retryRegex = Regex("""in (\d+) second""", RegexOption.IGNORE_CASE)
+
     override val client = network.client.newBuilder()
         .apply { interceptors().removeAll { it.javaClass.simpleName == "CloudflareInterceptor" } }
+        .addInterceptor { chain ->
```
AwkwardPeak7, line 60: perhaps increase the ratelimit instead?

disk-iq8: Not sure what number is good without slowing usual usage down

if 1 request a second currently still hits it
something like 2 requests in 1 second hits a limit, for example if you refresh any manga entry three times in a row you get rate limited

this code is like a final hope for library Update, it should never be needed for regular use

AwkwardPeak7: maybe we should enhance the ratelimiter to account for 429... outside of the scope for this pr ofc

disk-iq8: yeah something like retry failsafe would be nice, but this api in particular returns 200 status, so it's one more handicap

AwkwardPeak7: bruh

AwkwardPeak7 (approved):

PR #17570, opened 2026-07-12
Mangadotnet [EN]: Update Functionality

Alpha-782:
- Add 'Volumes' filter;
- Add 'Bookmarks' (Popular Mode, WebView cookies required);
- Add 'For You' suggestions (cached, only one network request after app restart). Requires WebView login (With MangaBaka sync enabled). The site also caches them for 24 hours, so not caching them is forcing a redundant network call. This is completely user personalized from MangaBaka and some tweaks from Mangadotnet, which makes this feature very useful for Komikku and other apps that support suggestions;
- Update volume URLs and language filtering, including deeplink with volumes;
- I was annoyed with inconsistent formatting, so I fixed that too;
- Bump to version 14.

Checklist:


src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Dto.kt
```diff
@@ -53,7 +53,9 @@ class ViewAllData(
 
 @Serializable
 class BrowseManga(
-    private val id: Int,
+    val id: Int,
+    @SerialName("manga_id")
+    private val mangaId: Int? = null,
```
AwkwardPeak7: use `@JsonNames` annotation on one field instead

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
```diff
@@ -171,32 +169,38 @@ abstract class Mangadotnet :
...
-            }
-            excludedDemographicsPref().forEach { demo ->
-                addQueryParameter("genre", "-$demo")
-            }
-            addQueryParameter("_routes", "pages/ViewAllPage")
-            includedTypes().forEach { origin ->
-                addQueryParameter("origin", origin)
+    // ============================== Popular ==============================
+    override fun popularMangaRequest(page: Int): Request {
+        val mode = popularModePref()
```
AwkwardPeak7: imo, make a checkbox filter for this instead of preferences

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
```diff
@@ -253,7 +257,6 @@ abstract class Mangadotnet :
 
     override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
         val url = "$baseUrl/search.data".toHttpUrl().newBuilder().apply {
-            addAdultParam()
```
AwkwardPeak7, line 256: why remove?

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
```diff
@@ -714,6 +728,104 @@ abstract class Mangadotnet :
         }
     }
 
+    // =========================== For You Cache ===========================
+    @Volatile
+    private var forYouMemoryCache: List<BrowseManga>? = null
+
+    private val forYouCacheFile: File by lazy {
+        applicationContext.cacheDir.resolve("source_$id/for_you.json")
```
AwkwardPeak7, line 739: do you need to cache this?

AwkwardPeak7 (changes requested):

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
Alpha-782: Search ignores that parameter, which is why fake images were added when an entry is NSFW.

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
Alpha-782: > Add 'For You' suggestions (cached, only one network request after app restart). Requires WebView login (With MangaBaka sync enabled). The site also caches them for 24 hours, so not caching them is forcing a redundant network call.

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
Alpha-782: Do you mean remove Popular Mode preference?

AwkwardPeak7: Yes and instead add it to filters. If this filter is enabled, redirect search to it.

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
AwkwardPeak7: Check if the site returns proper cache headers for this. If it does, okhttp will itself cache it.

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
Alpha-782: You're sure? 🤔

src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt
Alpha-782: I checked it out, and I don't think it does. I did assume it is, because of this announcement below:
image

Taken from: https://mangadot.net/community/101

Only mention of cache in the headers of the API call:
image

AwkwardPeak7 (approved):

PR #17652, opened 2026-07-16
Hiper (multi-src): Migrate to 1.6, Filters & default max_rating

cuong-tran:
* Add filters to search, with i18n
* Add maximum default rating to preferences
* Add type & rating to genre tag (so they can be clicked and search with filters)
* HiperDex: add mirrors and set default domain to `.tv` (since the error message in `.com` saying something like "use official at .tv")
* Migrate to lib 1.6
* Using `KeiSource`
* Using `memo`

Checklist:


lib-multisrc/hiper/src/eu/kanade/tachiyomi/multisrc/hiper/Hiper.kt
AwkwardPeak7, line 1: consider migrating to `KeiSource` too

src/en/hiperdex/build.gradle.kts
```diff
@@ -14,7 +14,10 @@ keiyoushi {
     source {
         lang = "en"
         baseUrl {
-            custom("https://hiperdex.com")
+            mirrors(
+                "https://hiperdex.tv",
```
AwkwardPeak7: set it to tv domain by default instead of mirrors?

cuong-tran: the CONTRIBUTING.md said first entry is default when using `mirrors`

AwkwardPeak7: I meant keeping custom domain with tv as default

AwkwardPeak7 (approved):

PR #17654, opened 2026-07-16
Buondua: Upgrade to 1.6 & add preference for options to split pages or combine all pages into 1 gallery

cuong-tran:
Checklist:

### DIFF
--- a/src/all/buondua/src/eu/kanade/tachiyomi/extension/all/buondua/BuonDua.kt
+++ b/src/all/buondua/src/eu/kanade/tachiyomi/extension/all/buondua/BuonDua.kt
@@ -3,8 +3,6 @@ package eu.kanade.tachiyomi.extension.all.buondua
 import android.content.SharedPreferences
 import androidx.preference.PreferenceScreen
 import androidx.preference.SwitchPreferenceCompat
-import eu.kanade.tachiyomi.network.GET
-import eu.kanade.tachiyomi.network.awaitSuccess
 import eu.kanade.tachiyomi.source.ConfigurableSource
 import eu.kanade.tachiyomi.source.model.Filter
 import eu.kanade.tachiyomi.source.model.FilterList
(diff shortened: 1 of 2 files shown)


AwkwardPeak7 (commented):
consider migrating to `KeiSource` too

src/all/buondua/src/eu/kanade/tachiyomi/extension/all/buondua/BuonDua.kt
```diff
@@ -73,88 +86,197 @@ abstract class BuonDua : HttpSource() {
...
-                    if (isNotEmpty()) append("\n\n")
-                    append(downloadLinks)
-                }
-                if (password.isNotBlank()) {
-                    if (isNotEmpty()) append("\n\n")
-                    append(password)
-                }
-            }.toString().trim()
+        val updatedManga = if (fetchDetails) parseMangaDetails(document) else manga
+        val updatedChapters = if (fetchChapters) parseChapterList(document) else chapters
```
AwkwardPeak7: you have updated info always, return both unconditionally

src/all/buondua/src/eu/kanade/tachiyomi/extension/all/buondua/BuonDua.kt
```diff
@@ -73,88 +86,197 @@ abstract class BuonDua : HttpSource() {
...
+                        .build()
+                    client.get(pageUrl).asJsoup()
+                }
+            }
+            doc.select(pageListSelector).map { imgEl ->
+                imgEl.absUrl("src")
+            }
+        }.mapIndexed { index, url ->
+            Page(index, imageUrl = url)
         }
```
AwkwardPeak7: instead of fetching all pages together like this, set Page.url with the html page link and get imgurl in `getImageUrl`

AwkwardPeak7 (changes requested):

src/all/buondua/src/eu/kanade/tachiyomi/extension/all/buondua/BuonDua.kt
cuong-tran: Sorry I don't get it. The `page`(s) which are being fetch in parallel are html page, each of them have a dozen of images (`Page`) inside.

Original code was put each of `page` into a chapter, then when chapter is loaded, it will fetch `page`.
New code is fetching all `page`(s) to find the list of imgurl and put them all together in 1 merged Gallery.

AwkwardPeak7: Ah yes. In that case, you don't need to silently swallow exceptions and miss pages. Inline the async code without runcatching

AwkwardPeak7 (approved):

PR #17676, opened 2026-07-17
Update Contributing Guide for ext-lib 1.6

AwkwardPeak7:


AGENTS.md
```diff
@@ -0,0 +1,55 @@
...
+Before writing or modifying any extension, multisrc theme, or lib code in this repository, **read
+[CONTRIBUTING.md](CONTRIBUTING.md) in full**. It is the source of truth for conventions here and is
+updated frequently - do not rely on prior knowledge of this codebase or of Tachiyomi/Mihon
+extensions in general, since common patterns have changed.
+
+A few points that are easy to get wrong from stale training data:
+
+- **New sources must extend `KeiSource`** (`libVersion = "1.6"`), never `HttpSource` directly.
+  `HttpSource`/`libVersion = "1.4"` only exists in extensions/themes not yet migrated - see
+  [KeiSource](CONTRIBUTING.md#keisource).
```
darwin-256, line 12:
```suggestion
  [KeiSource](CONTRIBUTING.md#keisource).
  - **Metadata is injected via KSP:** Do NOT manually declare or `override val name`, `lang`, `id`, or `baseUrl` in your `@Source` class. These are defined using `source {}` blocks in `build.gradle.kts` and injected automatically.
- **`SourceFactory` is obsolete:** Do NOT implement `SourceFactory` to create multiple sources. Instead, add multiple `source {}` blocks in the extension's `build.gradle.kts`.
- **Generated Preferences:** Do NOT write manual `SharedPreferences` logic for base URL mirrors or custom user URLs. Use `baseUrl { mirrors(...) }` or `baseUrl { custom(...) }` inside the `source {}` block in `build.gradle.kts`.
- **Deeplinks are DSL-driven:** Declare URL intent filters using the `deeplink {}` block in `build.gradle.kts`. Do NOT modify `AndroidManifest.xml` or write manual intent filtering logic.
- **DTOs must be regular classes:** Do NOT use `data class` for `@Serializable` JSON/Protobuf DTOs (it bloats bytecode). Use a regular `class`, make fields `private` where possible, and only use `@SerialName` when the JSON key differs from the camelCase property name.
- **No local JSON/Proto instances:** Do NOT create `private val json: Json by injectLazy()`. Use the shared `keiyoushi.utils` helpers like `response.parseAs<T>()`, `toJsonRequestBody()`, or `parseAsProto<T>()`.
```

PR #17679, opened 2026-07-17
Komiic: update to 1.6

hualiong:
Checklist:


src/zh/komiic/src/eu/kanade/tachiyomi/extension/zh/komiic/Komiic.kt
```diff
@@ -50,98 +52,120 @@ abstract class Komiic :
...
-        listingQuery(variables).request()
-    }
+    // Customize
 
-    override fun searchMangaParse(response: Response) = parseListing(response.parseGraphQLAs())
+    private val SManga.id get() = url.substringAfterLast("/")
+    private val SChapter.id get() = url.substringAfterLast("/")
 
-    override fun getMangaUrl(manga: SManga) = baseUrl + manga.url
+    private suspend fun OkHttpClient.query(body: RequestBody) = newCall(POST("$baseUrl/api/query", headers, body)).awaitSuccess()
```
AwkwardPeak7:
```suggestion
    private suspend fun OkHttpClient.query(body: RequestBody) = post("$baseUrl/api/query", body)
```

src/zh/komiic/src/eu/kanade/tachiyomi/extension/zh/komiic/Komiic.kt
```diff
@@ -50,98 +52,120 @@ abstract class Komiic :
...
+        val data = response.parseGraphQLAs<DataDto>()
 
-    private fun refreshToken(chain: Interceptor.Chain) {
-        client.cookieJar.loadForRequest(chain.request().url).find { it.name == "komiic-access-token" }?.let {
-            val payload = Base64.decode(it.value.split(".")[1], Base64.DEFAULT).decodeToString()
-            if (System.currentTimeMillis() + 3600_000 >= payload.parseAs<JwtPayload>().exp * 1000) {
-                val response = chain.proceed(POST("$baseUrl/auth/refresh", headers)).apply { close() }
-                if (!response.isSuccessful) throw IOException("刷新 Token 失敗：HTTP ${response.code}")
+        val sManga = if (fetchDetails) data.comicById!!.toSManga() else manga
+        val sChapters = if (fetchChapters) {
```
AwkwardPeak7, line 135: you have updated info always, so return both unconditionally

hualiong: `mangaQuery` dynamically fetches the corresponding comic details and chapters based on the passed `fetchDetails` and `fetchDetails` parameters, so conditional judgment must be used. Otherwise, `NullPointerException` will occur.

```kotlin
fun mangaQuery(id: String, fetchDetails: Boolean, fetchChapters: Boolean): RequestBody {
    val query = buildString {
        append($$"query mangaQuery($comicId: ID!) {")
        if (fetchDetails) append($$"comicById(comicId: $comicId) ", COMIC_BODY)
        if (fetchChapters) append($$"chaptersByComicId(comicId: $comicId) { id serial type size dateCreated }")
        append('}')
    }.trimIndent()
    val variables = buildJsonObject { put("comicId", id) }
    return graphQLBody(query, "mangaQuery", variables)
}
```

AwkwardPeak7: Ah I missed that. 
Tho since this is graphql, why not fetch both?

hualiong: en..... this is more about personal coding habits. I don’t really like fetching comic details along with chapter data when I only need the chapters — it feels kinda redundant.

Also, since these are two different types of data, fetching them together might mean the server has to query two database tables or something. I think fetching them separately would be faster.

AwkwardPeak7: Fair enough 

also just a fyi, app is expected to update both of them regardless of the passed flag.

AwkwardPeak7 (approved):

PR #17689, opened 2026-07-18
H-Comic: update to 1.6

hualiong:
For this source, implementing `HttpSource` is more convenient than `KeiSource`.

Checklist:

### DIFF
--- a/src/zh/hcomic/src/eu/kanade/tachiyomi/extension/zh/hcomic/HComic.kt
+++ b/src/zh/hcomic/src/eu/kanade/tachiyomi/extension/zh/hcomic/HComic.kt
@@ -6,56 +6,67 @@ import eu.kanade.tachiyomi.source.model.Page
 import eu.kanade.tachiyomi.source.model.SChapter
 import eu.kanade.tachiyomi.source.model.SManga
 import eu.kanade.tachiyomi.source.model.SMangaUpdate
-import eu.kanade.tachiyomi.source.online.HttpSource
 import keiyoushi.annotation.Source
 import keiyoushi.network.get
+import keiyoushi.source.KeiSource
 import keiyoushi.utils.firstInstance
 import keiyoushi.utils.int
 import keiyoushi.utils.long
-import kotlinx.serialization.json.jsonPrimitive
+import keiyoushi.utils.string
+import kotlinx.serialization.json.JsonElement
+import okhttp3.HttpUrl
 import okhttp3.HttpUrl.Companion.toHttpUrl
 
 @Source
-abstract class HComic : HttpSource() {
-
-    override val supportsLatest = true
+abstract class HComic : KeiSource() {
 
     private val imgUrl = "https://h-comic.link/api"
 
     // Popular (Random)
     override suspend fun getPopularManga(page: Int): MangasPage {
-        val result = client.get("$baseUrl/random/__data.json", headers).parseAsMangaList(0)
+        val result = client.get("$baseUrl/random/__data.json").parseAsMangaList(0)
         return MangasPage(result.first.map { it.toSManga(imgUrl) }, true)
     }
 
     // Latest
     override suspend fun getLatestUpdates(page: Int): MangasPage {
-        val result = client.get("$baseUrl/__data.json?page=$page", headers).parseAsMangaList(page)
+        val result = client.get("$baseUrl/__data.json?page=$page").parseAsMangaList(page)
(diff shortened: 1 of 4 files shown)


src/zh/hcomic/src/eu/kanade/tachiyomi/extension/zh/hcomic/HComic.kt
```diff
@@ -1,104 +1,80 @@
...
     private val imgUrl = "https://h-comic.link/api"
 
     // Popular (Random)
-
-    override fun popularMangaRequest(page: Int) = GET("$baseUrl/random/__data.json")
-
-    override fun popularMangaParse(response: Response): MangasPage {
-        val result = response.parseAsMangaList(0)
+    override suspend fun getPopularManga(page: Int): MangasPage {
+        val result = client.get("$baseUrl/random/__data.json", headers).parseAsMangaList(0)
```
AwkwardPeak7: headers are automatically passed via context parameters

src/zh/hcomic/src/eu/kanade/tachiyomi/extension/zh/hcomic/HComic.kt
```diff
@@ -1,104 +1,80 @@
...
+            client.get("${getMangaUrl(manga)}/__data.json", headers).parseAsManga().toSManga(imgUrl)
+        } else {
+            manga
+        }
+
+        val sChapter = SChapter.create().apply {
+            url = sManga.url
+            name = sManga.title
+            date_upload = sManga.memo["timestamp"]!!.jsonPrimitive.long
+            scanlator = sManga.memo["category"]!!.jsonPrimitive.content
```
AwkwardPeak7:
```suggestion
            date_upload = sManga.memo["timestamp"]!!.long
            scanlator = sManga.memo["category"]!!.string
```

src/zh/hcomic/src/eu/kanade/tachiyomi/extension/zh/hcomic/HComic.kt
```diff
@@ -1,104 +1,80 @@
...
-    override fun fetchPageList(chapter: SChapter): Observable<List<Page>> {
-        val info = chapter.url.split(':')
-        return Observable.just(
-            List(info.last().toInt()) {
-                Page(it, imageUrl = "$imgUrl/${info.first()}/pages/${it + 1}")
-            },
-        )
+    override suspend fun getPageList(chapter: SChapter): List<Page> {
+        val pageCount = chapter.memo["numPages"]!!.jsonPrimitive.int
+        val mediaId = chapter.memo["mediaId"]!!.jsonPrimitive.content
```
AwkwardPeak7:
```suggestion
        val pageCount = chapter.memo["numPages"]!!.int
        val mediaId = chapter.memo["mediaId"]!!.string
```

AwkwardPeak7 (commented):
please implement using `KeiSource`, it provides a standardized client and headers configuration. (e.g compression and caching)

if you can't or don't want to implement related manga method, you can set `supportsRelatedMangas` to false

hualiong:
> please implement using `KeiSource`, it provides a standardized client and headers configuration. (e.g compression and caching)
> 
> if you can't or don't want to implement related manga method, you can set `supportsRelatedMangas` to false

I'm a bit puzzled as to why `getMangaUrl` and `getChapterUrl` don't have default implementations—I have to implement these methods myself, which is a bit of a hassle.

Also, even if I set `supportsRelatedMangas` to `false`, I still have to manually throw `UnsupportedOperationException` for `getMangaByUrl` and `fetchRelatedMangaList` — that's a bit tedious too.😞

AwkwardPeak7:
> I'm a bit puzzled as to why getMangaUrl and getChapterUrl don't have default implementations

We have been recommending only saving part of url e.g slug or internal id as manga.url/chapter.url and putting other data in memo. so urls are more stable. default implementation of `baseUrl + manga.url` assumes it is a relative url.
Tho I get what you mean, probably should provide it as default for majority of old sources which still save relative urls

> getMangaByUrl

I suppose we can disable it by default. The motivation behind it was allowing users to paste manga url in browse to search for manga, as many sites have very bad search.

> fetchRelatedMangaList

It's default implementation on Komikku is to fetch `mangaDetailsRequest` and parsing with `popularMangaParse` both of which are deprecated and not implemented now, also that assumption is wrong that most sites will have a popular like listing on manga page.

Thanks for you feedback!

AwkwardPeak7 (approved):

PR #17730, opened 2026-07-19
Bump dependencies and minsdk

AwkwardPeak7:
:nyaboom:

- Bump dependencies
- Bump minSdk to 26 (Android 8.0)
  - Unlocks native `java.time` APIs.
  - Fixes a `kotlinx-serialization` default method crash on versions > 1.7.3.
    <details>
    <summary>Technical details</summary>
    <br>
    Bumping minSdk to 24 or higher fixes KT-84952 — a NoClassDefFoundError: GeneratedSerializer$-CC crash that started with kotlinx.serialization 1.8.0, which switched to compiling GeneratedSerializer with real Java 8 default interface methods instead of the old DefaultImpls shim. On Android versions below API 24, ART doesn't natively support default interface methods, so d8 has to desugar them: it rewrites calls to a default method into a call on a synthesized companion class (Interface$-CC), built by extracting the method body straight out of the interface's own class file at dex time. That's precisely what breaks for us — since kotlinx.serialization is a compileOnly dependency, its GeneratedSerializer interface is only visible to d8 for type-resolution, never handed over as program input to actually dex, so d8 rewrites the call site in our generated $serializer classes but has nothing to build the $-CC companion from, leaving a dangling reference that throws at runtime. Once minSdk >= 24 (Android 7.0 Nougat), ART natively supports default interface methods, so d8 skips desugaring entirely and leaves the call as a plain invokeinterface resolved directly against the interface's default implementation at runtime — no companion class needed, no dependency on the interface being bundled, and the crash never occurs.
    </details>
  - Note: This effectively requires all host apps to be minsdk 26+.
- Enable R8 optimizations

### DIFF
--- a/common/proguard-rules.pro
+++ b/common/proguard-rules.pro
@@ -1,5 +1,5 @@
 #-dontobfuscate
--dontoptimize
+#-dontoptimize
 
 ## Partially based on https://android.googlesource.com/platform/tools/base/+/refs/heads/mirror-goog-studio-main/build-system/gradle-core/src/main/resources/com/android/build/gradle/proguard-common.txt
 
--- a/gradle/kei.versions.toml
+++ b/gradle/kei.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 # Used in [/gradle/build-logic/src/main/kotlin/AndroidBasePlugin.kt]
-android-sdk-min = "21"
+android-sdk-min = "26"
 android-sdk-compile = "37"
 android-sdk-target = "37"
 java = "11"
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,9 +1,9 @@
 [versions]
 android-gradle = "9.3.0"
-coroutines = "1.10.2"
-serialization = "1.7.3"
+coroutines = "1.11.0"
+serialization = "1.11.0"
 junit = "4.13.2"
-kotlin-gradle = "2.4.0"
+kotlin-gradle = "2.4.10"
 ksp = "2.3.9"
 kotlinpoet = "2.3.0"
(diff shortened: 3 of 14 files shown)


Secozzi:
Did you check out `en/mgreadio`, among others, for the new java.time API?

common/proguard-rules.pro
```diff
@@ -1,5 +1,5 @@
 #-dontobfuscate
--dontoptimize
+#-dontoptimize
```
FourTOne5, line 3: remove?

FourTOne5 (approved):

common/proguard-rules.pro
AwkwardPeak7: Kept for documentation, if we ever need to quickly disable optimisation or obfuscation

FourTOne5 (approved):

Ashvith10:
@AwkwardPeak7 this commit seems to have broken extension for me, since I am on Android 7.1.2. Is it possible for me to pin to an older commit so that I can continue using some of the working extensions?

Smol-Ame:
The project no longer supports anything under Android 8, and any apps that can still be installed on Android 7 or below is considered incompatible/outdated.

Ashvith10:
I know this is not a fix, but I've pinned to https://raw.githubusercontent.com/keiyoushi/extensions/7d5052fb895d086ae2ec6e3cca861146ee3ea0ec/index.min.json as a temporary measure. It seems to be working in the sense that I can install APKs now, but no longer update it anymore, which is fine for me.

AwkwardPeak7:
while you can use it for now, we don't provide any guarantee that apk repo will continue to have old commits

PR #17738, opened 2026-07-19
MangaLivre: handle rotating reading-gate signatures

atiladpribeiro:
## Summary

- update the MangaLivre reading gate to match the current `toonlivre.net` request signature
- extract both fixed, byte-array-obfuscated signatures and the previous time-windowed signature format from the live bundle
- refresh rotating gate and encryption constants after a 403, rotate the verification cookie, and keep the WebView primer as the final recovery path
- isolate the public API verification cookie so stale session cookies cannot poison reader requests
- parse API DTOs without generated serializer ABI coupling
- bump the extension to `1.4.80`

## Root cause

The site changed `x-toon-signature` from a time-windowed SHA-256 value to fixed tokens encoded as byte arrays in the minified frontend bundle. The extension kept sending the previous signature, so chapter page-list requests were rejected with HTTP 403 even though the website itself still worked.

## Validation

- `:src:pt:mangalivre:clean`
- `:src:pt:mangalivre:assembleDebug`
- `:src:pt:mangalivre:lintDebug`
- live catalog, details, chapter-list, encrypted page-list, and image loading checks against `toonlivre.net`
- opened three separate source chapters (82, 18, and 210); real pages rendered and the reader remained open for more than 20 seconds with no 403, page-list failure, or crash in logs

Checklist:

- [x] Updated `baseVersionCode` in `build.gradle.kts` (not applicable: this is not a multisrc theme change)
- [x] Referenced all related issues in the PR body (none found)
- [x] Have explicitly kept the `id` if a source's name or language were changed (not applicable)
- [x] Have tested the modifications by compiling and running the extension through Android Studio-compatible Gradle tasks and on-device runtime checks
- [x] Have removed `web_hi_res_512.png` when adding a new extension (not applicable)

### DIFF
--- a/src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/Dto.kt
+++ b/src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/Dto.kt
@@ -3,8 +3,19 @@ package eu.kanade.tachiyomi.extension.pt.mangalivre
 import eu.kanade.tachiyomi.source.model.Page
 import eu.kanade.tachiyomi.source.model.SChapter
 import eu.kanade.tachiyomi.source.model.SManga
-import org.json.JSONArray
-import org.json.JSONObject
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonNull
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.booleanOrNull
+import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.contentOrNull
+import kotlinx.serialization.json.jsonArray
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
+import kotlinx.serialization.json.longOrNull
+import kotlinx.serialization.json.put
 import java.text.Normalizer
 
 class WrapperDto(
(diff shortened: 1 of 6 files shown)


src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/Dto.kt
AwkwardPeak7, line 1: use kotlinx serialization instead of org.json

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
```diff
@@ -179,18 +180,21 @@ abstract class MangaLivre :
...
         }
-        return parseAs<T>()
+        return runCatching { JSONObject(responseBody) }
+            .getOrElse { throw IOException(NON_JSON_MESSAGE, it) }
     }
 
     companion object {
         private const val ALTERNATIVE_TITLE_PREF = "alternativeTitlePref"
-        private const val MAX_PEEK = 1024L
+        private const val BROWSER_USER_AGENT =
```
AwkwardPeak7: hardcoding a useragent is not a good idea

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/Dto.kt
atiladpribeiro: Updated in 988b2e446. All JSON parsing now uses the kotlinx.serialization JsonObject/JsonArray tree API, and org.json has been removed from the source. This keeps parsing compatible without relying on generated serializers.

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
atiladpribeiro: Updated in 988b2e446. The request User-Agent is now obtained at runtime from WebSettings.getDefaultUserAgent(applicationContext), with the host default left in place if it is unavailable. No browser version is hardcoded.

AwkwardPeak7 (approved):

AwkwardPeak7:
@atiladpribeiro It appears users can't install this version, crash on startup...

atiladpribeiro:
Confirmed. The startup/catalog failure was caused by generated kotlinx.serialization serializers calling a newer runtime ABI. PR #17855 replaces them with the kotlinx.serialization JSON tree API, updates the current Rabbit key derivation, and adds bounded live-bundle recovery for future rotations. The final APK was validated through catalog, details, chapter lists, and multiple online chapters, and the upstream CI check is passing.

PR #17804, opened 2026-07-21
fix(mangadar): support new rows format and iterate all x-data containers

hamedhani1998:
## Summary

- Iterate all div[x-data] containers to find chapter data instead of single selectFirst`n- Support new compact array format: 
ows: [[id, number, url, timestamp], ...]`n- Keep legacy chapters: [{...}] format as fallback
- Add parseRowsJson helper for compact array parsing


src/ar/mangadar/src/eu/kanade/tachiyomi/extension/ar/mangadar/MangaDar.kt
```diff
@@ -200,23 +202,60 @@ abstract class MangaDar : KeiSource() {
...
+    private fun parseRowsJson(rowsJson: String): List<SChapter> {
+        // Parse the compact array format: [[id, number, url, timestamp, num], ...]
+        val rows = rowsJson.parseAs<JsonArray>()
+        return rows.mapNotNull { element ->
+            val row = element as? JsonArray ?: return@mapNotNull null
+            if (row.size < 4) return@mapNotNull null
+
+            val chapterNumber = (row[1] as? JsonPrimitive)?.content ?: ""
+            val chapterUrl = (row[2] as? JsonPrimitive)?.content ?: ""
+            val timestamp = (row[3] as? JsonPrimitive)?.content?.toLongOrNull() ?: 0L
```
Starmania, line 241: `JsonNull` is a subclass of `JsonPrimitive`, and `JsonNull.content == "null"`. So a null field does not fall through to the elvis branch — it yields the literal string `"null"`:

- `row[1]` null → `name = "الفصل null"`, `chapter_number = 0f`
- `row[2]` null → `isBlank()` is false, so a chapter with `url = "null"` is emitted
- `row[3]` null → `"null".toLongOrNull()` is null → falls back to `0L`, so this one is fine by accident

This is not too much hypothetical. The site's own sort function guards for exactly this case:

```js
sortKey(c) { return (c.num === null || c.num === undefined) ? -Infinity : c.num; }
```

To fix this, add `import kotlinx.serialization.json.contentOrNull` at the top
```suggestion
            val chapterNumber = (row[1] as? JsonPrimitive)?.contentOrNull.orEmpty()
            val chapterUrl = (row[2] as? JsonPrimitive)?.contentOrNull.orEmpty()
            val timestamp = (row[3] as? JsonPrimitive)?.contentOrNull?.toLongOrNull() ?: 0L
```

src/ar/mangadar/src/eu/kanade/tachiyomi/extension/ar/mangadar/MangaDar.kt
Starmania: There is a pre-existing bug to this PR:

In the Popular tab — and equally in a filter-only browse, since both go through `parseMangaListPage` — a phantom entry appears: title `مكتبة المانجا`, no thumbnail, and opening it 404s because its URL resolves to `/manga/manga`.

The live listing contains 4 `<a href="https://mangadar.com/manga/">` navigation links (labels `مكتبة المانجا`, `تصفح المكتبة`, `المكتبة`). `distinctBy { it.url }` collapses them into one and the first wins. There's no `` inside, so the title falls through to the `.font-semibold` branch and picks up the nav label, and the thumbnail comes out empty. `getMangaUrl()` then builds `$baseUrl/manga/manga` → 404.

Filtering the link based on the shape works. For example:

```suggestion
            val segments = element.absUrl("href").toHttpUrl().pathSegments.filter { it.isNotEmpty() }
            if (segments.size != 2 || segments[0] != "manga") return@mapNotNull null
            val slug = segments[1]
```

Starmania (commented):
Please keep the legacy `chapters:` branch only if you can point at a title that still uses it

`MangaDar.kt:221-231`, and `Dto.kt:22-36`

I couldn't find the old format anywhere. Sampled 7 titles spanning 2 to 541 chapters — 877 rows in total:

| title | chapters | format |
|---|---|---|
| `all-hail-the-sect-leader` | 541 | `rows:` |
| `blue-box` | 247 | `rows:` |
| `do-your-best-to-regret` | 40 | `rows:` |
| `hands-off-my-cleaner` | 35 | `rows:` |
| `green-skin` | 10 | `rows:` |
| `bungou-stray-dogs-dazai-wo-hirotta-hi` | 2 | `rows:` |
| `when-love-calls-upon-me` | 2 | `rows:` |

Every one of those pages has exactly 10 `div[x-data]`, with `rows:` always in container index 7 and every row of length 5. A `\bchapters:` search matches **nowhere** on any of the seven pages.

That makes the legacy branch unreachable, and `ChapterDto` unreachable with it — `Dto.kt:22-36` has exactly one consumer, the `parseAs<List<ChapterDto>>()` call at `:228`.

Seven titles is a sample, not a proof, so: if you know of a title that still serves `chapters:`, name it and the fallback is justified — keep it. Otherwise please delete the legacy branch and `ChapterDto` along with it. A fallback for a shape that can't be reproduced is speculative, and it's the kind of code that quietly rots. (`SearchResponse` / `SearchMangaItemDto` are unaffected.)

hamedhani1998:
@AwkwardPeak7

src/ar/mangadar/build.gradle.kts
```diff
@@ -6,7 +6,7 @@ plugins {
 
 keiyoushi {
     name = "MangaDar"
-    versionCode = 1
+    versionCode = 3
```
Starmania: Change back to 2

src/ar/mangadar/src/eu/kanade/tachiyomi/extension/ar/mangadar/MangaDar.kt
AwkwardPeak7: unresolved

AwkwardPeak7 (approved):

PR #17815, opened 2026-07-21
[RU]AllHentai add support of site mirror

abubaca4:
AllHentai has an official mirror (created by the owner of the main site) that does not support login/authorization. However, it displays certain titles that require a paid subscription on the main site and is accessible in Russia without using VPNs or bypass tools.

The current extension code requires authorization, so I added the mirror domain to the exception list for this check. Alternatively, it might be worth adding a toggle setting to allow switching this behavior if preferred.

For reference: The mirror does not load without a Referer header containing either the mirror's own URL or a search engine URL. This was implemented to make it harder to detect blocked content within Russia. If you want to test or open it in a browser, the easiest way is to follow [this](https://www.google.com/search?q=x.ahen.me) link and open the search result named Ahen (not "AllHentai").

Checklist:

### DIFF
--- a/src/ru/allhentai/src/eu/kanade/tachiyomi/extension/ru/allhentai/AllHentai.kt
+++ b/src/ru/allhentai/src/eu/kanade/tachiyomi/extension/ru/allhentai/AllHentai.kt
@@ -7,7 +7,7 @@ import keiyoushi.annotation.Source
 @Source
 abstract class AllHentai : GroupLe() {
 
-    override val isNeedAuth = true
+    override val isNeedAuth = baseUrl != "https://x.ahen.me"
 
     override fun getFilterList() = FilterList(
         OrderBy(),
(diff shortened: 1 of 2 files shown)


src/ru/allhentai/src/eu/kanade/tachiyomi/extension/ru/allhentai/AllHentai.kt
```diff
@@ -7,7 +7,7 @@ import keiyoushi.annotation.Source
 @Source
 abstract class AllHentai : GroupLe() {
 
-    override val isNeedAuth = true
+    override val isNeedAuth = baseUrl != "https://x.ahen.me"
```
AwkwardPeak7:
```suggestion
    override val isNeedAuth get() = baseUrl != "https://x.ahen.me"
```

AwkwardPeak7 (approved):

PR #17855, opened 2026-07-22
MangaLivre: handle chapter route tokens

atiladpribeiro:
## Summary

- request the official per-chapter route token with the current reading-gate signature
- attach `x-toon-route-token` only to chapter page-list requests
- cache tokens using the server expiry and refresh them after a chapter 404
- refresh both the seed and route token after a 403
- bump the extension to `1.4.87`

## Root cause

The site now requires a short-lived token from `/api/chapter-token/{mangaId}/{chapterId}` for each chapter API request. The chapter endpoint deliberately returns 404 when `x-toon-route-token` is missing or stale.

## Performance

The token is fetched once per chapter and cached until shortly before the server-provided expiry. Normal catalog, search, details, and image requests do not perform this extra call. Recovery is bounded to one retry and does not use a WebView.

## Validation

- `:src:pt:mangalivre:clean`
- `:src:pt:mangalivre:assembleDebug`
- `:src:pt:mangalivre:lintDebug`
- live catalog, details, and chapter list against `toonlivre.net`
- live page-list and image rendering for chapters 129 and 128 of *Mestre Espadachim Criado Pelas Estrelas*
- no HTTP 403, HTTP 404, ABI error, or crash in the final runtime logs

Checklist:

- [x] Updated `versionCode`
- [x] Kept the source name, language, and ID unchanged
- [x] Kept all added requests on `toonlivre.net`
- [x] Used kotlinx.serialization
- [x] Did not hardcode a User-Agent
- [x] Tested the complete source-to-reader path
- [x] This PR is AI-assisted; I reviewed the changes and validated the behavior

### DIFF
--- a/src/pt/mangalivre/build.gradle.kts
+++ b/src/pt/mangalivre/build.gradle.kts
@@ -6,7 +6,7 @@ plugins {
 
 keiyoushi {
     name = "Manga Livre"
-    versionCode = 86
+    versionCode = 87
     contentWarning = ContentWarning.SAFE
     libVersion = "1.4"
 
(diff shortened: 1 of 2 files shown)


AwkwardPeak7:
I believe this is outdated now(?)

PR #17887, opened 2026-07-23
MangaBox: Migrate to 1.6

tigeryu8900:
Checklist:


lib-multisrc/mangabox/src/eu/kanade/tachiyomi/multisrc/mangabox/MangaBox.kt
```diff
@@ -38,28 +41,19 @@ import okio.Buffer
...
     ConfigurableSource {
 
-    protected open val dateFormat: SimpleDateFormat = SimpleDateFormat(
-        "yyyy-MM-dd'T'HH:mm:ss.SSSSSS'Z'",
-        Locale.ROOT, // Changed to Locale.ROOT as per CONTRIBUTING.md rules
-    ).apply {
-        timeZone = TimeZone.getTimeZone("UTC")
-    }
-
     override val supportsLatest = true
```
AwkwardPeak7: default in KeiSource

lib-multisrc/mangabox/src/eu/kanade/tachiyomi/multisrc/mangabox/MangaBox.kt
```diff
@@ -219,9 +210,9 @@ abstract class MangaBox :
 
     open fun popularMangaSelector() = ":is(div.truyen-list > div.list-truyen-item-wrap, div.comic-list > .list-comic-item-wrap):has(a[data-id])"
 
-    override fun popularMangaRequest(page: Int): Request = GET("$baseUrl/$popularUrlPath$page", headers)
+    open suspend fun popularMangaResponse(page: Int): Response = client.get("$baseUrl/$popularUrlPath$page", headers)
```
AwkwardPeak7: inline it

lib-multisrc/mangabox/src/eu/kanade/tachiyomi/multisrc/mangabox/MangaBox.kt
```diff
@@ -238,9 +231,9 @@ abstract class MangaBox :
 
     open fun latestUpdatesSelector() = popularMangaSelector()
 
-    override fun latestUpdatesRequest(page: Int): Request = GET("$baseUrl/$latestUrlPath$page", headers)
+    open suspend fun latestUpdatesResponse(page: Int): Response = client.get("$baseUrl/$latestUrlPath$page", headers)
```
AwkwardPeak7: same here

lib-multisrc/mangabox/src/eu/kanade/tachiyomi/multisrc/mangabox/MangaBox.kt
AwkwardPeak7: perhaps we should use `manga.memo` to store full url in case it is on different domain

lib-multisrc/mangabox/src/eu/kanade/tachiyomi/multisrc/mangabox/MangaBox.kt
```diff
@@ -383,12 +383,12 @@ abstract class MangaBox :
 
     // ============================= Chapters ==============================
 
-    override fun chapterListRequest(manga: SManga): Request {
+    open suspend fun chapterListResponse(manga: SManga): Response {
         val slug = manga.url.split("/").last()
-        return GET("${apiChapterListUrl.replace("__SLUG__", slug)}?limit=$CHAPTER_LIST_TAKE&offset=0", headers)
+        return client.get("${apiChapterListUrl.replace("__SLUG__", slug)}?limit=$CHAPTER_LIST_TAKE&offset=0", headers)
```
AwkwardPeak7: headers are passed by default

lib-multisrc/mangabox/src/eu/kanade/tachiyomi/multisrc/mangabox/MangaBox.kt
```diff
@@ -383,12 +383,12 @@ abstract class MangaBox :
...
     }
 
-    override fun chapterListParse(response: Response): List<SChapter> {
+    open fun parseChapterList(response: Response): List<SChapter> {
```
AwkwardPeak7: inline/merge these methods, also maybe pass `limit=-1` to get all chapters? that is what the site seems to be using now

lib-multisrc/mangabox/src/eu/kanade/tachiyomi/multisrc/mangabox/MangaBox.kt
AwkwardPeak7, line 472: migrate this section to coroutines too

tigeryu8900:
Currently for the 1.6 version of MangaBox, the requests to the image server results in 429 responses even after adding headers in the interceptor. Interestingly, this doesn't happen when debugging. I'll mark this as draft until I find a fix for this.

tigeryu8900:
I've fixed some issues I encountered earlier, but it is incredibly slow. The 1.4 version works without problem, and the pages load instantly even with the merge images settings turned on. The 1.6 version on the other hand takes a few seconds to load the pages when the merge images settings is turned off, and it is much more likely to get a 429 response. With the merge images settings turned on, it can take more than 20 seconds just to load the image sizes for the merge images logic.

I speculate that this has something to do with the internal interceptors 1.6 adds to the network stack.

AwkwardPeak7:
Can you check the network via mitmproxy?

tigeryu8900:
I fixed the problem.

I checked the network and found out that the `Origin` header was causing CloudFlare cache misses, which significantly increased the response time (and sometimes caused timeouts). Removing the `Origin` header fixed this problem.

lib-multisrc/mangabox/src/eu/kanade/tachiyomi/multisrc/mangabox/MangaBox.kt
```diff
@@ -305,8 +292,11 @@ abstract class MangaBox :
 
     private fun mangaFromElement(element: Element, urlSelector: String = "h3 a"): SManga = SManga.create().apply {
         val urlElement = element.selectFirst(urlSelector)!!
-        url = urlElement.attr("abs:href")
-            .substringAfter(baseUrl) // intentionally not using setUrlWithoutDomain
+        val fullUrl = urlElement.attr("abs:href")
+        memo = buildJsonObject {
+            put("fullUrl", fullUrl)
+        }
+        url = fullUrl.substringAfter(baseUrl) // intentionally not using setUrlWithoutDomain
```
AwkwardPeak7: save the relative url/slug

lib-multisrc/mangabox/src/eu/kanade/tachiyomi/multisrc/mangabox/MangaBox.kt
```diff
@@ -323,22 +313,24 @@ abstract class MangaBox :
...
     private fun checkForRedirectMessage(document: Document) {
         if (document.select("body").text().startsWith("REDIRECT :")) {
             throw Exception("Source URL has changed")
         }
     }
 
-    override fun mangaDetailsParse(response: Response): SManga = mangaDetailsParse(response.asJsoup())
-
-    open fun mangaDetailsParse(document: Document): SManga = SManga.create().apply {
+    open fun parseMangaDetails(document: Document): SManga = SManga.create().apply {
```
AwkwardPeak7, line 340: need to save/update memo

AwkwardPeak7 (approved):

PR #17938, opened 2026-07-25
Noyacg: update to extlib 1.6

hualiong:
I think the current auto-generated mirror url configuration can't migrate login status (cookies) when users switch, which is a bit of a pain point. Could `KeiSource` use `CookieJar` to allow all hosts to share cookies? Just like I've implemented it on [NoyAcg](https://github.com/keiyoushi/extensions-source/pull/17938/changes#diff-d90f97ed1152963135d25e2a326e3889c93c2fbdafbc3ed7601346cba5107e70) right now.

I haven't considered the situation of other sources, so this is just a small suggestion.

Checklist:

### DIFF
--- a/src/zh/noyacg/src/eu/kanade/tachiyomi/extension/zh/noyacg/Dto.kt
+++ b/src/zh/noyacg/src/eu/kanade/tachiyomi/extension/zh/noyacg/Dto.kt
@@ -3,6 +3,7 @@ package eu.kanade.tachiyomi.extension.zh.noyacg
 import eu.kanade.tachiyomi.source.model.SManga
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonNames
 import java.text.SimpleDateFormat
 import java.util.Locale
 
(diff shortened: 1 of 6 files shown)


src/zh/noyacg/src/eu/kanade/tachiyomi/extension/zh/noyacg/Dto.kt
```diff
@@ -3,34 +3,29 @@ package eu.kanade.tachiyomi.extension.zh.noyacg
 import eu.kanade.tachiyomi.source.model.SManga
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonNames
 import java.text.SimpleDateFormat
 import java.util.Locale
 
 const val LISTING_PAGE_SIZE = 20
 val DATE_FORMAT = SimpleDateFormat("yyyy-MM-dd HH:mm:ss", Locale.ENGLISH)
```
AwkwardPeak7, line 11: please use java.time apis next time

AwkwardPeak7 (approved):

AwkwardPeak7:
> Could KeiSource use CookieJar to allow all hosts to share cookies

I don't this this is an assumption the framework can make, it depends on the specific source

PR #18036, opened 2026-07-27
Catharsis World: add login to read chapters

yuuues:
Catharsis World started gating chapter pages behind registration. Without a session,
any chapter URL answers HTTP 404 with the theme's "page not found" template, so the
HTML never carries the `chapter-protector` payload that `Madara.pageListParse` needs
to resolve the page list — reading is currently broken for everyone.

This adds username/password preferences and a login interceptor.

### How the site behaves

Verified against `catharsisworld.dig-it.info`:

| Request | Result |
| --- | --- |
| `POST /wp-login.php` without `testcookie` | reaches authentication (`div#login_error` reports the unknown username) |
| `POST /wp-login.php` with `testcookie=1` and no prior cookie | `div#login_error`: "cookies are blocked or not allowed" |
| chapter page, no session | 404, no `protectornonce='` |
| chapter page, with session | 200, `protectornonce='` present |

### Implementation notes

- `testcookie` is deliberately **not** submitted. WordPress only checks the test
  cookie when the form sends that field, so omitting it authenticates directly
  instead of needing a prior `GET /wp-login.php` just to pick the cookie up.
- The login POST sits behind a `ReentrantLock` with a double check on the session
  cookie: page prefetching would otherwise fire one login per in-flight request.
- Only 404s on the base host under `/serie/` are translated into a "log in" error.
  Anywhere else — a missing thumbnail, another host — a 404 is left as a real 404.
- A 404 while credentials are configured triggers one forced re-login and retry.
  Since the site answers 404 rather than redirecting to the login form, that case is
  indistinguishable from an expired cookie, and this recovers from a stale session or
  a changed password without the user having to clear cookies by hand.
- Login failures surface the site's own `div#login_error` text. It is already in
  Spanish and also covers the notice from the site's Simba TFA plugin, so an account
  with 2FA enabled reports why it failed instead of failing silently.

### One deliberate deviation from CONTRIBUTING.md

The "Empty Lists vs. Exceptions" rule says a locked or empty chapter should return
`emptyList()` rather than throwing a hardcoded exception, so the app can show a
localized error. This PR throws `IOException` instead. The reasoning:

- The 404 happens at the HTTP layer, so `pageListParse` never runs. There is no empty
  page list to return — only a request that failed.
- `emptyList()` surfaces as "no pages found", which gives the reader no hint that an
  account is required. Pointing the user at the credential settings is the entire
  point of the change.
- `src/ar/duskoryvile` already throws `IOException` with a "log in via extension
  settings" message for the same situation, so there is precedent for auth gates
  being treated differently from empty chapters.

Happy to switch to a plain `emptyList()` if you would rather keep the rule strict.

Checklist:

🤖 I was asked to make it possible to register a username and password in order to log
in on the Catharsis World source, and to follow how other extensions that ask for
values do it. This pull request was opened by an AI agent (Claude Code). The site's
login protocol was verified against the live site, and the repository owner built the
extension and confirmed on-device that reading works with credentials configured, and
reviewed the diff before checking the AI-assisted box above.


yuuues:
image
Proof on Kaizoku.NET also, that the extension is working against Catharsis World

Issue #14571, opened 2026-04-06
Source request (AR): ProChan (`procomic.net` & `procomic.pro`)


malekmahmoud2512007-hash:
### Source information

ProChan 

### Source language

Arabic

### Source new URL

https://procomic.net/ , https://procomic.pro

### Other details

Please add the new URL

### Acknowledgements

- [x] I have updated the extension to the latest version (as listed on the [extensions listing](https://keiyoushi.github.io/extensions/) page).
- [x] I have checked WebView to see the source's main URL is not up-to-date, or has to redirect into the new URL.
- [x] I have searched [existing issues](https://github.com/keiyoushi/extensions-source/issues/) both open & closed, and confirm that this is a new unreported issue.
- [x] I have written a short but informative title.
- [x] I will correctly fill out all of the requested information in this form.

### 

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/keiyoushi/extensions-source/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

MikeZeDev:
https://github.com/keiyoushi/extensions-source/issues/13889

MO350AZ:
Some findings from testing ProChan after deletion. I could not finish a working patch, but these observations may help whoever restores it.

What was confirmed:

1. Search still works through:
    /api/public/series/search

This endpoint still returns valid search results and appears usable for Mihon search parsing.

2. /series behavior changed.

Request observed:
GET /series

Site now uses Next.js App Router / RSC behavior.

Using:
rsc: 1

and parsing through:
extractNextJsRsc()

appears necessary for homepage content.

The old API-only assumption no longer matches site behavior.

3. /updates is different from /series.

Browser/frontend inspection showed update loading logic using:

/content/latest-chapter-updates?batch=0

with later requests incrementing batch:

batch=1
batch=2
etc.

This appears to drive infinite scrolling.

Observed frontend logic:

* first batch loads about 18 items
* later batches load larger batches
* frontend deduplicates by manga id
* pagination appears batch-based rather than page-based

Relevant frontend fetch logic found:

fetch(`${base}/content/latest-chapter-updates?batch=${batch}`)

4. Important issue encountered:

Attempting to reuse previous parsing logic or guessing endpoints resulted in:

* HTTP 500
* parsing failures
[...]

PR #18067, opened 2026-07-29
New source: Procomic (replaces MangaPro/ProChan)

bakali1:
Complete rewrite of the former MangaPro/ProChan extension as a new KeiSource (libVersion = "1.6"). The site and its API changed significantly — the old codebase was no longer salvageable, so this was built from scratch against the current API.

Changes
- Added src/ar/procomic/ with full implementation:
- REST API integration (/api/public/content, /api/public/series/search, /api/public/{type}/{id})
- Search with filters (type, sort, year, status)
- Popular / Latest updates
- Manga details and chapter listing
- Page list via Next.js RSC extraction + AVIF-to-JPEG transcoding via ImageDecoder
- Mirror support (procomic.pro / procomic.net) via baseUrl { mirrors(...) }
- Cookie interceptor for language/safety preferences
- Deeplink handling for /series/... URLs

In relations:

open #14571 ( Source request (AR): ProChan )
Closes #16185   ( remove mangapro / prochan)

Checklist:

- [X] Set the `contentWarning` configuration in `build.gradle.kts` appropriately
- [X] This PR is AI-assisted, I have reviewed the changes manually and confirmed they are not slop


src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,238 @@
...
+import tachiyomi.decoder.ImageDecoder
+import java.io.ByteArrayInputStream
+import java.io.ByteArrayOutputStream
+import java.util.Locale
+import kotlin.time.Instant
+
+@Source
+abstract class Procomic : KeiSource() {
+
+    override val supportsLatest = true
```
AwkwardPeak7: default in keiSource

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,238 @@
...
+                baseUrl.removePrefix("https://"),
+                listOf("safe_browsing" to "off", "language" to "ar"),
+            ),
+        )
+        addNetworkInterceptor(
+            Interceptor { chain ->
+                val response = chain.proceed(chain.request())
+                val contentType = response.header("Content-Type") ?: return@Interceptor response
+                if (contentType != "image/avif") return@Interceptor response
+                avifToJpeg(response)
```
AwkwardPeak7: why re-incode avif to jpg? no need to do it, app will handle avif just fine

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,238 @@
...
+            },
+        )
+    }
+
+    override fun Headers.Builder.configureHeaders() = apply {
+    }
```
AwkwardPeak7: remove empty override

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,238 @@
...
+            .addQueryParameter("page", page.toString())
+            .addQueryParameter("sort", effectiveSort)
+            .apply {
+                if (!isCatalog) addQueryParameter("search", query)
+                typeFilter?.selected?.also { addQueryParameter("type", it) }
+                yearFilter?.selected?.also { addQueryParameter("year", it) }
+                statusFilter?.selected?.also { addQueryParameter("status", it) }
+            }
+            .build()
+        val data = client.get(url, cacheControl = CacheControl.FORCE_NETWORK).parseAs<SearchResponse>()
```
AwkwardPeak7: why force network?

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,238 @@
...
+        val off = if (segs.firstOrNull() in listOf("ar", "en")) 1 else 0
+        if (segs.size < off + 4 || segs.getOrNull(off) != "series") return null
+        val type = segs[off + 1]
+        if (type !in SUPPORTED_TYPES) return null
+        val id = segs[off + 2]
+        val slug = segs[off + 3]
+        return SManga.create().also {
+            it.url = "/series/$type/$id/$slug"
+        }
+    }
```
AwkwardPeak7, line 120: check domain too

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,238 @@
...
+        val path = manga.url.removePrefix("/series/").split("/")
+        val apiManga = fetchApiManga(path[0], path[1])
+        val smanga = if (fetchDetails) apiManga.toSManga() else manga
+        val chapterList = if (fetchChapters) {
+            (apiManga.chapters ?: emptyList())
+                .filter { it.language == "AR" }
+                .map { it.toSChapter(path[0], path[1], apiManga.slug) }
+        } else {
+            chapters
+        }
```
AwkwardPeak7: you have all the info already, return both unconditionally

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,238 @@
...
+    private suspend fun fetchApiManga(type: String, id: String): ApiManga {
+        val url = "$baseUrl/api/public/$type/$id"
+        return client.get(url).parseAs<ApiManga>()
+    }
+
+    override suspend fun getPageList(chapter: SChapter): List<Page> {
+        val response = client.get(getChapterUrl(chapter), headersBuilder().set("rsc", "1").build())
+        return response.extractNextJs<ChapterImages>()?.appImages?.mapIndexed { i, img ->
+            Page(i, imageUrl = img.mobile ?: img.desktop ?: "")
+        } ?: emptyList()
```
AwkwardPeak7: they no longer have those split/obfuscated pages?

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    private fun ApiManga.toSManga(): SManga = SManga.create().apply {
+        url = "/series/$type/$id/$slug"
```
AwkwardPeak7: consider saving only unique id in url and other info in memo

ps if id is not globally unique, then you can save other thing in url

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,238 @@
...
+    private fun SearchItem.toSManga(): SManga = SManga.create().apply {
+        url = "/series/$type/$id/$slug"
+        title = this@toSManga.title
+        thumbnail_url = (coverImageApp?.desktop ?: coverImage)?.let {
+            if (it.startsWith("/")) "$baseUrl$it" else it
+        }
+    }
+
+    private fun ApiChapter.toSChapter(type: String, seriesId: String, slug: String): SChapter = SChapter.create().apply {
+        url = "/series/$type/$seriesId/$slug/$id/$chapterNumber"
```
AwkwardPeak7: same here, consider saving only unique id in url and other info in memo

AwkwardPeak7 (changes requested):

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
bakali1: it did have some caching problems it turn out that it just the cache in my phone after reinstalling the extension multiple times in a row 

so after remove it works fine , so i will remove it from the code

thank u for your review

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
bakali1: I had originally added this to resolve a "Failed to split downloaded image" error that appeared in logcat during downloads of chapters.

However, upon retesting without it, everything works fine. This turned out to be a non-essential Mihon feature that splits image downloads for readability

it doesn't affect the download itself or how images are displayed afterward.

so i will delete the avif to jpg function and it dependencies 

thank you for the review

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
bakali1: Thanks for the review!

I'll add the domain check. Quick question though — I was planning to use this:
kotlin

if (url.host != baseUrl.toHttpUrl().host) {
    return null
}

But Prochan has mirrors, so I'm not sure how to handle those. Is it worth adding a list of allowed mirror domains, or should I just stick with the main one?

Honestly, it's probably not a big deal since the main website is what users actually use — the mirrors are just redirects anyway. But I added it as a mirror because the API requests work through it too.

What do you think? Should I keep it simple with just the main domain, or build out a mirror list?

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
bakali1: Yes, the site no longer scrambles page images

they now serve clean URLs directly in the RSC payload under appImages[].mobile/desktop. 

I tested chapters across all three types (manga/manhwa/manhua) and none use the old puzzle/encryption system. The new getPageList matches their current API.

bakali1:
I used the extention for some time reading some Mangas I discovered some bugs I will try to fix
changes to be made :
- chapters missing images 
- the manga url should lead to the manga when clicking open in web view


dear reviewer am sorry for wasting your time dealing with this source 

I will try my best to fix the issue

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
AwkwardPeak7: it's fine, or if the mirror isn't really used, you can remove the mirror from extension entirely

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,225 @@
...
+        }
+        val segs = url.pathSegments
+        val off = if (segs.firstOrNull() in listOf("ar", "en")) 1 else 0
+        if (segs.size < off + 4 || segs.getOrNull(off) != "series") return null
+        val type = segs[off + 1]
+        if (type !in SUPPORTED_TYPES) return null
+        val id = segs[off + 2]
+        val slug = segs[off + 3]
+        return SManga.create().also {
+            it.url = "/$id"
```
AwkwardPeak7: no need to prefix with "/"

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,225 @@
...
+
+    private suspend fun fetchApiManga(type: String, id: String): ApiManga {
+        val url = "$baseUrl/api/public/$type/$id"
+        return client.get(url).parseAs<ApiManga>()
+    }
+
+    override suspend fun getPageList(chapter: SChapter): List<Page> {
+        val response = client.get(getChapterUrl(chapter), headersBuilder().set("rsc", "1").build())
+        return response.extractNextJs<ChapterImages>()?.appImages?.mapIndexed { i, img ->
+            Page(i, imageUrl = img.mobile ?: img.desktop ?: "")
```
AwkwardPeak7:
```suggestion
            Page(i, imageUrl = img.mobile ?: img.desktop!!)
```
shouldn't be empty

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,225 @@
...
+            put("slug", slug)
+        }
+        title = this@toSManga.title
+        thumbnail_url = (coverImageApp?.desktop ?: coverImage)?.let {
+            if (it.startsWith("/")) "$baseUrl$it" else it
+        }
+    }
+    override fun getChapterUrl(chapter: SChapter): String {
+        val m = chapter.memo
+        return "$baseUrl/series/${m["type"]!!.jsonPrimitive.content}/${m["seriesId"]!!.jsonPrimitive.content}/${m["slug"]!!.jsonPrimitive.content}/${chapter.url.removePrefix("/")}/${m["chapterNumber"]!!.jsonPrimitive.content}"
```
AwkwardPeak7:
```suggestion
        return "$baseUrl/series/${m["type"]!!.string}/${m["seriesId"]!!.string}/${m["slug"]!!.string}/${chapter.url.removePrefix("/")}/${m["chapterNumber"]!!.string}"
```
use JsonElement helper utils

bakali1: thank u for the review , i didn't know that exist i will use it instead of the current method

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
bakali1: thanks, i will apply it the change

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
bakali1: I’ll remove the prefix as well, and I won’t change the manga URL structure (`/series/$type/$id/$slug`) since that’s how it’s set on the website. That way, when you open it in the WebView, you can see your manga.

AwkwardPeak7: override `getMangaUrl` and `getChapterUrl` for webview urls

bakali1:
Correction to my earlier reply where I said the site no longer uses scrambled pages.

That was incorrect. Some chapters are still served as scrambled pages, which is what caused the "chapters missing images" bug I reported. This has been fixed in the latest commits.

I've verified the fix on a real device, and the affected chapters now load every page correctly.

I also discovered that some series are hidden behind the site's "Safe Browsing" setting. Guests can't disable this setting, so `getPageList` now throws an error for those series instead of returning an empty page list, making the issue much easier to identify.

That said, the site seems to change every couple of weeks. I suspect they're preparing or promoting their official app and are intentionally making the website harder to scrape. At this point, I'm considering marking this source as a draft because it may not be worth constantly chasing their frequent changes and debugging their ever-changing anti-scraping measures.

and thank you [AwkwardPeak7](https://github.com/AwkwardPeak7) so much for giving me advises and being friendly and patient with me 
i have learned a lot of thing building this extension after it i will try doing some more of the issues

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,414 @@
...
+        addInterceptor(::scrambledImageInterceptor)
+        addNetworkInterceptor(
+            CookieInterceptor(
+                baseUrl.removePrefix("https://"),
+                listOf("safe_browsing" to "off", "language" to "ar"),
+            ),
+        )
+    }
+
+    val rscHeaders by lazy { headersBuilder().add("rsc", "1").build() }
```
AwkwardPeak7, line 58:
```suggestion
    val rscHeaders get() = headersBuilder().add("rsc", "1").build()
```

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,414 @@
...
+                .build()
+        } finally {
+            pieceBitmaps.forEach { it.recycle() }
+            resultBitmap.recycle()
+        }
+    }
+
+    private suspend fun signImage(cdnUrl: String, referer: String): String? {
+        val payload = buildJsonObject { put("url", cdnUrl) }.toJsonString()
+            .toRequestBody("application/json".toMediaType())
```
AwkwardPeak7, line 280:
```suggestion
        val payload = buildJsonObject { put("url", cdnUrl) }.toJsonRequestBody()
```

src/ar/procomic/src/eu/kanade/tachiyomi/extension/ar/procomic/Procomic.kt
```diff
@@ -0,0 +1,414 @@
...
+
+        val rscheaders = rscHeaders.newBuilder()
+            .set("Referer", referer)
+            .set("Sec-Fetch-Site", "same-origin")
+            .build()
+        val response = client.post("$baseUrl/api/cdn-image/sign", rscheaders, payload)
+
+        if (!response.isSuccessful) {
+            response.close()
```
AwkwardPeak7, line 289: `client.post` ensures http success

AwkwardPeak7:
> promoting their official app

perhaps you can sniff the api of the app and use it ;)

PR #18140, opened 2026-08-01
Bilimanga: fix no search results and read error

hualiong:
Checklist:

### DIFF
--- a/src/zh/bilimanga/src/eu/kanade/tachiyomi/extension/zh/bilimanga/BiliManga.kt
+++ b/src/zh/bilimanga/src/eu/kanade/tachiyomi/extension/zh/bilimanga/BiliManga.kt
@@ -17,8 +17,12 @@ import keiyoushi.utils.getPreferencesLazy
 import keiyoushi.utils.tryParse
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.async
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.supervisorScope
 import kotlinx.serialization.json.JsonElement
+import okhttp3.CacheControl
+import okhttp3.Cookie
 import okhttp3.Headers
 import okhttp3.HttpUrl
 import okhttp3.HttpUrl.Companion.toHttpUrl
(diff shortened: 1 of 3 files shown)


hualiong:
I encountered a very strange issue while developing this extension: once the server response includes a `Set-Cookie` header, neither modifying the `Cookie` field in the request header via `OkHttpClient.Builder.configureClient()` nor using `addInterceptor` to modify it takes effect. The only way to restore the original state is to use the "Clear Cookies" button in the advanced settings. Alternatively, I found that using `addNetworkInterceptor` instead of `addInterceptor` to set the cookie does work as a workaround.

Is this a bug? I recall this has been happening for quite a while now.

AwkwardPeak7:
> once the server response includes a Set-Cookie header, neither modifying the Cookie field in the request header via OkHttpClient.Builder.configureClient() nor using addInterceptor to modify it takes effect

I don't understand, do you want to modify the server's set-cookie header? or set your own cookie? In both cases, the interceptor needs to be a network interceptor

src/zh/bilimanga/src/eu/kanade/tachiyomi/extension/zh/bilimanga/BiliManga.kt
```diff
@@ -100,12 +104,33 @@ abstract class BiliManga :
         return desc.toString()
     }
 
+    private suspend fun ensureSearchTicket() {
+        val names = client.cookieJar.loadForRequest(baseUrl.toHttpUrl()).map { it.name }
+        if (!names.contains("jieqiSearchCss") || !names.contains("jieqiSearchJs")) {
+            coroutineScope {
+                launch { client.get("$baseUrl/search.html?search_guard=css").close() }
```
AwkwardPeak7, line 111: why launch an unused request?

src/zh/bilimanga/src/eu/kanade/tachiyomi/extension/zh/bilimanga/BiliManga.kt
```diff
@@ -100,12 +104,33 @@ abstract class BiliManga :
...
+                launch {
+                    COOKIE_REGEX.find(client.get("$baseUrl/search.html?search_guard=js").body.string())
+                        ?.groupValues?.get(1)
+                        ?.let { cookie ->
+                            val url = baseUrl.toHttpUrl()
+                            Cookie.parse(url, cookie)?.let { client.cookieJar.saveFromResponse(url, listOf(it)) }
+                        }
```
AwkwardPeak7, line 118: cookie is in the body? not in header?

src/zh/bilimanga/src/eu/kanade/tachiyomi/extension/zh/bilimanga/BiliManga.kt
hualiong: The response to the request `$baseUrl/search.html?search_guard=css` will carry a `Set-Cookie` header, which will automatically set the cookie. No additional handling of the response is required.

The cookie for `$baseUrl/search.html?search_guard=js` is in the response body rather than in the headers.

src/zh/bilimanga/src/eu/kanade/tachiyomi/extension/zh/bilimanga/BiliManga.kt
hualiong: Yes

hualiong:
> I don't understand, do you want to modify the server's set-cookie header? or set your own cookie? In both cases, the interceptor needs to be a network interceptor

OK, get it

AwkwardPeak7 (approved):

src/zh/bilimanga/src/eu/kanade/tachiyomi/extension/zh/bilimanga/BiliManga.kt
AwkwardPeak7: weird source

PR #18184, opened 2026-08-04
utils: add new date utils

manti-X:


core/src/main/kotlin/keiyoushi/utils/Date.kt
```diff
@@ -2,9 +2,103 @@ package keiyoushi.utils
...
 import java.text.ParsePosition
 import java.text.SimpleDateFormat
+import java.time.LocalDate
+import java.time.LocalDateTime
+import java.time.ZoneId
+import java.time.ZonedDateTime
+import java.time.format.DateTimeFormatter
+import kotlin.time.Instant
 
 fun SimpleDateFormat.tryParse(date: String?): Long {
```
AwkwardPeak7, line 15: deprecate this

core/src/main/kotlin/keiyoushi/utils/Date.kt
```diff
@@ -2,9 +2,103 @@ package keiyoushi.utils
...
+ *
+ * Any offset or zone id present in [date] is ignored, since only date fields are extracted.
+ * Returns 0 if [date] is null or cannot be parsed.
+ *
+ * @see tryParseDateTime
+ * @see tryParseZonedDateTime
+ */
+fun DateTimeFormatter.tryParseDate(
+    date: String?,
+    zone: ZoneId = ZoneId.systemDefault(),
```
AwkwardPeak7, line 34: should default zone be utc?

manti-X: It matches SDF if no timezone is set

AwkwardPeak7 (approved):

PR #18351, opened 2026-08-12
Upload Extensions as Github Release

AwkwardPeak7:


Starmania (commented):
Looks good: Not much to say, good idea to use sha256 over rehashing remote

PR #18360, opened 2026-08-13
CI: skip already uploaded assets and retry on primary ratelimits

AwkwardPeak7:


.github/scripts/publish-repo.py
Starmania:
```patch
-         f.write(gzip.compress(index.SerializeToString(deterministic=True)))
+         f.write(gzip.compress(index.SerializeToString(deterministic=True), mtime=0))
```
This is what make our previous .pb changed while nothing apparently did. gzip add a timestamp...

Github suggestion is a little bit broken currently, so not using that.

Starmania (commented):
Good enough

Starmania (commented):
Good, added a side note

PR #18506, opened 2026-08-19
HoneyManga: fix getting chapters

Exinaus:
Checklist:

Previous page size of 10000 now returns 400 error. 
Updated code to new limit of 1000 per request.


src/uk/honeymanga/src/eu/kanade/tachiyomi/extension/uk/honeymanga/HoneyManga.kt
```diff
@@ -180,16 +180,25 @@ abstract class HoneyManga :
...
+                        mangaId = mangaId,
+                        page = page,
+                        pageSize = 1000,
+                        sortOrder = "DESC",
+                    ).toJsonRequestBody()
+
+                    val data = client.post(chaptersUrl, body).parseAs<ChapterResponse>()
+                    fetchedChapters.addAll(data.data.mapNotNull { it.toSChapter(hideLocked) })
+                    page++
+                } while (data.hasNextPage)
```
AwkwardPeak7, line 199: can/should they be fetched in parallel?

Exinaus: I don't think it could be done in parallel. You need to get first response (from `page = 1`) to check `data.hasNextPage` from it, and fetch next batch if it's true.

AwkwardPeak7 (approved):

PR #18550, opened 2026-08-21
Manga18fx: migrate to 1.6, remove Madara, sort genres, popular page 2+, search with `'`

vetleledaal:
Probably still have some weird abstractions, not that familiar with 1.6 yet.

Lacking areas (not exhaustive):
- Not using slug for manga URL
- Not using `DateTimeFormatter`
- Conventions
- English: the app will stop pagination if a page only contains non-English entries

Marking as a draft as I still need to test the migration flow.

Used AI for `fetchRelatedMangaList` / `memo`. 

Checklist:


src/all/manga18fx/src/eu/kanade/tachiyomi/extension/all/manga18fx/Manga18fx.kt
```diff
@@ -0,0 +1,368 @@
...
+import org.jsoup.nodes.Document
+import org.jsoup.nodes.Element
+import java.text.SimpleDateFormat
+import java.util.Locale
+
+// Similar to Madara, but not really
+@Source
+abstract class Manga18fx : KeiSource() {
+    // DateTimeFormatter does not support 2-digit years
+    private val chapterDateFormat = SimpleDateFormat("dd MMM yy", Locale.ENGLISH)
```
manti-X: DateTimeFormatter does support two digit years tho (only between 2000 to 2099): https://developer.android.com/reference/kotlin/java/time/format/DateTimeFormatter#:~:text=Year%3A%20The%20count,to%202099%20inclusive.

src/all/manga18fx/src/eu/kanade/tachiyomi/extension/all/manga18fx/Manga18fx.kt
```diff
@@ -0,0 +1,368 @@
...
+            val url = baseUrl.toHttpUrl().newBuilder()
+                .addPathSegment("manga-genre")
+                .addPathSegment(selectedGenre)
+                .apply { if (page != 1) addPathSegment(page.toString()) }
+                .apply { if (selectedGenreOrderBy != null) addQueryParameter("orderby", selectedGenreOrderBy) }
+                .build()
+            return mangaParse(client.get(url).asJsoup())
+        }
+
+        error("Select a genre to search")
```
AwkwardPeak7, line 75: fallback to latest?

src/all/manga18fx/src/eu/kanade/tachiyomi/extension/all/manga18fx/Manga18fx.kt
```diff
@@ -0,0 +1,368 @@
...
+        val relatedMangas = manga.memo["relatedMangas"]?.parseAs<List<RelatedManga>>()
+            ?: run {
+                val mangaUpdate = fetchMangaUpdate(
+                    manga = manga,
+                    chapters = emptyList(),
+                    fetchDetails = true,
+                    fetchChapters = false,
+                )
+                mangaUpdate.manga.memo["relatedMangas"]?.parseAs<List<RelatedManga>>()
+            }
```
AwkwardPeak7: no need for this fallback, app will call related after calling for manga update

src/all/manga18fx/src/eu/kanade/tachiyomi/extension/all/manga18fx/Manga18fx.kt
```diff
@@ -0,0 +1,368 @@
...
+    private fun SManga.toRelatedItem() = RelatedManga(url, title, thumbnail_url)
+
+    // Filters
+    override fun getFilterList(data: JsonElement?) = FilterList(
+        buildList {
+            add(Filter.Header("Filters do not apply to text search"))
+            add(
+                GenreFilter(
+                    listOf(
+                        Genre("<select>", null),
```
AwkwardPeak7, line 256: perhaps fetch genres?

vetleledaal: Not all genres are present on the regular site, got these from the sitemap. Should I just parse that?

AwkwardPeak7: you can but it's fine as is too

src/all/manga18fx/src/eu/kanade/tachiyomi/extension/all/manga18fx/Manga18fx.kt
vetleledaal: But initially for people on older versions `memo` will be `null`. Won't it fail in that case?

AwkwardPeak7: it's fine, no need to handle that case

src/all/manga18fx/src/eu/kanade/tachiyomi/extension/all/manga18fx/Manga18fx.kt
vetleledaal: Is that desirable? The user might get the wrong impression if they just change the sort order and see that it has no effect.

AwkwardPeak7: true

src/all/manga18fx/src/eu/kanade/tachiyomi/extension/all/manga18fx/Manga18fx.kt
vetleledaal: Not sure what I did to break it previously, but it works now :+1:

src/all/manga18fx/src/eu/kanade/tachiyomi/extension/all/manga18fx/Manga18fx.kt
vetleledaal: Komikku seems to handle it gracefully :+1:

AwkwardPeak7 (approved):

PR #18551, opened 2026-08-22
MangaFire: add a solver for shape-selecting captcha

tigeryu8900:
MangaFire's shape-selecting captcha still shows up for certain IP addresses despite them switching over to Cloudflare challenge.

This change adds a solver for MangaFire's captcha via a WebView using OpenCV.js.

Since the success rate of the solver is around 10% to 20%, the solver is run in batches of 10, up to a total of 100 attempts. The captcha APIs apparently don't need rate limiting.

There are two main reasons I did not implement the solver in Kotlin:
1. The OpenCV dependency (org.opencv:opencv:5.0.0.1) is bloated (145.9 MB aar file) compared to v5.0 OpenCV.js (4321 KB over the network).
2. The .so files would be under the extension's app-specific directory instead of the reader's app-specific directory, which means we would need to store the .so files as assets and write them to temp files before manually loading them.

I used ChatGPT to help debug and improve the accuracy of the OpenCV magic, hence the AI-assisted checkbox.

<details>
  <summary>Example captcha</summary>

image

</details>

<details>
  <summary>Video of solver in action in Komikku</summary>

https://github.com/user-attachments/assets/e5a97ed7-2449-484f-87d7-c51decc22a62

</details>

Checklist:

### DIFF
--- a/src/all/mangafire/src/eu/kanade/tachiyomi/extension/all/mangafire/ChallengeSolverInterceptor.kt
+++ b/src/all/mangafire/src/eu/kanade/tachiyomi/extension/all/mangafire/ChallengeSolverInterceptor.kt
@@ -32,7 +32,6 @@ class ChallengeSolverInterceptor(
         val error: String?,
     )
 
-    @SuppressLint("SetJavaScriptEnabled")
     override fun intercept(chain: Interceptor.Chain): Response {
         val call = chain.call()
         val request = chain.request()
--- a/src/all/mangafire/src/eu/kanade/tachiyomi/extension/all/mangafire/ChallengeSolverInterceptor.kt
+++ b/src/all/mangafire/src/eu/kanade/tachiyomi/extension/all/mangafire/ChallengeSolverInterceptor.kt
@@ -1,28 +1,27 @@
 package eu.kanade.tachiyomi.extension.all.mangafire
 
-import eu.kanade.tachiyomi.network.NetworkHelper
 import keiyoushi.utils.parseAs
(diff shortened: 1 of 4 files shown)


AwkwardPeak7 (commented):
I don't think this solver is a good idea, it can break easily and does not have too high of a success rate. perhpas it will be better to detect and throw on captcha so a helpful message is displayed to the user so they can solve it in webview

tigeryu8900:
The success rate per solve attempt may be low, but each attempt is quick and can be verified in parallel, so the final success rate is near 100%.

I can put the solve part behind a setting so that it can be turned off if it breaks in the future.

tigeryu8900:
I have updated the logic so that it only solves captchas if the user explicitly turns on the setting. Otherwise, it only detects if there's a captcha and throws a message when a captcha is detected.

When all solve attempts fail, it will throw a message to prompt the user to solve the captcha manually.

src/all/mangafire/src/eu/kanade/tachiyomi/extension/all/mangafire/ChallengeSolverInterceptor.kt
AwkwardPeak7, line 1: use `runWebViewBlocking` helper

tigeryu8900:
I also added
1. a check to see if the captcha is solved in another call so the solver isn't run back to back
2. locks so that other calls are paused when the solver is running

AwkwardPeak7 (approved):

PR #18559, opened 2026-08-22
Mangas-Origines.fr: fix page list

### LINKED ISSUE #18555: Mangas-Origines: `UnsupportedOperationException: null` error
### Source information

Mangas-Origines.fr 1.6.57

### Source language

French

### Steps to reproduce

Open a chapter

### Expected behavior

Have the chapter 

### Actual behavior

UnsupportedOperationException: null

### Mihon/Mihon fork version

0.20.4

### Android version

One ui 8.5

### Other details

_No response_

### Acknowledgements

### 

---

AwkwardPeak7:
closes https://github.com/keiyoushi/extensions-source/issues/18555

Checklist:

### DIFF
--- a/lib-multisrc/origines/build.gradle.kts
+++ b/lib-multisrc/origines/build.gradle.kts
@@ -3,6 +3,6 @@ plugins {
 }
 
 keiyoushi {
-    baseVersionCode = 0
+    baseVersionCode = 1
     libVersion = "1.6"
 }
--- a/lib-multisrc/origines/src/eu/kanade/tachiyomi/multisrc/origines/Origines.kt
+++ b/lib-multisrc/origines/src/eu/kanade/tachiyomi/multisrc/origines/Origines.kt
@@ -268,8 +268,12 @@ abstract class Origines : KeiSource() {
     override suspend fun getPageList(chapter: SChapter): List<Page> {
         val document = client.get("$baseUrl/$mangaPath/${chapter.url.toChapterSlug()}/").asJsoup()
 
-        return document.select("div.reading-content img.wp-manga-chapter-img").mapIndexed { index, image ->
-            Page(index, imageUrl = image.attr("src").trim())
+        return document.select("div.reading-content img.wp-manga-chapter-img").mapIndexed { index, img ->
+            val image = when {
+                img.hasAttr("data-src") -> img.absUrl("data-src").trim()
+                else -> img.absUrl("src").trim()
+            }
+            Page(index, imageUrl = image)
         }
     }
 


Trytotry2:
> **Disclosure:** this investigation and the patch below were produced with an AI assistant
> (Claude). I would not have found this on my own, so please review it on its merits rather than
> taking it on trust. The one part I can personally vouch for is the outcome: I installed the
> built extension on my phone and chapter 166 reads and downloads correctly.

Independent investigation of the same bug, done before finding this PR. Same root cause, same
`baseVersionCode` call — so this is only offered as extra evidence plus some optional hardening.
Take whatever is useful, ignore the rest.

## What the page actually serves

Measured on `https://mangas-origines.fr/oeuvre/the-knight-king-who-returned-with-a-god/chapitre-166/`
on 2026-08-22, rendered in a real browser (Cloudflare returns a managed challenge to plain clients):

- The `src` attribute is **absent from the tag entirely** — not present-but-empty. The full tag is
  ``
- The URL in `data-src` is **padded with a leading space**. `absUrl()` handles it, because
  `java.net.URL` trims leading/trailing chars `<= ' '` — so the `.trim()` in this PR is harmless
  belt-and-braces, not load-bearing. Worth knowing it is safe either way.
- The page loads `wp-content/themes/madara/js/lazysizes/lazysizes.min.js?ver=5.3.2`. **This is not
  anti-scraping** — it is Madara's lazy-load reading option being switched on, and `data-src` is
  lazysizes' hardcoded contract. That makes the attribute stable, and it likely means other Madara
  sources will show the same symptom whenever that option gets toggled.
- Chapter images answer `200` to a plain client with no `Referer` and no challenge. No login wall on
  the tested chapter (the only "premium" hit in the HTML is the Yoast SEO plugin banner).
- Selectors were never the problem: `div.reading-content img.wp-manga-chapter-img` still matches all
  24 pages.

## Suggestions

### 1. `mapIndexed` → `mapIndexedNotNull` (the one that actually matters)

As written, if a page has neither `data-src` nor `src`, the `else` branch produces `""` and
`Page(index, imageUrl = "")` reaches OkHttp — which is exactly the original
`expected URL scheme http or https but no scheme was found for` crash. Dropping the page instead
turns that into a clean "no pages found", which names the real problem.

### 2. If you add more attribute names, keep `srcset` out of a naive list

`srcset` packs `<url> <descriptor>` pairs. Resolved whole it yields a *plausible but wrong* URL that
still passes any `startsWith("http")` + extension check — a silently wrong image instead of a loud
failure, which is worse than the bug being fixed here.

### 3. A blind attribute pass needs an extension guard

If you sweep every attribute looking for something URL-shaped, note that `absUrl("id")` on
`id="image-0"` resolves to `https://mangas-origines.fr/oeuvre/…/chapitre-166/image-0` — an absolute
`http` URL. The image-extension check is what rejects it, so it is load-bearing, not decoration.

### 4. Optional: a raw-markup last resort

Matching page URLs in the served markup survives the tags being renamed, the container changing, or
the list moving into a `<script>` as JSON. Madara's own data directory turns out to be a perfect
discriminator: on the tested chapter, `wp-content/uploads/WP-manga/data/` matches **24 URLs, all of
them chapter pages, zero false positives**. JSON escapes its slashes, hence the unescaping.

## Verification

Transposed the logic faithfully and ran it with Jsoup 1.18.1 against the captured HTML, mutating the
markup to simulate future breakage:

| Scenario | Result |
|---|---|
| Site as served today (`data-src`) | 24/24, order preserved |
| Back to a populated `src` | 24/24 |
| Attribute renamed to an unknown `data-ori-page` | 24/24 (blind pass) |
| `wp-manga-chapter-img` class renamed | 24/24 (raw markup) |
| `reading-content` container renamed | 24/24 (raw markup) |
| Tags emptied, URLs moved into a `<script>` as escaped JSON | 24/24 |
| URLs absent from the HTML (assembled in JS) | 0 pages — clean "no pages found", no crash |

Also built as a debug APK and confirmed on device: chapter 166 reads and downloads.

Known limits, for honesty: none of this survives URLs assembled from fragments in JS, encoded URLs,
or per-session signed tokens. Only a WebView-rendered approach would, at a cost that is not worth
paying today.

Note the theme is shared with Hentai-Origines, so the `baseVersionCode` bump in this PR correctly
covers both sources.

## Code

```kotlin
override suspend fun getPageList(chapter: SChapter): List<Page> {
    val document = client.get("$baseUrl/$mangaPath/${chapter.url.toChapterSlug()}/").asJsoup()

    val images = document.select("div.reading-content img.wp-manga-chapter-img")
    val fromTags = images.mapNotNull { it.imageUrl() }

    // Pages missing from the tags means the template moved on: fall back to the raw markup,
    // but only if it actually accounts for more of the chapter than the tags did.
    val urls = when {
        fromTags.isNotEmpty() && fromTags.size == images.size -> fromTags
        else -> document.pageUrlsFromMarkup().takeIf { it.size > fromTags.size } ?: fromTags
    }

    return urls.mapIndexed { index, url -> Page(index, imageUrl = url) }
}

/**
 * The reader lazy-loads its pages through lazysizes: the tags carry no `src` at all and the
 * URL sits in `data-src`, padded with a leading space. The attributes lazy-loading is known to
 * use are tried first, then every remaining one, so the theme renaming the holder does not
 * break reading again. A page without any usable URL is left to the fallback below rather than
 * reaching the client as a blank URL.
 */
private fun Element.imageUrl(): String? = IMAGE_ATTRS.firstNotNullOfOrNull { attr ->
    absUrl(attr).trim().takeIf(String::isNotBlank)
} ?: attributes().asSequence()
    .map { absUrl(it.key).trim() }
    .firstOrNull { it.startsWith("http") && IMAGE_EXT.containsMatchIn(it) }

/**
 * Last resort, reading the URLs out of the served markup itself: it holds whether the reader
 * keeps them in tags at all, so it also covers the list being handed to a script as JSON.
 * JSON escapes its slashes, hence the unescaping before matching.
 */
private fun Document.pageUrlsFromMarkup(): List<String> = PAGE_URL_REGEX.findAll(html().replace("""\/""", "/"))
    .map { it.value }
    .distinct()
    .toList()

companion object {
    // … existing DATE_REGEX / TIME_ZONE …

    /**
     * Attributes a lazy-loading reader is known to hold the page URL in, likeliest first.
     * `srcset` is left out on purpose: it packs `<url> <descriptor>` pairs, so resolving it
     * whole yields a plausible but wrong URL, failing silently instead of loudly.
     */
    private val IMAGE_ATTRS = listOf(
        "data-src",
        "data-lazy-src",
        "data-original",
        "data-cfsrc",
        "data-url",
        "src",
    )

    /**
     * Guards the blind pass: an attribute like `id` also resolves to an absolute URL.
     */
    private val IMAGE_EXT = Regex("""\.(jpe?g|png|webp|avif|gif)""", RegexOption.IGNORE_CASE)

    /**
     * Chapter pages are the only images served out of Madara's own data directory, so the
     * path is what keeps the reader's decoration out of the last resort.
     */
    private val PAGE_URL_REGEX = Regex(
        """https?://[^\s"'<>]+/wp-content/uploads/WP-manga/data/[^\s"'<>]+\.(?:jpe?g|png|webp|avif|gif)""",
        RegexOption.IGNORE_CASE,
    )
}
```

godshiroi (commented):
Thank you! When the patch is comming ? The issue is still there

Trytotry2:
> Thank you! When the patch is comming ? The issue is still there

The update has been pushed yesterday. Update your sources and it should worked.

PR #18613, opened 2026-08-24
Add TruyenQQ VN

wdchocopie:
Adds a new extension for `https://truyenqq.com.vn`.

### Why a new extension instead of extending `vi/truyenqq`

Despite the shared branding, `truyenqq.com.vn` is a different site from the one the existing `TruyenQQ` extension targets (`truyenqqko.com`). None of the existing extension's endpoints or selectors work there:

| Existing extension expects | `truyenqq.com.vn` returns |
|---|---|
| `/truyen-yeu-thich` (popular) | HTTP 500 |
| `/truyen-moi-cap-nhat` (latest) | HTTP 500 |
| `/truyen-tranh/<slug>` (details) | HTTP 410 Gone |

Manga live at `/<slug>` and chapters at `/<slug>/<chapter>`, and the markup differs entirely (`.inner .item` / `.info h3 a` vs `.book_info .qtip`). Pointing the existing extension's custom base URL at this domain would just break it, so this is a separate source.

### Site notes

- Popular `/truyen-hot`, latest `/truyen-moi`, completed `/truyen-full`, genre `/the-loai/<slug>`, all paginated with `?page=N`. The homepage ignores `page`, so `/truyen-moi` backs latest.
- Details and the full chapter list come from the same page, so `fetchMangaUpdate` parses both from a single request.
- Chapter dates mix relative labels ("2 giờ trước", "hôm qua") with `dd/MM/yyyy`; both are handled, falling back to `0L`.
- Page images are `img.lazy[data-src]` pointing at a CDN that returns 403 without a `Referer` — covered by the header `KeiSource` already sets. The CDN host and image extension vary per series, so neither is hardcoded.
- The site has no sort or advanced-search parameters (no `<select>` anywhere, the search form has a single `s` param), so filters are limited to list and genre selection.
- The site fills the author field with the placeholder "Đang cập nhật" when unknown; that value is dropped rather than stored.

### Checklist

Tested on a physical device (Android 12, Mihon 0.20.4): popular, latest, search, details, chapter list, chapter dates and page images all load; `lintRelease` passes.

---

🤖 I am an AI agent. I was asked to research this extension repository, build a Mihon extension for `truyenqq.com.vn`, and open this pull request. I surveyed the site, wrote the extension, built it, and verified it end-to-end on a real device. This PR was opened by an AI agent; the AI-assisted checklist item above is intentionally left unchecked for the human maintainer of this fork to confirm after their own review.


src/vi/truyenqqvn/src/eu/kanade/tachiyomi/extension/vi/truyenqqvn/Filters.kt
```diff
@@ -0,0 +1,84 @@
...
+            "Truyện mới" to "/truyen-moi",
+            "Truyện full" to "/truyen-full",
+        ),
+    )
+
+class GenreFilter :
+    UriPartFilter(
+        "Thể loại",
+        arrayOf(
+            "Tất cả" to "",
```
AwkwardPeak7: consider fetching genres from site. see `supportsFilterFetching`

src/vi/truyenqqvn/src/eu/kanade/tachiyomi/extension/vi/truyenqqvn/TruyenQQVN.kt
```diff
@@ -0,0 +1,201 @@
...
+        Filter.Header("Bộ lọc bị bỏ qua khi tìm kiếm bằng tên"),
+        ListFilter(),
+        GenreFilter(),
+    )
+
+    companion object {
+        private const val POPULAR_PATH = "/truyen-hot"
+        private const val LATEST_PATH = "/truyen-moi"
+        private const val PLACEHOLDER_VALUE = "Đang cập nhật"
+
```
AwkwardPeak7, line 224: inline

src/vi/truyenqqvn/build.gradle.kts
```diff
@@ -0,0 +1,21 @@
...
+    contentWarning = ContentWarning.MIXED
+    libVersion = "1.6"
+
+    source {
+        baseUrl = "https://truyenqq.com.vn"
+        lang = "vi"
+    }
+
+    deeplink {
+        path("/..*")
```
AwkwardPeak7, line 20: consider a bit more specific path

src/vi/truyenqqvn/src/eu/kanade/tachiyomi/extension/vi/truyenqqvn/Filters.kt
wdchocopie: Done — switched to `supportsFilterFetching` with `fetchFilterData()` scraping `a[href*=/the-loai/]` off the homepage. Checked against the live page: it yields the same 59 genres the hardcoded list had, so nothing is lost by dropping the static array. The genre filter is now omitted entirely until the data lands, rather than showing a select with only "Tất cả" in it.

src/vi/truyenqqvn/src/eu/kanade/tachiyomi/extension/vi/truyenqqvn/TruyenQQVN.kt
wdchocopie: Done — inlined. It was only referenced once, so the constant was not earning its place.

src/vi/truyenqqvn/build.gradle.kts
wdchocopie: Added `/..*/chapter-..*`, and extended `getMangaByUrl` to accept two-segment chapter urls as well — it previously only matched `/<slug>`, so a chapter deeplink would have resolved to nothing.

The series pages are the awkward part: they sit at the root as `/<slug>` with no prefix to anchor on, so `/..*` is the only pattern that matches them, and it necessarily subsumes the chapter pattern too. As far as I can tell `pathPattern` has no way to exclude the listing paths (`/tim-kiem`, `/truyen-hot`, `/the-loai/...`), which is what makes the broad pattern uncomfortable.

So the realistic options are to keep `/..*` as-is, or drop it and register only the chapter pattern — that would lose browser deeplinks for series pages, though pasting a series url into search still resolves, since that path does not go through the intent filter. I have kept it for now; happy to drop it if you would rather the filter stay narrow.

src/vi/truyenqqvn/src/eu/kanade/tachiyomi/extension/vi/truyenqqvn/TruyenQQVN.kt
AwkwardPeak7: paths not inlined

AwkwardPeak7 (approved):

PR #18622, opened 2026-08-25
Comix: fix fast chapter deduplication

### LINKED ISSUE #18612: Comix: fast ch fetch breaks deduplicate ch feat
### Source information

Comix 1.6.35

### Source language

English 

### Steps to reproduce

1. Faster chapter list fetching = enabled
2. Deduplicate chapters = enabled
3. Update a series to grab new chapters
4. Newly uploaded chapters uploaded by comix-marked uploaders don't get deduplicated even if the chapter has already been marked as read.

### Expected behavior

Newly uploaded chapters deduplicate, prioritizing comix-marked uploads, but still respect read history and chapters read.

### Actual behavior

Newly uploaded chapters from comix-marked uploads get prioritized no matter the read history or chapters read. Will always show up in series' chapter list even it's a duplicate.

Image
Image
Image

### Mihon/Mihon fork version

TachiyomiSY 1.13.2

### Android version

Android 17

### Other details

Deduplicate chapter feature prioritizes Comix-marked uploaders such as Tapas and Webtoon.

Comix-marked chapters have started appear in my library as unread even though they're duplicates that have already been uploaded months prior. This issue started happening with the release of the faster chapter fetch feature.

### Acknowledgements

### 

---

CasperKristiansson:
Closes #18612

## Problem

With both **Faster chapter list fetching** and **Deduplicate Chapters** enabled, Comix stopped when it found the newest known chapter and merged the fetched result with Mihon's cached chapters by URL.

Different uploads of the same chapter number have different URLs, so both the newly selected upload and the cached upload could survive. On duplicate-heavy titles such as *My Bias Gets on the Last Train*, that produced interleaved WebToon, DivaScans, RinkoComics, and other uploads instead of one row per chapter number.

## Implementation

- Store the values used to rank an upload in its chapter memo: chapter ID, votes, Comix's official flag, and scanlation-group ID.
- Run one selection pass over newly fetched and cached chapters, applying URL uniqueness, the scanlator blacklist, and chapter-number deduplication consistently.
- Keep the original fast-fetch stopping behavior: once the newest known chapter is found, fetching stops and cached ranking data is used for the rest of the comparison. Cached values may become stale, which is the intended tradeoff for keeping this path fast.
- Force one complete refresh when deduplication mode or the blacklist changes, since previously hidden alternatives cannot be reconstructed from the selected chapter alone. This also migrates existing cached chapters to the new ranking memo.
- Apply the same mapping and selection flow to the native API and WebView paths.
- Clarify the fast-fetch and deduplication preference descriptions.
- Bump Comix from `1.6.35` to `1.6.36`.

## Validation

Local checks on commit `5f56061bdb`:

- `./gradlew :src:en:comix:spotlessCheck`
- `./gradlew :src:en:comix:lintRelease`
- `./gradlew :src:en:comix:assembleRelease`
- `git diff --check`

End-to-end testing used the release APK in Mihon `0.20.4` on a rooted Android 15 ARM64 emulator:

| Scenario | Persisted Mihon result |
| --- | --- |
| Initial refresh with fast fetch + deduplication | 92 rows / 92 distinct chapter numbers |
| Deduplication disabled | 689 uploads / 92 numbers; seven variants each for chapters 74 and 81 |
| Deduplication re-enabled | Returned to 92 / 92; chapters 74 and 81 selected WebToon |
| Recreated the reported stale cache with DivaScans chapter 81, then fast refreshed | Returned to one WebToon chapter 81 and remained 92 / 92 |
| Simulated the old chapter-memo schema | One full refresh repopulated ranking data on all 92 chapters |
| Blacklisted WebToon | RinkoComics replaced chapter 74 and DivaScans replaced chapter 81 |
| Cleared the blacklist | WebToon returned; chapter 74 remained read through both replacements |

As an additional cross-check, all 689 captured uploads were ranked independently using the same documented priority. The extension returned all 92 expected winners with zero mismatches.

Checklist:

🤖 Casper asked Codex to diagnose the Comix fast-fetch/deduplication regression, implement and extensively validate the fix, and update this pull request. Casper reviewed the implementation and test results and approved proceeding. This PR was opened and updated by an AI agent.


src/en/comix/src/eu/kanade/tachiyomi/extension/en/comix/Comix.kt
```diff
@@ -389,39 +389,72 @@ abstract class Comix :
         }
 
         val deduplicateChapters = preferences.deduplicateChapters()
+        val scanlatorBlacklist = preferences.scanlatorBlacklist()
+        val blacklistSignature = scanlatorBlacklist.sorted().joinToString(",")
+        val chapterSelectionSignature = "$CHAPTER_SELECTION_VERSION:$blacklistSignature"
```
AwkwardPeak7: what is CHAPTER_SELECTION_VERSION?

AwkwardPeak7 (commented):
I think this can be simplified

src/en/comix/src/eu/kanade/tachiyomi/extension/en/comix/Comix.kt
CasperKristiansson: It's just a cache version for the dedup selection. The idea was that if the ranking logic changes later, bumping it would force one full refresh instead of reusing the old cached selection. I added it here; it isn't a pattern used by other extensions. Looking at it again, the new memo key already invalidates old caches for this update, so I agree it's probably unnecessary. I can simplify this to track only the blacklist, since the dedup setting is already stored separately. Is that the simplification you had in mind, or did you mean the fetch flow more broadly?

AwkwardPeak7: I just think this might be overcomplicating it a bit. for example a simpler implementation can store views/etc in chapter memo so deduplication logic can have the same info available all the time.
ofcourse it may go stale in memo but it is worth the tradeoff for fast path

CasperKristiansson: Yeah, I agree — the previous version was doing too much. I reworked it around your suggestion: each chapter now keeps the inputs Comix actually uses for ranking (official flag, group ID, votes, and chapter ID) in its memo, and dedupe runs once over the newly fetched and cached chapters together. Fast mode can stop at the newest known chapter again, so the boundary-fetch and selection-version machinery is gone.

I kept a one-time full refresh when the dedupe mode or blacklist changes, since those settings can hide candidates that aren't recoverable from the cached winner. For normal refreshes, stale memo values are now the deliberate speed tradeoff you mentioned.

I retested it with My Bias Gets on the Last Train, including recreating a cached DivaScans chapter 81 and then doing a fast refresh; it returned to a single WebToon chapter 81. I also compared all 689 uploads against an independent ranking and got the expected winner for all 92 chapter numbers. Thanks, this ended up much cleaner.

AwkwardPeak7 (commented):
bump version again

CasperKristiansson:
@AwkwardPeak7 Done — the branch is up to date and Comix is now bumped to version 37. Everything is green now. Thanks!

AwkwardPeak7 (approved):

PR #15756, opened 2026-05-10
[RU]Com-X bypass __guard antibot

martinbirn:
The site replaced the old `Protected by Batman` antibot with a SHA-256 PoW challenge served as HTTP 404 from `/_c`. The previous detection (`bodyString.contains("Protected by Batman")`) no longer matches anything, so every manga details / chapter list / search request now fails with a generic `HTTP 404` and the source is effectively unusable.

### What the new antibot does

1. `GET <page>` → `302` to `/_c?t=<token>&u=<encoded-target>` + `Set-Cookie: __guard_token=...`
2. `GET /_c?...` → **HTTP 404** (intentional) with an HTML+JS body
3. JS computes SHA-256 of `<token>:<nonce>` until the hash starts with `00`, gathers a small fingerprint (touch / screen / WebGL renderer / timezone / DPR / CDP-detector flags) and `POST`s it form-encoded to `/_v`
4. `/_v` returns `200 OK` + `Set-Cookie: __guard_trust=...`
5. The original URL now returns the real page

### What this PR does

Replaces the dead `"Protected by Batman"` detection with detection by URL path (`response.request.url.encodedPath == "/_c"`) and adds `solveGuardChallenge` which performs the same flow in Kotlin: parse the token from the body, run the SHA-256 PoW (`MessageDigest`), `POST /_v` with a static fingerprint, then retry the original request. The `__guard_trust` cookie lands in the okhttp `cookieJar` automatically and subsequent requests skip the challenge.

The fingerprint values were verified against the live server with multiple `User-Agent`s (iPhone CriOS / desktop Chrome / Android Chrome) — the chosen Apple-GPU fingerprint passes regardless of UA, which keeps the patch UA-agnostic.

### Checklist

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [ ] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [ ] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [x] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [ ] Have removed `web_hi_res_512.png` when adding a new extension
- [x] This PR is AI-assisted, I have reviewed the changes manually and confirmed they are not slop

PR #18144, opened 2026-08-01
Com-X (RU) migrate to 1.6

Exinaus:
Checklist:

- migrate to 1.6
- add deeplink
- update filter to fetch data from site and exclude functionality
- webview based antiguard from batcave/mangabay
- related manga (komikku)

.

- [x] Updated `versionCode` value in `build.gradle.kts`
- [ ] Updated `baseVersionCode` in `build.gradle.kts` (if updated multisrc theme code)
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Set the `contentWarning` configuration in `build.gradle.kts` appropriately
- [x] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [ ] Have removed `web_hi_res_512.png` when adding a new extension
- [ ] This PR is AI-assisted, I have reviewed the changes manually and confirmed they are not slop

PR #18654, opened 2026-08-27
Com-X (RU): fix WebView antibot bypass in Tachimanga

martinbirn:
`runWebViewBlocking` installs a `WebViewClient` that Tachimanga intercepts as custom URL-scheme
tasks, causing guarded detail requests to time out. Use a client-free WebView to wait for the final
page DOM, preserve real 404 responses, and use `getMangaUrl(manga)` to avoid a double slash.

Related to #18144.

Checklist:

### DIFF
--- a/src/ru/comx/src/eu/kanade/tachiyomi/extension/ru/comx/DleGuardResolver.kt
+++ b/src/ru/comx/src/eu/kanade/tachiyomi/extension/ru/comx/DleGuardResolver.kt
@@ -1,25 +1,61 @@
 package eu.kanade.tachiyomi.extension.ru.comx
 
-import android.webkit.CookieManager
-import keiyoushi.utils.runWebViewBlocking
+import android.annotation.SuppressLint
+import android.os.Handler
+import android.os.Looper
+import android.webkit.WebView
+import keiyoushi.utils.applicationContext
+import keiyoushi.utils.parseAs
+import kotlinx.serialization.Serializable
 import okhttp3.Call
 import okhttp3.Interceptor
+import okhttp3.MediaType.Companion.toMediaType
+import okhttp3.ResponseBody.Companion.toResponseBody
 import java.io.IOException
-import kotlin.time.Duration.Companion.milliseconds
-import kotlin.time.Duration.Companion.seconds
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
 
 object DleGuardResolver {
 
     private const val TIMEOUT_SECONDS = 30L
     private const val POLL_INTERVAL_MS = 250L
-    private const val PARALLEL_TRUST_WINDOW_MS = 5_000L
-    private const val TRUST_COOKIE = "__guard_trust"
+    private const val FAILURE_RETRY_DELAY_MS = 5_000L
 
-    @Volatile
-    private var failedOnce = false
+    private val htmlMediaType = "text/html; charset=UTF-8".toMediaType()
 
-    @Volatile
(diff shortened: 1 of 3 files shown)


src/ru/comx/src/eu/kanade/tachiyomi/extension/ru/comx/DleGuardResolver.kt
```diff
@@ -1,81 +1,219 @@
...
+
+        val finalRequest = originalRequest.newBuilder()
+            .url(pageResponse.url)
+            .build()
+
+        responseBuilder
+            .request(finalRequest)
+            .code(pageResponse.code)
+            .message(pageResponse.message)
+            .removeHeader("Content-Encoding")
```
Exinaus, line 93:
```suggestion
            .headers(headers)
            .removeHeader("Content-Encoding")
```

I'm asking just in case: did you test it in android? Because your [previous fix #15756](https://github.com/keiyoushi/extensions-source/pull/15756) didn't work [without headers](https://github.com/keiyoushi/extensions-source/pull/15880/changes) in request. I assume Tachimanga handles them automatically in requests while android - does not, so add them.

martinbirn: You are right. I forgot to check android. My bad. I will do it

martinbirn: Thanks. You were right: my initial change in `e88aada26` regressed Mihon on Android.

I tested the exact pre-MR v39 build from a clean state, and the guard completed successfully. With `e88aada26`, Android rejected the guard redirect with `Denied starting an intent without a user gesture` and timed out.

Missing headers were not the cause. `.headers(headers)` here would add headers to the synthesized OkHttp response, not to the guard request made by WebView.

Fixed in `9298f2a40` by assigning a `WebViewClient` outside Tachimanga. Mihon needs it to keep the redirect inside WebView, while Tachimanga intercepts the challenge requests when a client is assigned.

I retested the current v40 build from a clean state. The guard completed in 13 seconds, and details, chapters, reader images, popular, latest, and search all worked.

src/ru/comx/src/eu/kanade/tachiyomi/extension/ru/comx/DleGuardResolver.kt
AwkwardPeak7, line 1: direct webview usage is discouraged. use the helpers `runWebViewBlocking`. if tachimanga doesn't work with it, report to the app dev. 
you may also patch the helper if needed

Exinaus: Maybe, if you just need a working bypass, restore previous as an option in preferences. 

Disable it by default, and then something like this:

```
    override fun OkHttpClient.Builder.configureClient() = apply {
        if ( preferences.getBoolean(USE_ALTERNATIVE_BYPASS, false) == false) {
            addInterceptor(DleGuardResolver.interceptor(baseUrl))
        } else {
            // add old interceptor
        }
        rateLimit(3)
    }
```

AwkwardPeak7:
I have informed Tachimanga dev, he will fix it in app.

martinbirn:
So, it looks like this PR can be closed for now while we wait for the fix on Tachimanga's side. I can reopen it if that fix turns out to be incomplete or does not fully resolve the issue.

PR #18695, opened 2026-08-28
DrakeScans & WitchScans: bump version

Exinaus:
Checklist:

<details>
  <summary>DrakeScans info</summary>
Screenshot 2026-08-28 at 22-45-56 Update extensions repo · keiyoushi_extensions@b5a5c74
</details>

48 + 1 (from multisrc) = 49

<details>
  <summary>WitchScans info</summary>

Screenshot 2026-08-29 at 01-36-35 Update extensions repo · keiyoushi_extensions@b5a5c74
</details>

32 + 1 (from multisrc) = 33

> New Version code is less that previous one so android doesn't let it "downgrade"

### DIFF
--- a/src/en/witchscans/build.gradle.kts
+++ b/src/en/witchscans/build.gradle.kts
@@ -6,7 +6,7 @@ plugins {
 
 keiyoushi {
     name = "WitchScans"
-    versionCode = 0
+    versionCode = 32
     contentWarning = ContentWarning.SAFE
     libVersion = "1.6"
     theme = "vinetheme"
--- a/src/en/witchscans/build.gradle.kts
+++ b/src/en/witchscans/build.gradle.kts
@@ -6,7 +6,7 @@ plugins {
 
 keiyoushi {
     name = "WitchScans"
-    versionCode = 32
+    versionCode = 0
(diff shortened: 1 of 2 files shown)


src/en/witchscans/build.gradle.kts
```diff
@@ -6,7 +6,7 @@ plugins {
 
 keiyoushi {
     name = "WitchScans"
-    versionCode = 0
+    versionCode = 32
```
Secozzi:
```suggestion
    versionCode = 33
```

Exinaus: 32 from extension + 1 from multisrc = 33

Secozzi: Oh woops, missed the multisrc

vetleledaal:
~~Do all apps treat 1.6.1 as older than 1.4.32? If they do then that's a bug with the apps. We should still work around the issue by using 1.6.33 here if that's the case.~~

> New Version code is less that previous one so android doesn't let it "downgrade"

AwkwardPeak7 (approved):

PR #18700, opened 2026-08-29
Allow overriding pkg suffix

AwkwardPeak7:
Checklist:


gradle/build-logic/src/main/kotlin/ExtensionPlugin.kt
```diff
@@ -316,6 +312,8 @@ class ExtensionPlugin : Plugin<Project> {
     }
 }
 
+private val APPLICATION_ID_SUFFIX_REGEX = Regex("^[a-z0-9_]+(\\.[a-z0-9_]+)+$")
```
vetleledaal: Use `"""` to avoid double escaping regex.
```suggestion
private val APPLICATION_ID_SUFFIX_REGEX = Regex("""^[a-z0-9_]+(\.[a-z0-9_]+)+$""")
```

.github/scripts/audit-packages.py
```diff
@@ -0,0 +1,61 @@
...
+
+    return errors
+
+
+def main() -> None:
+    errors = audit()
+
+    if errors:
+        for error in errors:
+            print(error)
```
vetleledaal: Consider using workflow commands: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#example-setting-an-error-message

Bonus if you can point to specific files/lines, but that requires additional changes.

.github/scripts/audit-packages.py
```diff
@@ -0,0 +1,61 @@
...
+        sys.exit(1)
+
+    modules = len(list(Path("src").glob("*/*/build.gradle.kts")))
```
vetleledaal: Double work- length should be returned from `audit()` -- prefer dataclass, but tuple works too for such small scripts.

.github/scripts/audit-packages.py
```diff
@@ -0,0 +1,72 @@
+import re
+import sys
+from pathlib import Path
+
+EXTENSION_PREFIX = "eu.kanade.tachiyomi.extension"
+PKG_NAME_REGEX = re.compile(r"""pkgName\s*=\s*["']([^"']+)["']""")
+SUFFIX_REGEX = re.compile(r"^[A-Za-z0-9_]+(\.[A-Za-z0-9_]+)+$")
```
vetleledaal: Can be simplified:
```suggestion
SUFFIX_REGEX = re.compile(r"^[\w+(\.\w+)+$")
```

gradle/build-logic/src/main/kotlin/ExtensionPlugin.kt
```diff
@@ -316,6 +312,8 @@ class ExtensionPlugin : Plugin<Project> {
     }
 }
 
+private val APPLICATION_ID_SUFFIX_REGEX = Regex("^[A-Za-z0-9_]+(\\.[A-Za-z0-9_]+)+$")
```
vetleledaal: Can be simplified:
```suggestion
private val APPLICATION_ID_SUFFIX_REGEX = Regex("""^\w+(\.\w+)+$""")
```

gradle/build-logic/src/main/kotlin/ExtensionPlugin.kt
vetleledaal: Outdated, see other comment.

vetleledaal (approved):

PR #18734, opened 2026-08-30
Fix a few bugs for Pawchive

towermom9:
Checklist:


towermom9:
this pr fixes 403 error when only full res is allowed to be used and fixed webview not showing image. Initially when open in webview, the link would have a **?f=filename** parameter that specifically tells pawchive to serve the file as a forced download (Content-Disposition: attachment) rather than displaying it normally as an image in the browser/app.

src/all/pawchive/src/eu/kanade/tachiyomi/extension/all/pawchive/Dto.kt
```diff
@@ -103,5 +103,5 @@ class PawchivePostDto(
 
 @Serializable
 class PawchiveFileDto(val name: String? = null, val path: String? = null) {
-    override fun toString() = path + if (name != null) "?f=$name" else ""
+    override fun toString() = path ?: ""
```
AwkwardPeak7: why fallback to empty? old code appends to the path

src/all/pawchive/src/eu/kanade/tachiyomi/extension/all/pawchive/Pawchive.kt
```diff
@@ -47,13 +47,25 @@ abstract class Pawchive :
     override fun OkHttpClient.Builder.configureClient() = apply {
         addInterceptor { chain ->
             val request = chain.request()
+            val urlString = request.url.toString()
 
-            val modifiedRequest = if (request.url.pathSegments.firstOrNull() == "api") {
-                request.newBuilder().header("Accept", "text/css").build()
+            val requestBuilder = request.newBuilder()
+                .header("Referer", "$baseUrl/")
```
AwkwardPeak7: referer is there by default

AwkwardPeak7 (changes requested):

src/all/pawchive/src/eu/kanade/tachiyomi/extension/all/pawchive/Dto.kt
towermom9: if "?f=filename" is added to the end of the link, then it would download the image instead of loading it. thats why the image doesn't load when open in webview.

src/all/pawchive/src/eu/kanade/tachiyomi/extension/all/pawchive/Pawchive.kt
towermom9: i see, let me fix that

ZenVinny:
closes #18747 if merged

towermom9:
> closes #18747 if merged

The bug report just got reported so I didn't tick the "Referenced all related issues in the PR body" mb mb

src/all/pawchive/src/eu/kanade/tachiyomi/extension/all/pawchive/Dto.kt
AwkwardPeak7: make path not nullable in the class

towermom9: @AwkwardPeak7 something like that?

AwkwardPeak7 (approved):

towermom9:
@AwkwardPeak7 soo approved to merge?

PR #18749, opened 2026-08-30
ToonLivre: update FCaptcha reader flow

### LINKED ISSUE #18745: ToonLivre(PT-BR) Timed out waiting for WebView after 1min30s
### Source information

ToonLivre 1.4.88

### Source language

Portuguese

### Steps to reproduce

1 Enter a manga
2 Try to open a chapter

### Expected behavior

Showing the pages

### Actual behavior

It doesn't, and says "Timed out waiting for WebView after 1min30s". The WebView works fine though

### Mihon/Mihon fork version

TachyiomiSY stable 1.13.2 

### Android version

Android 16

### Other details

_No response_

### Acknowledgements

### 

---

zLuckKing:
## Summary

Updates the ToonLivre reader after the site replaced its previous verification flow with FCaptcha.

- Checks `/api/reader/chapter/access` before opening a WebView.
- Opens the real chapter page in a visible WebView only when manual verification is required.
- Lets the user complete FCaptcha normally; no token generation, capture, reuse, or bypass is implemented.
- Automatically closes the verification screen once chapter images are available and returns to Mihon's native reader.
- Blocks top-level navigation to external advertising destinations while keeping page resources available.
- Prevents duplicate verification screens and false cancellation during orientation changes.
- Preserves strict CDN chapter filtering, proxy-to-CDN normalization, and numeric page ordering.
- Keeps the existing source ID, package, module, class, and `versionId`.

Closes #18745

## Why this is needed

Version 1.4.88 waits for the previous reader verification flow and now times out after 90 seconds because ToonLivre's current reader requires an explicit FCaptcha interaction. The new flow exposes that verification to the user, then hands the discovered chapter pages back to Mihon.

## Testing

Tested with release APK `1.4.89` on a physical Android device with Mihon:

- source details and chapter list load normally;
- FCaptcha is displayed and remains manually operated;
- external ad redirects do not replace the verification page;
- after verification, the WebView closes and the native reader opens automatically;
- all 16 pages load in the correct numeric order;
- reopening the chapter does not create a verification loop;
- `:src:pt:mangalivre:spotlessCheck` passed;
- `:src:pt:mangalivre:assembleRelease` passed.

### 1. ToonLivre title and chapter list in Mihon

ToonLivre title and chapter list in Mihon

### 2. Manual FCaptcha verification displayed for a protected chapter

Manual FCaptcha verification displayed for a protected chapter

### 3. Automatic transition after successful verification

Automatic transition after successful verification

### 4. Chapter opened in Mihon's native reader

Chapter opened in Mihon's native reader

### 5. Native reader showing page 1 of 16

Native reader showing page 1 of 16

## Checklist

- [x] Updated `baseVersionCode` in `build.gradle.kts` (if updated multisrc theme code) — N/A, this is not a multisrc theme change
- [ ] Have tested the modifications by compiling and running the extension through Android Studio — tested with a release APK built by GitHub Actions and installed on a physical device
- [x] Have removed `web_hi_res_512.png` when adding a new extension — N/A, this is an existing extension


AwkwardPeak7 (changes requested):
please migrate this source to 1.6 with KeiSource and use `runWebView` suspend helpers there

zLuckKing:
@AwkwardPeak7 Migrated the source to lib 1.6 with `KeiSource` and converted the network operations to the suspend helpers.

The visible verification Activity was intentionally kept because FCaptcha requires explicit user interaction. The shared `runWebView` helper creates a non-interactive WebView using the application context, so it cannot display the checkbox to the user. After the user completes the verification, the Activity closes automatically and returns the collected pages to Mihon's native reader.

I also fixed a duplicated `Origin` header exposed by the migration. The updated implementation passes `spotlessKotlinCheck`, `lintRelease`, and `assembleRelease`, and the generated APK was installed and tested successfully on a physical device.

AwkwardPeak7:
please test release build as well, (`assembleRelease`), r8 might strip the activity away. also can you change it to kotlin?

AwkwardPeak7:
@zLuckKing I know I said to use kotlin, but we probably should not be bundling kotlin stdlib. you may use kotlin without using the stdlib.
or just java is fine too ig.

zLuckKing:
@AwkwardPeak7 Thanks for the follow-up. I tested the Kotlin conversion with an `assembleRelease` APK on a physical device. Without bundling the Kotlin stdlib, the verification Activity crashed with `NoClassDefFoundError: kotlin.jvm.internal.Intrinsics`. Bundling the stdlib then caused a classloader conflict while Mihon loaded the source, preventing the extension from appearing.

Since you confirmed Java is acceptable, I restored the previously tested Java Activity and removed the bundled Kotlin stdlib. The new release APK passes `spotlessApply`, `lintRelease`, and `assembleRelease`. I installed it on a physical device and confirmed that the extension loads correctly, FCaptcha verification completes, the Activity closes automatically, and the chapter opens successfully in Mihon's native reader.

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/Dto.kt
```diff
@@ -81,13 +86,32 @@ class ChapterReferenceDto(
...
     val timestamp: Long,
 ) {
-    fun toSChapter(slug: String, mangaId: String) = SChapter.create().apply {
+    fun toSChapter(
+        slug: String,
+        mangaId: String,
+    ) = SChapter.create().apply {
         name = "Capítulo $number"
         date_upload = timestamp
         url = "/$slug/$number#${ChapterReferenceDto(mangaId, id).toJsonString()}"
```
AwkwardPeak7: store extra things in memo; keep url to a stable id or slug

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
```diff
@@ -1,92 +1,102 @@
...
+import kotlin.time.Duration.Companion.minutes
 import kotlin.time.Duration.Companion.seconds
 
 @Source
 abstract class MangaLivre :
-    HttpSource(),
+    KeiSource(),
     ConfigurableSource {
-
     private val baseUrlHost by lazy { baseUrl.toHttpUrl().host }
```
AwkwardPeak7: inline

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
```diff
@@ -1,92 +1,102 @@
...
 
     override val supportsLatest: Boolean = true
```
AwkwardPeak7: defaults to true

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
```diff
@@ -1,92 +1,102 @@
...
 
-    override val client: OkHttpClient = network.client.newBuilder()
-        .rateLimit(2, 1.seconds) { it.host == baseUrlHost }
-        .build()
+    override fun OkHttpClient.Builder.configureClient(): OkHttpClient.Builder = rateLimit(2, 1.seconds) { it.host == baseUrlHost }
 
     private val apiUrl: String = "$baseUrl/api"
```
AwkwardPeak7:
```suggestion
    private val apiUrl: String get() = "$baseUrl/api"
```

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
```diff
@@ -1,92 +1,102 @@
...
 
-    private val popularFilter = FilterList(
-        listOf(
-            OrderByFilter(options = listOf("" to SORT_POPULAR)),
-            OrderDirectionFilter(options = listOf("" to DIRECTION_DESC)),
-        ),
-    )
-
-    override fun popularMangaRequest(page: Int): Request = searchMangaRequest(page, "", popularFilter)
+    private val popularFilter =
```
AwkwardPeak7: inline these filter vals

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
```diff
@@ -116,91 +122,100 @@ abstract class MangaLivre :
...
+        }
+    }
 
-                javaScriptEnabled = true
-                domStorageEnabled = true
+    private suspend fun fetchReaderAccess(ref: ChapterReferenceDto): ReaderAccessResponseDto? {
+        val requestHeaders =
+            headers
+                .newBuilder()
+                .set("Origin", baseUrl)
```
AwkwardPeak7: origin is there by default

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/ReaderVerificationActivity.java
```diff
@@ -0,0 +1,181 @@
...
+          result = new ArrayList<>(pages);
+        }
+        delivered = true;
+        Bundle bundle = new Bundle();
+        bundle.putStringArrayList(EXTRA_PAGES, result);
+        receiver.send(RESULT_PAGES, bundle);
+        finish();
+      };
+
+  @SuppressWarnings("deprecation")
```
AwkwardPeak7, line 58: what deprecated thing are you using?

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/ReaderVerificationActivity.java
AwkwardPeak7, line 1: also look at japscans, would the approach used there (app's own WebView Activity) work here?

AwkwardPeak7 (changes requested):

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/Dto.kt
zLuckKing: Updated. The chapter URL now keeps only the stable slug/number path, while `mangaId` and `chapterId` are stored in `SChapter.memo`.

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
zLuckKing: Updated. The host lookup is now inlined in `configureClient()`.

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
zLuckKing: Removed the redundant `supportsLatest` override.

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
zLuckKing: Applied. `apiUrl` is now a computed property so it follows the injected `baseUrl`.

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
zLuckKing: Updated. The popular and latest filter values are now inlined at their call sites.

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
zLuckKing: Removed the explicit `Origin` header and kept the default request headers.

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/ReaderVerificationActivity.java
zLuckKing: The deprecated API is the pre-API 33 `Intent.getParcelableExtra(String)` overload used to retrieve the `ResultReceiver`. The extension supports Android versions below API 33, so the suppression is scoped to this compatibility call.

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/ReaderVerificationActivity.java
zLuckKing: I implemented and tested the Japscan-style approach using Mihon's own `WebViewActivity`. The verification completed and the chapter rendered inside that WebView, but the activity exposes no completion/result callback to the extension, and the site did not produce a stable observable cookie change that could be used as a completion signal. Polling `chapter/access` also renewed the challenge cookie on each 403 and interfered with the active verification. I therefore restored the dedicated activity: it does not automate the verification; it only observes chapter-specific CDN image requests after the user completes it, returns those pages through the `ResultReceiver`, and calls `finish()`. The final release build was installed and tested successfully on a physical device.

AwkwardPeak7 (approved):

PR #18816, opened 2026-09-03
Manhwa XXL -> HentaiTnT: fix pagination and status

vetleledaal:
PR does not address:
- ~~1.6~~
- `date_upload`

~~Unsure if we need to keep `versionId` when `id` is set.~~

Checklist:


src/en/hentaitnt/build.gradle.kts
```diff
@@ -5,14 +5,16 @@ plugins {
...
+    pkgName = "en.manhwaxxl"
+    versionCode = 7
     contentWarning = ContentWarning.NSFW
     libVersion = "1.4"
 
     source {
         lang = "en"
         baseUrl = "https://hentaitnt.net"
+        id = 1823408414151467671L
         versionId = 2
```
AwkwardPeak7: no need when id is hardcoded

AwkwardPeak7 (commented):
consider also migrating to 1.6 with KeiSource if you are up for it

src/en/hentaitnt/src/eu/kanade/tachiyomi/extension/en/hentaitnt/HentaiTnT.kt
```diff
@@ -74,27 +76,37 @@ abstract class ManhwaXXL :
...
                 else -> SManga.UNKNOWN
             }
         }
     }
 
-    override fun chapterListParse(response: Response): List<SChapter> {
-        val detailsDocument = response.asJsoup()
+    private suspend fun fetchChapters(manga: SManga): List<SChapter> {
+        val detailsDocument = client.get(getMangaUrl(manga)).asJsoup()
         val mangaId = detailsDocument.selectFirst("#post_manga_id")?.attr("value")
```
AwkwardPeak7: save mangaId in `manga.memo` so next time document is not needed for chapters

src/en/hentaitnt/src/eu/kanade/tachiyomi/extension/en/hentaitnt/HentaiTnT.kt
```diff
@@ -74,30 +79,63 @@ abstract class ManhwaXXL :
...
+    override suspend fun fetchMangaUpdate(
+        manga: SManga,
+        chapters: List<SChapter>,
+        fetchDetails: Boolean,
+        fetchChapters: Boolean,
+    ): SMangaUpdate {
+        val mangaId = manga.memo.getStringOrNull("id")
+
+        if (!fetchChapters) {
+            val updatedManga = if (fetchDetails) fetchMangaDetails(manga) else manga
```
AwkwardPeak7, line 95: if one is false, the other is guaranteed (by KeiSource) to be true, so this if is no-op

AwkwardPeak7 (approved):

src/en/hentaitnt/src/eu/kanade/tachiyomi/extension/en/hentaitnt/HentaiTnT.kt
vetleledaal: It's easier to reason about the code if you ignore such implementation details and rather assume any input to the function is valid. I think it's fine to leave as-is.

PR #18832, opened 2026-09-04
MangaFox: migrate to 1.6

vetleledaal:
Not sure what the status was on bumping `versionCode` when going from 1.4 to 1.6. The new version here is in any case 1.6.1. I can change it to 1.6.10 if desired.

Checklist:

### DIFF
--- a/src/en/mangafox/src/eu/kanade/tachiyomi/extension/en/mangafox/MangaFox.kt
+++ b/src/en/mangafox/src/eu/kanade/tachiyomi/extension/en/mangafox/MangaFox.kt
@@ -1,42 +1,39 @@
 package eu.kanade.tachiyomi.extension.en.mangafox
 
 import android.webkit.CookieManager
-import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.source.model.Filter
 import eu.kanade.tachiyomi.source.model.FilterList
 import eu.kanade.tachiyomi.source.model.MangasPage
 import eu.kanade.tachiyomi.source.model.Page
 import eu.kanade.tachiyomi.source.model.SChapter
 import eu.kanade.tachiyomi.source.model.SManga
-import eu.kanade.tachiyomi.source.online.HttpSource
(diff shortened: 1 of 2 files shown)


src/en/mangafox/src/eu/kanade/tachiyomi/extension/en/mangafox/MangaFox.kt
```diff
@@ -1,42 +1,39 @@
...
 
     private val mobileUrl get() = baseUrl.replace("://", "://m.")
 
-    override val supportsLatest: Boolean = true
-
-    override val client: OkHttpClient = network.client.newBuilder()
-        // Force readway=2 cookie to get all page URLs at once
-        .cookieJar(
+    override fun OkHttpClient.Builder.configureClient() = // Force readway=2 cookie to get all page URLs at once
+        cookieJar(
```
AwkwardPeak7: use `keiyoushi.utils.setCookie` interceptor

src/en/mangafox/src/eu/kanade/tachiyomi/extension/en/mangafox/MangaFox.kt
AwkwardPeak7:
```suggestion
            filters.forEach { filter ->
```

AwkwardPeak7, line 1: please also replace `select().first()` with `selectFirst()`
also `text().ifBlank` with `textOrNull()`

AwkwardPeak7 (changes requested):

src/en/mangafox/src/eu/kanade/tachiyomi/extension/en/mangafox/MangaFox.kt
vetleledaal: I switched to using `addCookie`, but had to change the order. Otherwise the cookies for the subdomain were never set.

AwkwardPeak7 (approved):

PR #18124, opened 2026-07-31
MangaDex: migrate to 1.6

tigeryu8900:
This change is a bit hacky due to Komikku using MangaDex's `latestUpdatesRequest` and `latestUpdatesParse`.

Checklist:

- [x] Updated `versionCode` value in `build.gradle.kts`
- [ ] Updated `baseVersionCode` in `build.gradle.kts` (if updated multisrc theme code)
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Set the `contentWarning` configuration in `build.gradle.kts` appropriately
- [x] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [ ] Have removed `web_hi_res_512.png` when adding a new extension
- [ ] This PR is AI-assisted, I have reviewed the changes manually and confirmed they are not slop

PR #18899, opened 2026-09-07
MangaDex: Migrate to 1.6

tigeryu8900:
Since the recent changes to ext lib, it is now possible to update MangaDex to 1.6 without it breaking on Komikku.

Continuation of #18124

Checklist:

### DIFF
--- a/src/all/mangadex/build.gradle.kts
+++ b/src/all/mangadex/build.gradle.kts
@@ -6,9 +6,9 @@ plugins {
 
 keiyoushi {
     name = "MangaDex"
-    versionCode = 212
+    versionCode = 213
     contentWarning = ContentWarning.MIXED
-    libVersion = "1.4"
+    libVersion = "1.6"
 
     listOf(
         "af", "sq", "ar", "az", "eu", "be", "bn", "bg", "my", "ca", "zh-Hans", "zh-Hant",
--- a/src/all/mangadex/build.gradle.kts
+++ b/src/all/mangadex/build.gradle.kts
@@ -6,7 +6,7 @@ plugins {
 
 keiyoushi {
     name = "MangaDex"
-    versionCode = 213
+    versionCode = 0
     contentWarning = ContentWarning.MIXED
     libVersion = "1.6"
 
--- a/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
+++ b/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
@@ -46,7 +46,6 @@ import okhttp3.Request
 import okhttp3.Response
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
-import java.util.Date
 
 @Source
 class MangaDex(
--- a/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
+++ b/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
@@ -1,5 +1,6 @@
 package eu.kanade.tachiyomi.extension.all.mangadex
 
+import android.app.Application
 import android.content.SharedPreferences
 import android.os.Build
 import androidx.preference.EditTextPreference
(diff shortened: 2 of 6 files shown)


src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
```diff
@@ -62,35 +65,36 @@ class MangaDex(
...
+    @Suppress("DEPRECATION")
+    private val appPreferences by lazy { PreferenceManager.getDefaultSharedPreferences(applicationContext)!! }
+
+    private val komikku = applicationContext.packageName.startsWith("app.komikku")
+
+    private fun delegateSources() = try {
+        appPreferences.getBoolean("eh_delegate_sources", komikku)
+    } catch (_: ClassCastException) {
+        false
+    }
```
AwkwardPeak7, line 80: inline all these into the `delegateSources` function and catch `Throwable` since this can break suwayomi otherwise.

also what is the purpose?

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
```diff
@@ -62,35 +65,36 @@ class MangaDex(
...
 
-    override val client = network.client.newBuilder()
-        .rateLimit(3)
-        .build()
+    override fun OkHttpClient.Builder.configureClient(): OkHttpClient.Builder = rateLimit(3)
 
     // Popular manga section
 
-    override fun popularMangaRequest(page: Int): Request {
+    override suspend fun getPopularManga(page: Int): MangasPage {
```
AwkwardPeak7, line 98: I was thinking changing popular to popular today/this week that the website shows at the top, since popular all time is basically useless since the dmca

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
```diff
@@ -101,10 +105,10 @@ class MangaDex(
             .addQueryParameter("originalLanguage[]", preferences.originalLanguages)
             .build()
 
-        return GET(url, headers, CacheControl.FORCE_NETWORK)
+        return parseMangasPage(client.get(url, CacheControl.FORCE_NETWORK))
```
AwkwardPeak7: no need for force network

src/all/mangadex/build.gradle.kts
```diff
@@ -6,9 +6,9 @@ plugins {
 
 keiyoushi {
     name = "MangaDex"
-    versionCode = 212
+    versionCode = 213
```
AwkwardPeak7: reset it to 1 or 0, app will update based on 1.4 -> 1.6

AwkwardPeak7 (changes requested):

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
tigeryu8900: This determines whether to include future updates when retrieving the latest updates and searching by uploader. A chapter with a future update is an external chapter, handled only when delegate sources are enabled.

Currently, Komikku handles the latest updates case by updating the request URL but does not handle the search case.

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
tigeryu8900: I added the `createdAtSince` parameter set to 30 days ago at midnight in the local time zone, matching what the website uses for popular new titles.

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
```diff
@@ -62,35 +67,30 @@ class MangaDex(
...
         }
 
-    override val supportsLatest = true
-
     private val preferences by getPreferencesLazy { sanitizeExistingUuidPrefs() }
 
+    @Suppress("DEPRECATION")
+    private val appPreferences by lazy { PreferenceManager.getDefaultSharedPreferences(applicationContext)!! }
+
+    private val komikku = applicationContext.packageName.startsWith("app.komikku")
```
AwkwardPeak7: I meant inline these as well, in a `isDelegated` function/getter property for example. wrapped in try catch

tigeryu8900:
I also updated the `isInvalid` check to include removed external chapters

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
```diff
@@ -62,35 +67,33 @@ class MangaDex(
             else -> lang
         }
 
-    override val supportsLatest = true
-
     private val preferences by getPreferencesLazy { sanitizeExistingUuidPrefs() }
 
+    @Suppress("DEPRECATION")
+    private fun isDelegate() = try {
+        val komikku = applicationContext.packageName.startsWith("app.komikku")
```
AwkwardPeak7: btw sy also has delegated

maybe better option would be to check if app has `eh_delegate_sources` preference

tigeryu8900: That preference might not exist in the preference store if it has not been touched. Currently in this PR, it checks for that preference and defaults to true for Komikku and false for other apps.

The only solution I can think of for now is to use reflection to load the `DelegateSourcePreferences` class and check if it has the `delegateSources` method.

tigeryu8900: That preference might not exist in the preference store if it has not been touched. Currently in this PR, it checks for that preference and defaults to true for Komikku and false for other apps.

The only solution I can think of for now is to use reflection to load the

AwkwardPeak7: then just add check for sy package, no need for anything else

AwkwardPeak7 (approved):
I was also thinking about changing the default content rating preference to be "up to x", so selecting erotica enables all three below. and setting default to erotica.
useful since disabling all now makes it default to site behaviour which is to enable safe,suggestive,erotica when none specified

I will merge this now, but feel free to do it in a followup pr if you want, otherwise I will do it myself later.

PR #18500, opened 2026-08-19
Add Guazimanhua (瓜子漫画) Chinese source

muzhuoyun:
Closes #18423

### Source
**guazimanhua** (瓜子漫画) - Chinese (zh) - https://www.guazimanhua.com/

### Summary
Adds a new Chinese source for 瓜子漫画 (guazimanhua.com), a Chinese manhua site. Supports popular / latest listings, text search, filtering (genre, region, audience, status, sort), manga details, chapters, and page reading. Uses a desktop User-Agent because the site serves a "download our app" page (no images) for the latest chapter when it detects an Android mobile browser UA.

### Checklist
- [x] Updated `versionCode` value in `build.gradle.kts`
- [x] Updated `baseVersionCode` in `build.gradle.kts` (if updated multisrc theme code)
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Set the `contentWarning` configuration in `build.gradle.kts` appropriately
- [x] Have not changed source names
- [x] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio
- [x] Have removed `web_hi_res_512.png` when adding a new extension
- [x] This PR is AI-assisted, I have reviewed the changes manually and confirmed they are not slop

### AI-assisted disclosure
I was asked to build and submit this extension with the help of an AI agent. I reviewed the code, verified it builds (`assembleRelease`), passes `lintRelease`, and tested it end-to-end on the Mihon app (popular/latest/search/filter/detail/reader/download all work).

PR #18989, opened 2026-09-10
Guazimanhua: fetch the newest chapter through the app API

Catapult291:
### What

A freshly published chapter is served to Android / Dalvik user agents without any page images: `/chapter.php?id=...` comes back with the whole `section.reader-images` block missing (only an app download prompt remains), so the reader ends up empty. A desktop `User-Agent` receives those pages, and the official app receives them too, so this PR keeps the desktop `User-Agent` from #18500 and reads the chapter from the app API whenever the web path still yields no images.

### Changes

- **App API fallback (`GuaziAppApi.kt`, new).** Reconstructed from `guazi.apk` (v1.6.1, versionCode 53): guest login `POST /api/v2/login/visitor` (needs only `identifier`, token cached in memory) and `GET /api/v2/mcomic/pics?chapter_id={id}`. Both carry the global `identifier` + `versionCode` params and the `deviceType` / `token` headers. Each `images[].img` is base64 of AES/CBC ciphertext: the key is the MD5 hex string of `"guazi" + yyyyMMdd` used as 32 raw bytes, the IV is characters 8..24 of the same string, and the date is taken from the response `Date` header converted to `Asia/Shanghai` so a device in another time zone derives the key of the right day. `getPageList` only calls this when the web page yielded no images, so every other chapter is unaffected and the app API being unavailable cannot break normal reading.
- **Kept the desktop `User-Agent` override** from #18500. Re-checked on 2026-09-15: the withholding applies to freshly published chapters — of the 12 most recently updated chapters, 10 answered the mobile UA with the app-download stub (≈5.1 KB, no `reader-images`) and all 10 returned their full page list to the desktop UA (11–45 images). An earlier check of mine had shown byte-identical responses for mobile, desktop and Dalvik, but it ran on chapters whose gate had already lifted; that conclusion was wrong.
- **Chapter list**: reversed to oldest first and numbered by list position (`chapter_number` 1..N). The site's own numbering cannot be used because the list also contains unnumbered 公告 / 预告 / 角色档案 entries.
- **Parsing**: `getMangaUrl()` / `getChapterUrl()` instead of string concatenation, `mapNotNull` for list cards, `absUrl` for covers, `hasNextPage` as a private helper.
- `versionCode` 6 → 7.

### Verification

- User-Agent gate, 2026-09-15, the 12 newest chapters on `/update.php`:

  ```
  chapter   mobile UA                     desktop UA
  2549649    4 images                      4 images      (gate already lifted)
  2549648   no reader-images (5,115 B)     29 images
  2549646   no reader-images (5,122 B)     11 images
  2549627   no reader-images (5,115 B)     37 images
  2549620   no reader-images (5,115 B)     23 images
  2549618   no reader-images (5,116 B)     31 images
  2549616   no reader-images (5,115 B)     33 images
  2549614   25 images                     25 images      (gate already lifted)
  2549613   no reader-images (5,116 B)     27 images
  2549611   no reader-images (5,114 B)     16 images
  2549605   no reader-images (5,120 B)     37 images
  2549602   no reader-images (5,115 B)     45 images
  ```
- Desktop web reader vs app API on gated chapters: identical page lists, first and last image included (29 = 29, 37 = 37, 45 = 45, 56 = 56).
- The app API flow was replayed outside the app on a JVM (same endpoints, params, headers, `Date` → key date derivation and AES/CBC decryption as in `GuaziAppApi.kt`): guest token issued, ciphertext decrypted to absolute `img.guazicdn.com` URLs, first image returns `HTTP 200`.
- Page counts of the newest chapter were compared between the web reader and the app API for the 12 most recently updated titles: identical (34, 25, 21, 19, 19, 21, 21, 29, 19, 27, 32, 22).
- `assembleRelease`, `spotlessCheck` and `lintRelease` pass locally; the generated manifest still carries the existing deep links unchanged.
- Reading in Mihon was confirmed on device with the 1.6.9 build of this change.

### Notes

- `VERSION_CODE = "53"` is copied from the current app APK. If the site starts rejecting older app versions the API call fails and the affected chapters go back to being unreadable until that constant is updated.
- On app API errors the source throws rather than returning an empty list, so the app shows a localized error instead of a silently empty chapter.

Checklist:

🤖 Disclosure: I was asked to make the newest chapter of this source readable by using the app API and to open this PR. The extension code, the reverse engineering of the app API and the verification above were produced with an AI agent; I reviewed the changes and built the APK myself.

### DIFF
--- a/src/zh/guazimanhua/build.gradle.kts
+++ b/src/zh/guazimanhua/build.gradle.kts
@@ -6,7 +6,7 @@ plugins {
 
 keiyoushi {
     name = "Guazimanhua"
-    versionCode = 6
+    versionCode = 7
     contentWarning = ContentWarning.SAFE
     libVersion = "1.6"
 
--- a/src/zh/guazimanhua/src/eu/kanade/tachiyomi/extension/zh/guazimanhua/Dto.kt
+++ b/src/zh/guazimanhua/src/eu/kanade/tachiyomi/extension/zh/guazimanhua/Dto.kt
@@ -1,0 +1,34 @@
+package eu.kanade.tachiyomi.extension.zh.guazimanhua
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+class AppTokenResponse(
+    @SerialName("error_code")
+    val errorCode: Int = 0,
+    val data: AppTokenData? = null,
+)
+
+@Serializable
+class AppTokenData(
+    val token: String? = null,
+)
+
+@Serializable
+class AppPicsResponse(
+    @SerialName("error_code")
+    val errorCode: Int = 0,
+    val data: AppPicsData? = null,
+)
+
+@Serializable
+class AppPicsData(
+    val images: List<AppImage> = emptyList(),
+)
+
+@Serializable
+class AppImage(
+    // base64 of AES/CBC ciphertext, decrypts to the absolute image URL
+    val img: String,
+)
--- a/src/zh/guazimanhua/src/eu/kanade/tachiyomi/extension/zh/guazimanhua/GuaziAppApi.kt
+++ b/src/zh/guazimanhua/src/eu/kanade/tachiyomi/extension/zh/guazimanhua/GuaziAppApi.kt
@@ -1,0 +1,140 @@
+package eu.kanade.tachiyomi.extension.zh.guazimanhua
+
+import android.content.SharedPreferences
+import android.util.Base64
+import keiyoushi.network.get
+import keiyoushi.network.post
+import keiyoushi.utils.parseAs
+import okhttp3.FormBody
+import okhttp3.Headers
+import okhttp3.HttpUrl
+import okhttp3.HttpUrl.Companion.toHttpUrl
+import okhttp3.OkHttpClient
+import java.security.MessageDigest
+import java.time.Instant
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
+import java.util.Date
+import java.util.Locale
+import java.util.UUID
+import javax.crypto.Cipher
+import javax.crypto.spec.IvParameterSpec
+import javax.crypto.spec.SecretKeySpec
+
+/**
+ * Client for the 瓜子漫画 app API, reconstructed from guazi.apk (v1.6.1, versionCode 53).
+ *
+ * - Every request carries the global params identifier + versionCode and the headers
+ *   deviceType: android and token: <guest token>; the app injects those from one place
+ * - Guest login: POST /api/v2/login/visitor with only identifier, the token is cached in memory
+ * - Pages: GET /api/v2/mcomic/pics?chapter_id={id}, each images[].img is base64 of AES/CBC ciphertext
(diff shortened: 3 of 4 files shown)


AwkwardPeak7:
doesn't the desktop useragent serve the actual chapter? also is this an official publisher of some capacity?

Catapult291:
Good catch on the user agent — you're right, thanks for the pointer.

Re-checked just now (2026-09-15): of the 12 most recently updated chapters on `/update.php`, 10 answered a mobile UA with the app-download stub (~5.1 KB, no `reader-images`) and all 10 returned their full page list to the desktop UA (11–45 images). For three of them the desktop-UA page list is identical to the app API's, first and last image included (29 = 29, 37 = 37, 45 = 45). Within a series the gate sits on its newest chapter only: comic 18741 chapter 85 is gated for the mobile UA, chapter 84 is not.

The matrix in the PR description came from a chapter that was not gated at the time (1041107 — an old chapter by now, byte-identical for every UA), which is why it wrongly concluded the user agent no longer matters. The desktop `User-Agent` override is restored in bc8f741.

That leaves the app API: with the user agent back it only runs when the web reader still returns no images, which I can no longer reproduce. Do you want `GuaziAppApi.kt` dropped and this PR reduced to the user agent plus the parsing and chapter-numbering changes, or kept as a fallback? Either is fine by me.

On your second question — not an official publisher as far as I can tell: it is an aggregator (~23k titles, One Piece / Naruto / 一人之下 / 斗罗大陆 among them) whose footer carries placeholder legal text ("京ICP备案信息示例" / "网络文化经营许可证示例") and names no company or rights holder. If you know something I don't, say so and I'll drop the API path.

Catapult291:
Numbers for the question above: I compared the desktop web reader against the app API for the 13 most recently published chapters (the 12 from `/update.php` plus `2521681`, the chapter that first looked app-only). All 13 return an identical page list, first and last image included — `2521681` comes back with its 42 pages to the desktop UA as well. So with the user agent restored, I cannot point at a chapter that still needs the app API.

The one case the user agent does not cover is `/chapter.php` failing outright: on 2026-09-11 the site answered every chapter page with a 429 for a while (a real browser got the same), and then only the app API returns anything. I can either drop `GuaziAppApi.kt` and keep this PR to the user agent plus the parsing and chapter-numbering changes, or keep it and let it also kick in on a non-2xx chapter page. Your call.

Catapult291:
Closing this — your nudge was the right one.

Re-testing showed the premise was wrong: the source on `main` already sets a desktop `User-Agent` (#18500), so the newest chapter is readable as-is. My earlier matrix had been taken on a chapter whose gate had already lifted, which is what led me to remove it here.

With that gone, what is left is the app API fallback, and it only kicks in when `/chapter.php` itself yields no pages (the transient 429 case of 2026-09-11). For a source whose newest chapter the user agent already covers, I don't think that is worth keeping ~170 lines of reverse-engineered app protocol plus a hardcoded app `versionCode` in the repo. The chapter list is also reordered oldest first here, which contradicts CONTRIBUTING's "sorted descending by the source order", so that would have to go anyway.

Keeping the fallback in my own build only. Thanks for the review.

PR #18997, opened 2026-09-11
ZonaTMO.org (unoriginal): Fix and speed up latest updates

Sagenav0:
## What this fixes

The current Latest implementation uses `/biblioteca` ordered by `release_date`. That is the manga's release date, not the chapter upload time, so a manga receiving a new chapter is not reliably surfaced in actual upload order.

`/ultimas-subidas` is ZonaTMO's real chronological upload feed. However, it contains one row per chapter/upload rather than one row per manga, so parsing those rows directly would create duplicate manga entries.

## Example

If chapters 83–90 of **Conquistando la Academia con solo un Cuchillo de Sashimi** are uploaded, `/ultimas-subidas` can contain eight rows for that one manga. They all belong to `/library/manhwa/4182/conquistando-la-academia-con-solo-un-cuchillo-de-sashimi`, so Latest must collapse them into one manga entry while preserving their upload position.

## Implementation

- Parse the real order from `/ultimas-subidas` and deduplicate rows by title and type.
- Use the same stable `/biblioteca?title=...&type=...` `SManga.url` identity in Latest, Popular, Search, and URL imports, so Mihon recognizes the same manga across sections.
- Store the actual `/library/{type}/{id}/{slug}` path in `manga.memo` whenever Popular, Search, or URL import already provides it.
- For a Latest entry without memo, resolve the canonical path once in `fetchMangaUpdate`, return it in `manga.memo`, and reuse it on subsequent updates.
- Continue fetching existing `/library/` library entries directly. No `/view_uploads/` manga-entry migration is included because that implementation was never shipped.
- Keep Latest listing at one request per page; it does not call the suggestion endpoint for every listed manga.

## Validation

- Live page 1: 48 upload rows collapsed into 16 manga entries.
- All 16 title/type resolutions matched the expected canonical manga.
- Live pages 2, 10, and 20 remained non-empty.
- GitHub Actions: module build, `assembleRelease`, Spotless, and `lintRelease` passed for commit `9128c8c`; an APK artifact was produced.

## Checklist

- [ ] Updated `baseVersionCode` in `build.gradle.kts` (not applicable; this is not a multisrc change)
- [x] Referenced all related issues in the PR body (no related issue)
- [x] Have explicitly kept the `id` if a source's name or language were changed (not applicable; neither changed)
- [ ] Have tested the latest revision by compiling and running the extension (CI compiled it successfully; the preceding revision was tested in Mihon)
- [ ] Have removed `web_hi_res_512.png` (not applicable; this is not a new extension)

🤖 I was asked to fix inaccurate, duplicate, and slow ZonaTMO Latest entries. This PR was opened and updated by an AI agent after the changes were reviewed and tested with live site data by the user.


AwkwardPeak7:
what is it trying to fix? please give example. 
also seem like you are calling suggest api endpoint for almost every manga in latest, which is n calls. not ideal, can we avoid that

Sagenav0:
Thanks — the concrete issue is that `/ultimas-subidas` returns a different `/view_uploads/{chapterId}` URL for every chapter. For example, it can list chapters 83–90 of **Conquistando la Academia con solo un Cuchillo de Sashimi** as eight separate URLs even though they all belong to `/library/manhwa/4182/conquistando-la-academia-con-solo-un-cuchillo-de-sashimi`. Mihon identifies a manga by `SManga.url`, so those chapter URLs appeared as duplicate manga entries.

You were right about the suggest calls. A cold page had 48 upload rows representing 16 unique manga, which still meant up to 16 extra requests.

I changed the implementation so opening a Latest page now performs only the single `/ultimas-subidas` request. Each row receives a deterministic `/biblioteca?title=...&type=...` identity, and rows are deduplicated by that identity. The canonical `/library/` URL is resolved only when details or chapters are requested. Legacy `/view_uploads/` library entries remain supported.

Validation against the live site:

- Page 1: 48 uploads → 16 stable manga entries.
- All 16 deferred resolutions matched the expected title and type.
- Pages 2, 10, and 20 remained non-empty, avoiding the deep-pagination error.
- The latest GitHub Actions module build, `assembleRelease`, Spotless, and `lintRelease` all passed and produced the APK artifact.

This keeps the Latest listing at O(1) requests per page instead of O(N).

AwkwardPeak7:
you still didn't answer what is this trying to fix. e.g what is wrong with the current one?
also there are no library entries with `/view_uploads/` since this was never shipped...
also using different urls here means mihon will not recognize them as in library. so maybe save same kind of `/biblioteca?title=...&type=...` for all manga. and save the actual id/slug url in manga.memo; and for the latest section, populate manga.memo in `fetchMangaUpdate` once if it is not already there, instead of requesting url every time.

Sagenav0:
Thanks, this is updated now.

The current implementation of Latest uses `/biblioteca` ordered by `release_date`; that is the manga release date, not the chapter upload time, so newly uploaded chapters are not reliably shown in actual upload order. `/ultimas-subidas` is the real chronological feed, but it contains one row per chapter/upload, which is why those rows must be grouped into one manga entry.

I applied your identity/memo suggestion:
- Latest, Popular, Search, and URL imports now use the same `/biblioteca?title=...&type=...` `SManga.url`.
- Popular/Search/import store the known `/library/{type}/{id}/{slug}` path in `manga.memo`.
- Latest resolves that path once in `fetchMangaUpdate` only when memo is missing, then returns it in memo for reuse.
- Existing `/library/` entries are fetched directly.
- I removed the unused `/view_uploads/` manga-entry migration path.

The Latest listing remains one request per page, and subsequent updates do not repeat the suggestion lookup after memo is populated. I also updated the PR description with the concrete before/after example. CI is green for `9128c8c` and produced the APK artifact.

AwkwardPeak7 (approved):

PR #19068, opened 2026-09-13
Improve tag search for hentaihand/nhentai 

0x2d5:
Change in behavior: 
Tapping a genre tag and tapping search in the app filters with the tag instead of doing a plain text search. 
Searching with an exact text match as a tag would filter by the tag instead of doing a plain text search.
Example: Searching "elf" (which is a valid tag) would now filter by the tag "elf" instead of seaching the text "elf". 
This functionality is added for tag, artist and character. 
This also works around a problem where nhentai.com will return http 500 on searche requests. 

Downsides: 
During the lookup process it will try to resolve the search content as a tag. 
So if you are actually doing a free text search it will do 3 requests trying to resolve it as a tag, artist and character and the 4th request would actually send the search. 

Changes:
- `searchMangaRequest`: pure text queries (no lookup filter set) are now
  resolved to a tag/artist/character/parody/group/category/language id via
  the existing lookup endpoints, and searched by `uri[0]=id` with no `q`.
  Falls back to `q` when nothing matches, preserving title search.
- `lookupFilterId`: builds the lookup URL with `HttpUrl.Builder` (proper
  encoding of names with spaces) and prefers an exact case-insensitive
  name match (`"Schoolgirl"` previously resolved to `"Schoolgirl Uniform"`).
- `IdDto`: adds `name` 

Checklist:

- [x] Have tested the modifications by compiling and running the extension through Android Studio (Tested on a device running Android 14)


lib-multisrc/hentaihand/src/eu/kanade/tachiyomi/multisrc/hentaihand/HentaiHand.kt
```diff
@@ -85,32 +85,59 @@ abstract class HentaiHand :
...
                 }
             }.toBlocking().first()
     }
 
     override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
         val url = "$baseUrl/api/comics".toHttpUrl().newBuilder()
             .addQueryParameter("page", page.toString())
-            .addQueryParameter("q", query)
+
+        val effectiveFilters = if (filters.isEmpty()) getFilterList() else filters
```
AwkwardPeak7, line 114: no need for this

lib-multisrc/hentaihand/src/eu/kanade/tachiyomi/multisrc/hentaihand/HentaiHand.kt
```diff
@@ -85,32 +85,59 @@ abstract class HentaiHand :
...
+        // which breaks tapping a genre tag (the app searches for the tag name as text).
+        // Resolve the query to a tag/artist/character id when possible and search by id instead.
+        var queryFilter: Pair<String, Int>? = null
+        val trimmedQuery = query.trim()
+        if (trimmedQuery.isNotEmpty() && !hasLookupState) {
+            for (uri in QUERY_LOOKUP_URIS) {
+                val id = runCatching { lookupFilterId(trimmedQuery, uri, exactMatchOnly = true) }.getOrNull() ?: continue
+                queryFilter = uri to id
+                break
+            }
```
AwkwardPeak7, line 127: cache id in memory in a hashMap with removeEldest to something like 20; this way page > 1 dont have to fetch again

lib-multisrc/hentaihand/src/eu/kanade/tachiyomi/multisrc/hentaihand/HentaiHand.kt
```diff
@@ -137,7 +142,7 @@ abstract class HentaiHand :
             url.addQueryParameter("languages[${-index - 1}]", it.toString())
         }
 
-        effectiveFilters.forEach { filter ->
+        (if (filters.isEmpty()) getFilterList() else filters).forEach { filter ->
```
AwkwardPeak7, line 145: no need for this

PR #19177, opened 2026-09-17
Astramanga (ru): migrate to 1.6

Exinaus:
Checklist:

- migrated to `KeiSource`
- added deeplink
- about `fetchMangaUpdate` and changes in fetching chapters.
1. Not using `async` to fetch manga and chapters since chapters request requires `id` (manga can be fetched with `slug`). 
2. Changed getting chapters logic since it was getting only the branch labeled as `main`, leaving a lot of chapters in some cases. For example: https://astramanga.org/manga/nanomashiny?tab=chapters - only chapters from 309 to 329 were fetched, ignoring chapters from 1 to 308 (old translator stopped translating and new one was marked as `main`). Translator name now assigned properly and users should use app filtering by translator if they need only one branch.
- Changed fallback request size to 5000, since it might be the limit that [api returns](https://api.astramanga.org/api/v1/branches/17119/chapters?page=1&page_size=10000) even when requesting more than 5000.
- Maybe too much of `async` for fetching chapters? Manga with most chapters has ~1500, while limit are 5000. Maybe just do a single request to a branch, instead of doing `async` for a theoretical pagination?
- Updated filters: new sort options, and tags. Not sure if it's possible to fetch them from site. Tags api [returns list with two tags](https://api.astramanga.org/api/v1/tags), and tags on site are not loaded by default. Site loads them with JS when user clicks on menu to open a list of tags (list is cached somewhere in JS). Too much trouble to fetch them.

.


src/ru/astramanga/src/eu/kanade/tachiyomi/extension/ru/astramanga/AstraManga.kt
```diff
@@ -1,139 +1,160 @@
...
+        fetchDetails: Boolean,
+        fetchChapters: Boolean,
+    ): SMangaUpdate {
+        val slug = manga.url
+        val mangaData = client.get("$apiUrl/titles/${manga.url}").parseAs<TitleDetailResponse>()
+
+        val mangaNew = mangaData.data.toSMangaDetails(mediaUrl)
+
+        val chaptersNew = if (fetchChapters) {
+            val titleId = mangaData.data.id
```
AwkwardPeak7: save id in memo so next time manga details are not required

AwkwardPeak7 (approved):

PR #19201, opened 2026-09-18
Rüya Manga: fix covers served through Photon (403)

### LINKED ISSUE #19146: Rüya Manga: cover image fails to load for some series (CDN returns 403)
### Source information

Rüya Manga 1.4.4 (tr, `src/tr/ruyamanga`, base URL https://www.ruyamanga2.com)

### Source language

Turkish

### Steps to reproduce

1. Open one of these series in the app (or refresh its details):
   - https://www.ruyamanga2.com/manga/barbarin-oyun-rehberi/
   - https://www.ruyamanga2.com/manga/cigir-acan-oyuncu/
2. Look at the cover / thumbnail.

### Expected behavior

Cover image loads.

### Actual behavior

Cover never loads (Suwayomi returns 500 for `/api/v1/manga/{id}/thumbnail`; in Mihon the cover stays blank). Chapters and pages work fine.

The thumbnail URL the source returns is the site's origin CDN, e.g.
- https://cdn1.cdnformanga.xyz/uploads/manga/3321/cover.webp
- https://cdn1.cdnformanga.xyz/uploads/manga/2774/cover.webp

Both answer `403 "This bucket cannot be viewed"` (Google Cloud Storage bucket not public), also with a `Referer` header. The website itself only shows the cover because it goes through the Jetpack Photon proxy (`i3.wp.com/cdn1.cdnformanga.xyz/...` with query params); a plain `i0.wp.com/...` request is also 403.

Other series from the same source load their covers fine (9 of 11 in my library), so it seems to affect only some series whose cover was moved to that bucket.

### Mihon/Mihon fork version

Suwayomi-Server (latest); also reproducible with curl

### Android version

N/A

### Other details

Possible fix on the extension side: for `cdnformanga.xyz` covers, request the same Photon-proxied URL the site uses (`` on the series page instead of `og:image`), or fall back to it when the origin returns 403.

### Acknowledgements

- [X] I have searched [existing issues](https://github.com/keiyoushi/extensions-source/issues) both open & closed, and confirm that this is a new unreported issue.
- [X] I have updated the app to version **[0.20.1](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I have updated the extension to the latest version (as listed on the [extensions listing](https://keiyoushi.github.io/extensions/) page).
- [X] I have tried the [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the app repository.
- [X] I will correctly fill out all of the requested information in this form.

### 

---

brdteam:
Closes #19146

Covers on ruyamanga2.com are served through Jetpack Photon (`i*.wp.com/cdn1.cdnformanga.xyz/...`). The origin bucket answers 403 (private GCS bucket), so Photon can only serve variants it already has cached, and it does so only when the request accepts webp (`Vary: Accept`; OkHttp sends no `Accept` by default). On top of that the listing markup uses `?w=175&resize=175,238&ssl=1` while the details/search pages use `?w=193&resize=193,278&ssl=1`, and Photon caches those independently - for some series only one of them exists (the site's own popular page fails to show 7 of 10 covers for the same reason).

Changes:
- `configureHeaders`: add `Accept: image/webp,*/*`
- `configureClient`: interceptor that, on a 403 from `*.wp.com`, retries once with the other size variant. The query is written literally because Photon keys its cache on the exact query string (`HttpUrl.Builder` reorders params and encodes the comma, which yields a different key).

Tested on Suwayomi-Server (not Android Studio): before the change, `/thumbnail` returned 500 (upstream 403) for the series in the issue; after, all 125 covers across the first 5 popular pages load, including the two from the issue.

## Checklist

🤖 Opened by an AI agent (Claude Code), asked to fix #19146.


src/tr/ruyamanga/src/eu/kanade/tachiyomi/extension/tr/ruyamanga/RuyaManga.kt
AwkwardPeak7, line 1: maybe remove the photon proxy from urls in the interceptor instead?

brdteam: Tried that first - the origin bucket is what's actually broken. `https://cdn1.cdnformanga.xyz/uploads/manga/3324/cover.webp` (and every other cover I checked, with or without Referer) returns `403 "This bucket cannot be viewed"`; the site itself only shows covers because Photon still has them cached. Photon can't refetch from origin either, which is why it only serves the exact variants it already has (`w=175`/`w=193`) and only with a webp Accept. Stripping the proxy would make every cover fail instead of a handful.

AwkwardPeak7 (approved):

PR #19217, opened 2026-09-19
Team X (ar), MangaSpark (ar), MangaStarz (ar): send browser headers to bypass Cloudflare HTTP 400

Momokh99:
HTTP 400 errors occurred when fetching manga/reader data from olympustaff.com, sparkmanga.net, and starzmanga.com. Send a browser User-Agent and Accept-Language header on all requests.

Tested by compiling debug APKs and installing in Komikku; downloads and reader verified working.

- [x] Updated versionCode (teamx 34, mangaspark 9, mangastarz 11)
- [x] Have compiled and run the extension


src/ar/teamx/src/eu/kanade/tachiyomi/extension/ar/teamx/TeamX.kt
```diff
@@ -272,3 +278,6 @@ abstract class TeamX : KeiSource() {
         else -> SManga.UNKNOWN
     }
 }
+
+private const val BROWSER_USER_AGENT =
```
vetleledaal: Can `:lib:randomua` be used instead?

Momokh99: i did as suggested is it implemented correctlly

AwkwardPeak7:
why do you need to force a useragent? app defaults to a chrome mobile useragent already. only set the other header, and leave useragent

Momokh99:
sorry its a problem of my own i use a proxy to compress images and it was the one who bocked the requests 
i forgot i have one

